PR #149, opened 2020-08-06
Add SDK log level selection ability to demo and library

devalevenkatesh:
**Issue #:** 

**Description of changes:**
- Adds `logLevel` prop to `MeetingProvider` in-order to update the `MeetingManager` `logLevel`.
- Log levels selection can now be done using URL query params. For e.g. `?logLevel=debug`
- By default it is `LogLevel.WARN`, but can select from `LogLevel.DEBUG`, `LogLevel.INFO`, `LogLevel.ERROR` or `LogLevel.OFF`

**Testing**

2. How did you test these changes? Locally running the demo and checking the logs in console

3. If you made changes to the component library, have you provided corresponding documentation changes? No changes to component library which need documentation changes


src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -120,6 +122,10 @@ export class MeetingManager implements AudioVideoObserver{
     this.audioVideoObservers = {};
   }
 
+  setMeetingConfigurationLogLevel = (logLevel: LogLevel) => {
+    this.meetingConfiguration.logLevel = logLevel;
+  }
```
p-foucht, line 127: Rather than having an explicit API, I think it would be a bit better if we allowed it to be passed with the MeetingProvider

```
const Root = () => {
  <MeetingProvider logLevel="warn">
     <App />
  </MeetingProvider>
}
```

Something like that maybe. What do you think?

demo/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -128,6 +140,13 @@ const MeetingForm: React.FC = () => {
         value={region}
         label="Select AWS region"
       />
+      <FormField
+        field={Select}
+        options={getFormattedOptionsForSelect(LOG_LEVELS)}
+        onChange={handleLogLevelChange}
+        value={logLevel}
+        label="Select SDK log level"
+      />
```
p-foucht, line 149: I don't think this should be a part of the form. It's really just for developers, not for users

devalevenkatesh: Hmm, Ok, I thought it would be good to do that selection from UI itself without needing to change the code elsewhere. But yes, we can go the prop route as you said above.

p-foucht: If we want to be able to toggle the log level in our demo, we could probably just stick it in the query params

?logging=info or something

devalevenkatesh: Thats cool, dint thought of this approach. Thanks will add it.

src/providers/MeetingProvider/MeetingManager.ts
devalevenkatesh: Yes this can also be done without explicit API, will update it.

src/providers/MeetingProvider/index.tsx
```diff
@@ -12,11 +12,17 @@ import { LocalVideoProvider } from '../LocalVideoProvider';
 import { FeaturedVideoTileProvider } from '../FeaturedVideoTileProvider';
 import { LocalAudioOutputProvider } from '../LocalAudioOutputProvider';
 import { ContentShareProvider } from '../ContentShareProvider';
+import { LogLevel } from 'amazon-chime-sdk-js';
+
+interface Props {
+  logLevel?: LogLevel;
+  children: ReactNode;
```
p-foucht, line 19: `React.FC` includes children by default, so I don't think you should need to include that here

Also can you update the MeetingProvider/MeetingManager docs with this change?

demo/meeting/README.MD
```diff
@@ -20,9 +20,31 @@ This is the SDK react meeting demo app. It shows how to use the Amazon Chime SDK
 
 ### Troubleshooting
 
+- You can change the `amazon-chime-sdk-js` [log levels](https://aws.github.io/amazon-chime-sdk-js/enums/loglevel.html) by providing `logLevel` as a query parameter in the URL.
+  - WARN logs (default)
```
p-foucht, line 24: To avoid making the readme too long, can we reduce this to something like 

`You can adjust the log levels by including a "logLevel" param with one of the following values: 'warn', 'info', etc...

See the MeetingProvider docs [link to docs] for more information around logging"

demo/meeting/src/app.tsx
```diff
@@ -62,4 +64,27 @@ const Theme: React.FC = ({ children }) => {
   );
 };
 
+function useQuery() {
+  return new URLSearchParams(useLocation().search);
+}
+
+const DemoMeetingProvider: React.FC = ({ children }) => {
+  const query = useQuery();
+  const querylogLevel = query.get('logLevel');
```
p-foucht, line 73: I think we could just keep the original <MeetingProvider> in the `App` above and calculate this logLevel value outside of the component, since it's only occurring when we initialize MeetingManager when we first load the app

```
const params = new URLSearchParams(window.location.search);
const logLevel = urlParams.get('logLevel') || 'warn';

const App = () => {
 <MeetingProvider logLevel={logLevel}>

 </MeetingProvider>
}
```

src/providers/MeetingProvider/index.tsx
devalevenkatesh: Got it, updated the docs as well. Thanks!

demo/meeting/README.MD
devalevenkatesh: Is this fine? 
- You can change the `amazon-chime-sdk-js` [log levels](https://aws.github.io/amazon-chime-sdk-js/enums/loglevel.html) by providing `logLevel` as a query parameter with one of these values 'warn' | 'error' | 'info' | 'debug' | 'off'.
   ```
    https://0.0.0.0:9000/?logLevel=debug
    ```

demo/meeting/src/app.tsx
devalevenkatesh: Updated to:
```
const urlParams = new URLSearchParams(window.location.search);
const queryLogLevel = urlParams.get('logLevel') || 'warn';
const logLevel = SDK_LOG_LEVELS[queryLogLevel] || SDK_LOG_LEVELS['warn'];
```

PR #162, opened 2020-08-14
Fix css prop not taking precedence over queries

p-foucht:
**Issue #:** 
Styled Components adjusts the order of style declarations when it comes to media queries, so the `css` prop was not taking precedence over previously declared styles

[Ticket with more details here](https://github.com/styled-components/styled-components/issues/2095)

**Description of changes:**
- Wrap the `css` prop in a blank media query so it retains its order

**Testing**

2. How did you test these changes? manually

3. If you made changes to the component library, have you provided corresponding documentation changes? n/a

### DIFF
--- a/src/components/ui/Base/index.ts
+++ b/src/components/ui/Base/index.ts
@@ -13,5 +13,6 @@ export interface BaseProps extends SpaceProps {
   className?: string;
 }
 
-export const baseStyles = ({ css }: BaseProps) => (css ? `${css};` : '');
+export const baseStyles = ({ css }: BaseProps) =>
+  css ? `@media { ${css} };` : '';
 export const baseSpacing = (props: BaseProps) => space(props);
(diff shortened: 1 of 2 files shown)


src/components/ui/Base/index.ts
```diff
@@ -13,5 +13,6 @@ export interface BaseProps extends SpaceProps {
   className?: string;
 }
 
-export const baseStyles = ({ css }: BaseProps) => (css ? `${css};` : '');
+export const baseStyles = ({ css }: BaseProps) =>
+  css ? `@media { ${css} };` : '';
```
bradmanAWS, line 17: Is there a way to write responsive media queries in the css prop?

p-foucht: Yeah you should be able to still pass in media queries as part of the `css` string

PR #197, opened 2020-09-10
Creating chat bubble component

bradmanAWS:
**Description of changes:**
Add chat message component and update search input snapshot tests

**Testing**
yes
unit tests and snapshot tests
Yes


src/components/ui/Chat/Message/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+const Message = (props: MessageProps) => {
+
+    return (
+        <StyledMessage {...props} data-testid='chat-message'>
+            <StyledMessageContent {...props}>
+                <div className='ch-sender-name'>{props.name}</div>
+                <div>{props.content}</div>
+                {props.showCaret && <ChatCaret height='9px' width='4px' />}
+            </StyledMessageContent>
+            <StyledMessageInfo>
```
davejlane, line 38: If `moreOptions` is an optional prop, the `PopOver` that renders them should be conditionally rendered based on `moreOptions` being non-null. I can think of use cases where someone might want to build a chat experience w/o including a menu.

src/components/ui/Chat/Message/Styled.tsx
```diff
@@ -0,0 +1,83 @@
...
+    font-weight: bold;
+    margin-bottom: 0.5rem;
+  }
+
+  & svg {
+    position: absolute;
+    margin-top: -3px;
+    margin-left: -20px;
+
+    & g g g {
```
davejlane, line 45: That's quite an unusual selector, and possibly not very efficient. Why was it written this way?

src/components/ui/icons/ChatCaret/index.tsx
```diff
@@ -0,0 +1,20 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+import Svg, { SvgProps } from '../Svg';
+
+const ChatCaret: React.SFC<SvgProps> = props => (
+  <Svg {...props} viewBox="0 0 4 9">
```
davejlane, line 8: Why doesn't this icon have the same viewBox dimensions as the others? What is different about the `ChatCaret` compared to the existing `Caret`?

bradmanAWS: It's not really the same at all. Per the design, some chat bubbles will have the little fin on the bubble and some won't. It's not used in the same way as any other icon and would only be attached to a chat bubble  <div>.

src/components/ui/Chat/Message/Styled.tsx
bradmanAWS: This caret is an Svg component that is constructed this way. The most inner <g> fill prop is what colors the icon.

src/components/ui/Chat/Message/index.tsx
bradmanAWS: Great call out. I'll update.

src/components/ui/icons/ChatCaret/index.tsx
bradmanAWS: Maybe since this icon is so different and specific, it should only be defined in the html of the chat message component. What do you think?

davejlane: Thanks for explaining. I was having a hard time visualizing what this is used for / looked like. I can't think of any other scenario where this might be used, which makes it a good candidate for keeping it with the main chat message component. I'd recommend doing that, we can always isolate it if there ends up being a need for it somewhere else.

src/components/ui/Chat/Message/Styled.tsx
davejlane: Gotcha. In cases like these, I tend to add something like a `ClassName` to the element I want to target so that the browser has to do less work to match the element with the style.  It's a small thing, but tends to add up when you have a lot of selectors like this.

src/components/ui/icons/ChatCaret/index.tsx
bradmanAWS: Screen Shot 2020-09-10 at 1 53 48 PM

It's only the tiny tail on this chat bubble. I'm updating to just be part of the chat message component now.

src/components/ui/Chat/Message/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+    return (
+        <StyledMessage {...props} data-testid='chat-message'>
+            <StyledMessageContent {...props}>
+                <div className='ch-sender-name'>{props.name}</div>
+                <div>{props.content}</div>
+                {props.showCaret &&   
+                <svg viewBox="0 0 4 9">
+                    <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+                    <g transform="translate(0.000000, -53.000000)" className='ch-chat-caret'>
+                        <path d="M4,62 L3.92789928,61.999999 C2.89671177,62.0004988 1.33197354,61.8123902 0.200755581,60.8250184 C-0.0781312712,60.5814641 -0.0639788041,60.0290387 0.229060515,59.8181428 C1.47198013,58.9247413 3.99237825,57.6821586 4,52.9112516 L4,52.9112516 L4,62 Z" id="Combined-Shape"/>
```
p-foucht, line 40: Why does this have an ID?

bradmanAWS: The data-testid is used in tst/components/ui/Chat/Message/index.test.tsx

src/components/ui/Chat/Message/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+                    </g>
+                    </g>
+                </svg>}
+            </StyledMessageContent>
+            <StyledMessageInfo>
+                <span>{props.time}</span>
+                {props.moreOptions?.length && <PopOver
```
p-foucht, line 47: Probably want to use a ternary statement here. If a length of 0 is passed, you'll just render "0"

src/components/ui/Chat/Message/Message.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { useTheme } from 'styled-components';
+import { select, boolean } from '@storybook/addon-knobs';
+
+import Flex from '../../Flex';
+import Message from './';
+import MessageDocs from './Message.mdx';
+import { PopOverItemProps } from '../../PopOver/PopOverItem';
+
+export default {
+  title: 'UI Components/Chat',
```
p-foucht, line 14: Will this render as UI Components > Chat > Message?

Dave mentioned this in the past, but we probably want to group our components and documentation as Chat vs Meeting

src/components/ui/Chat/Message/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+    time: string;
+    /* The text content of the message. */
+    content: string;
+    /* The options available in the more options button. */
+    moreOptions?: PopOverItemProps[] | null;
+    /* Adds the bubble caret to a message. */
+    showCaret?: boolean;
+}
+
+const Message = (props: MessageProps) => {
```
p-foucht, line 29: I know this is out of scope for the initial release, but just in terms of naming - you might want to think about Markdown and attachments and other message variants. For example, if you plan on just creating separate components for a MarkdownMessage and AttachmentMessage, should this be a TextMessage? Or do you think you'll end up managing that all through props with a single component?

src/components/ui/Chat/Message/index.tsx
bradmanAWS: Wow, this was just an oversight by me. I thought a value of "0" would be falsey and would therefore be ignored. I'll add better testing here too.

src/components/ui/Chat/Message/Message.stories.tsx
bradmanAWS: Yeah, it will follow that folder pattern. That's what we want, right?

src/components/ui/Chat/Message/index.tsx
bradmanAWS: I haven't gotten that far ahead. Initially, I believe we can handle an attachment inside this same component as a prop. If you have some vision around markdown and would recommend naming this to be a different component, I'm open to that.

src/components/ui/Chat/Message/index.tsx
p-foucht: This has an `id` attribute, not a `data-testid` attribute though

src/components/ui/Chat/Message/Message.stories.tsx
p-foucht: Yeah, I think grouping them by purpose should be good. Eventually we should probably do the same for a /meetings directory too

src/components/ui/Chat/Message/index.tsx
p-foucht: Nope no vision, it was just a general comment to ask if we were thinking about that stuff. Sounds like we are though!

src/components/ui/Chat/Message/index.tsx
bradmanAWS: Ooooh, the `<path>`. I didn't see it at the end. I'll remove. Great catch.

p-foucht (approved):

PR #200, opened 2020-09-10
Create a chat InfiniteList component

bradmanAWS:
**Description of changes:**
This adds an infinite list container component that will be used with chat. It hands loading new items into a list upon scroll.

**Testing**
Yes
Unit tests and visually in storybook
Yes


src/components/ui/InfiniteList/index.tsx
```diff
@@ -0,0 +1,90 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { useRef, useEffect, useState, HTMLAttributes, ReactNode } from 'react';
+
+import { StyledInfiniteList } from './Styled';
+import { BaseProps } from '../Base';
+import { Cog as Spinner } from '../icons'; // TODO: update icon
```
davejlane, line 8: I'm surprised that we don't already have a `spinner` icon. We should remedy that with design. 

One piece of feedback we have been getting from users is that for components such as this, which have both logic and some visual elements, they would like the option to pass in their own ui components. In this component we could include a prop for the spinner component, and the defaultProp could be this one. What do you think?

src/components/ui/InfiniteList/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const currentTopItem = useRef<HTMLLIElement>(null);
+    const firstNew = useRef<HTMLLIElement>(null);
+    const prevLength = useRef(items.length)
+    const newLength = useRef(0);
+
+    const [isLoading, setIsLoading] = useState(false);
+
+    // Scroll to the bottom of the list on load
+    useEffect(() => {
+        feedRef.current?.addEventListener('scroll', handleScroll);
```
davejlane, line 33: I would strongly recommend debouncing any listeners attached to scroll events, as they can fire several times per second.

src/components/ui/InfiniteList/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+            setIsLoading(true)
+            setTimeout(
+                () => {
+                    handleLoad()
+                    setIsLoading(false)
+                    // scroll back down to the first new message
+                    firstNew.current?.scrollIntoView();
+                },
+                // simulate an async messages fetch
+                500
```
davejlane, line 50: Can you explain the comment and the use of a timeout here?

src/components/ui/InfiniteList/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+    useEffect(() => {
+        if (items.length !== prevLength.current) {
+            prevLength.current = newLength.current;
+        }
+        newLength.current = items.length;
+    });
+
+    return (
+        <StyledInfiniteList ref={feedRef} className={isLoading ? 'ch-not-scrollable' : ''} {...props} data-testid='infinite-list'>
+            {isLoading && <li className="spinner"><Spinner /></li>}
```
davejlane, line 68: Should the className be prefixed with "ch-"?

src/components/ui/InfiniteList/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+                    // It's not the initial load
+                    items.length > batchSize &&
+                    // It's the first message from the new batch
+                    index === newLength.current - prevLength.current - 1 &&
+                    // New messages were loaded
+                    newLength.current !== prevLength.current
+                ) {
+                    return <li key={index} ref={firstNew}>{item}</li>
+                }
+                return <li key={index}>{item}</li>
```
davejlane, line 83: It might be cleaner to have a method that decides what the `ref` value is for the `<li>` and assign `null` when it doesn't match the other 2 cases. So:

```
  getRef = () => {
    // all of this logic is now outside of the render block
  }
 <li key={index} ref={getRef()}>{item{/li>
```

src/components/ui/InfiniteList/index.tsx
bradmanAWS: I'll look into that!

src/components/ui/InfiniteList/index.tsx
bradmanAWS: This is supposed to visually simulate an async fetch for more messages. For instance, the intended use case, when a user scrolls to the top of a chat feed, the api call to load more messages will take some time. This 500ms timeout is to test that feel and have some opportunity to show the spinner.

src/components/ui/InfiniteList/index.tsx
bradmanAWS: I actually meant to fix the spinner icon and this portion of the code before submitting the PR. I'll update for the next revision.

src/components/ui/InfiniteList/index.tsx
davejlane: Ah, gotcha. I've seen some implementations where the fetching is done _as the user is scrolling towards the top of the feed_, so that when they do reach the top, more often than not, the fetch has already occurred and the new content is loaded, which makes for a more seamless experience. IDK what the design guidance on this component was though.

One other thing to consider: screen readers will not scroll a page, so anyone not using a mouse or a trackpad is not going to be able to fetch more content. A lot of times, a button might be rendered (sometimes visually hidden, but screen-readable) so that those users that are traversing the DOM using a keyboard have a means to trigger the fetch. We should consider including that here.

src/components/ui/InfiniteList/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const prevLength = useRef(items.length)
+    const newLength = useRef(0);
+
+    const [isLoading, setIsLoading] = useState(false);
+
+    // Scroll to the bottom of the list on load
+    useEffect(() => {
+        feedRef.current?.addEventListener('scroll', handleScroll);
+        scrollToBottom();
+        return feedRef.current?.removeEventListener('scroll', handleScroll, true)
```
p-foucht, line 35: `return () => feedRef.current.remove...`, right?

src/components/ui/InfiniteList/InfiniteList.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+  };
+
+  const [items, setItems] = useState([...createBatch()]);
+  const itemsRef = useRef(items);
+
+  useEffect(() => {
+      itemsRef.current = items;
+      batchRef.current = batchNumber;
+  });
```
p-foucht, line 50: Is there any need for this to be in an effect?

src/components/ui/InfiniteList/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const scrollToBottom = () => {
+        listEnd.current?.scrollIntoView()
+    };
+
+    useEffect(() => {
+        if (items.length !== prevLength.current) {
+            prevLength.current = newLength.current;
+        }
+        newLength.current = items.length;
+    });
```
p-foucht, line 64: Does this need to be in an effect?

src/components/ui/InfiniteList/index.tsx
p-foucht: Do we actually need `firstNew`? It looks like we're just using it to scroll back into view, but what's the reason we can't just scroll to the `listEnd` element below and get rid of all of this logic?

src/components/ui/InfiniteList/index.tsx
p-foucht: Did we consider using `IntersectionObserver` here, instead of scroll listeners? 

Related article [here](https://medium.com/walmartglobaltech/infinite-scrolling-the-right-way-11b098a08815). Google touches on it [here](https://developers.google.com/web/updates/2016/07/infinite-scroller), and acknowledges that there may be issues with it (although the article is from 2017). It would probably worth some consideration, at least

src/components/ui/InfiniteList/Styled.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+
+export const StyledInfiniteList = styled.ul<InfiniteListProps>`
+  overflow-y: scroll;
+  display: flex;
+  flex-direction: column;
+  list-style: none;
+  padding: 0;
+
+  /* disable scrolling with fetching */
```
p-foucht, line 27: while*

src/components/ui/InfiniteList/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+import { StyledInfiniteList } from './Styled';
+import { BaseProps } from '../Base';
+import { Cog as Spinner } from '../icons'; // TODO: update icon
+
+export interface InfiniteListProps extends Omit<HTMLAttributes<HTMLUListElement>, 'css'>, BaseProps {
+    /* The elements that will populat the list. These elements will be wrapped in <li> tags, so that can be any element. */
+    items: ReactNode[];
+    /* A callback function that will make an api call to load the next batch of items. */
+    handleLoad: () => void;
```
p-foucht, line 14: If we're managing the loading state, this needs to be `async`

src/components/ui/InfiniteList/Styled.tsx
bradmanAWS: Thanks. Fixed.

I don't know why, but I can't reply to your comment about `IntersectionObservers`. I looked into this briefly, but I was able to achieve my goal without using it. I'll look further into it if it'll be advantageous.

src/components/ui/InfiniteList/index.tsx
bradmanAWS: Right, I'll update.

src/components/ui/InfiniteList/index.tsx
bradmanAWS: Same answer as above. If we don't update the values continually, the data becomes stale.

src/components/ui/InfiniteList/InfiniteList.stories.tsx
bradmanAWS: I think so. In order to keep these values up to date in real time, they need to be updated continually. If we try to just reference state, the data becomes stale.

src/components/ui/InfiniteList/index.tsx
bradmanAWS: Dang it, yes. That gets me every time.

src/components/ui/InfiniteList/index.tsx
bradmanAWS: `firstNew` ensures that the first item in the new batch is scrolled into view. In my story included here, you have 250 items in a list. First you load with item 201-250 showing. Once you scroll up and the next batch is loaded, item 200 will be at the top of the visible list.

src/components/ui/InfiniteList/InfiniteList.stories.tsx
p-foucht: I was more asking if we need this logic in the effect, vs just in the render function directly. This is running on every render anyways, so I was wondering if there was a reason it needed explicitly to be in an effect

Since you mention it though, where is the data going stale? I don't see us creating an closures, so I'm wondering where we're needing these refs instead of just state

src/components/ui/InfiniteList/index.tsx
bradmanAWS: I tried to implement this, but I'm running into a problem. Say you are looking at items 51 - 100. Currently, when you scroll up to item 51, it will load 1 - 50 and force item 50 to the top of the screen. It's a pretty smooth transition. 

Do you have a recommendation of how to implement your suggestion above? I can't implement the current behavior. Say it fetches the next batch when you get to item 60 instead of 51, in order to preload the next messages. Currently the list would just from item 60 to item 50. Not great. But if we don't implement any programmatic scrolling, you'll jump all the way to item 1. How could we maintain the user's position without freezing the list when the items load?

src/components/ui/InfiniteList/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const feedRef = useRef<HTMLUListElement>(null);
+  const listEnd = useRef<HTMLDivElement>(null);
+  const currentTopItem = useRef<HTMLLIElement>(null);
+  const firstNew = useRef<HTMLLIElement>(null);
+  const prevLength = useRef(items.length)
+  const newLength = useRef(0);
+
+  const handleLoad = async () => {
+    await onLoad();
+  }
```
p-foucht, line 33: We aren't awaiting anything since we removed the loading responsibility. We should just be able to call `onLoad` directly in our IntersectionObserver

src/components/ui/InfiniteList/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  useEffect(() => {
+    firstNew.current?.scrollIntoView();
+  }, [items.length])
+
+  useEffect(() => {
+    listEnd.current?.scrollIntoView();
+    const observer = new IntersectionObserver(entries => {
+      const topEntry = entries[0];
+      if (topEntry.isIntersecting) {
+        handleLoad()
```
p-foucht, line 44: handleLoad/onLoad can go stale in this effect

src/components/ui/InfiniteList/InfiniteList.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  const [isLoading, setIsLoading] = useState(false);
+
+  const [items, setItems] = useState([...createBatch()]);
+  const itemsRef = useRef(items);
+
+  useEffect(() => {
+    itemsRef.current = items;
+    batchRef.current = batchNumber;
+  });
```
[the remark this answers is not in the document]
bradmanAWS: The state does go stale, and it does not function properly outside of an effect. I think the state goes stale as the `onLoad` function is passed to the list component. Maybe I'm misundestanding something, but the list never populates in order with any other data arrangement.

src/components/ui/InfiniteList/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+import { StyledInfiniteList } from './Styled';
+import { BaseProps } from '../Base';
+import { Spinner } from '../icons'; // TODO: update icon
+
+export interface InfiniteListProps extends Omit<HTMLAttributes<HTMLUListElement>, 'css'>, BaseProps {
+  /* The elements that will populat the list. These elements will be wrapped in <li> tags, so that can be any element. */
+  items: JSX.Element[];
+  /* A callback function that will make an api call to load the next batch of items. */
+  onLoad: () => Promise<void>;
```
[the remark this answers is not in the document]
bradmanAWS: Correct, I went backwards on that one.

src/components/ui/InfiniteList/index.tsx
bradmanAWS: Fixed.

src/components/ui/InfiniteList/InfiniteList.stories.tsx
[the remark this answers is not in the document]
p-foucht: If onLoad is going stale when passing it to InfiniteList, then that sounds like a bug in InfiniteList. A developer should be able to update their handler as needed and not have to create a workaround with refs

src/components/ui/InfiniteList/index.tsx
bradmanAWS: Can you expand on this? I see that `handleLoad` is not needed and we can just call `onLoad` directly. What would go stale here?

p-foucht: Your affect runs on mount, so it will only reference the original handleLoad function passed via props (which references original onLoad function passed). onLoad can be updated and it will still be calling the original onLoad function

Whether you use handleLoad or onLoad directly, you'll need to add either one of them to the dependency array (I would do onLoad directly, handleLoad will cause the effect to run every render anyways)

src/components/ui/InfiniteList/InfiniteList.stories.tsx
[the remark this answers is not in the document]
bradmanAWS: Updated

src/components/ui/InfiniteList/index.tsx
bradmanAWS: Fixed

p-foucht (approved):

PR #208, opened 2020-09-16
Add chat channel list component

bradmanAWS:
**Description of changes:**
Add channel list component

**Testing**
Yes
Unit and snapshot tests
Yes


src/components/ui/Chat/ChannelList/ChannelItem.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export const ChannelItem = (props: ChannelItemProps) => {
+  const { name, arn, moreOptions } = props;
+  const context = useChannelListContext();
+  const isActive = context.activeChannelArn === arn;
+  return(
+    <StyledChannelItem
+      {...props}
+      className={isActive ? 'isActive' : ''}
```
p-foucht, line 27: For classes, we typically just add "active" instead of "isActive"

Also, please add "ch-" prefix

src/components/ui/Chat/ChannelList/ChannelList.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+```jsx
+<ChannelList
+  activeChannelArn={activeChannelArn}
+  setActiveChannelArn={setActiveChannelArn}
+>
+  <ChannelItem
+    name='Channel 1'
+    arn='abc'
+    moreOptions={[{ children: <span>Option 1</span>, onClick: () => setActiveChannelArn('abc') }]}
```
p-foucht, line 36: Little hard to get this without seeing the UI, but why do we need an onClick for setting the activeChannelArn when we have `onClick={() => context.setActiveChannelArn(arn)}` in `ChannelItem` above?

src/components/ui/Chat/ChannelList/ChannelItem.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export interface ChannelItemProps extends Omit<HTMLAttributes<HTMLLIElement & HTMLLIElement>, 'css'>, BaseProps {
+  name: string;
+  arn: string;
+  moreOptions?: PopOverItemProps[];
+}
+
+export const ChannelItem = (props: ChannelItemProps) => {
+  const { name, arn, moreOptions } = props;
+  const context = useChannelListContext();
```
p-foucht, line 22: Not sure how I feel about tying a UI component to a context provider. Maybe this is something the consumer should manage

bradmanAWS: What's the draw back? The reason to use context here is so the `ChannelList` container can render children implicitly and manage which `ChannelItem` is active. I thought this would be a lightweight solution to accomplish this.

src/components/ui/Chat/ChannelList/ChannelList.mdx
bradmanAWS: OOPS! No, this is just a mistake. Fixing.

src/components/ui/Chat/ChannelList/ChannelItem.tsx
bradmanAWS: Yes, I missed the `ch-`. Fixing. I usually use `isActive` so that in the css `element.active` and `element:active` don't get mixed up. I can change it though.

src/components/ui/Chat/ChannelList/ChannelItem.tsx
p-foucht: The two drawbacks I see is that the UI is now tied to some piece of context/state. So if you wanted to use this UI in another scenario, you couldn't

The other is a matter of responsibility. It seems like it's getting a bit close to managing application state

Just something to be aware of, and depends on how you see the components/SDK integration working with external developers' apps. I'm good with approving if you're already considering this and think this is a good solution

bradmanAWS: I understand what you mean. I guess this state management is pretty much the main reason this `channelList` component exists. Otherwise, it's just a list of buttons with slightly different styling. Also, nothing would break if a developer just didn't use the `activeChannelArn` property. The other solution I see to this is passing an array of children to the channelList and managing the `activeChannel` directly inside the `channelList` component. I know we also prefer implicitly rendered children. 

If you think there is a better solution, I'm up for it, or if you think we should generate the child elements explicitly, we could do that. Let me know.

p-foucht (approved):

p-foucht (approved):

p-foucht (approved):

PR #212, opened 2020-09-21
Include prop for ControlBarButton active state

davejlane:
**Description of changes:**
Currently there's not way to manipulate a `button` in the `ControlBar` to give visual feedback that it is active, or toggled. This can be useful to give more visual feedback for states such as when a users' camera is on. This change introduces an optional `isActive` prop for a `ControlBarButton` component.

**Testing**
I am experiencing issues with running snapshot tests. Other than that release ran without issue.

Manually, (and hopefully with an updated snapshot test when I can get them to work).

Yes

### DIFF
--- a/src/theme/dark.ts
+++ b/src/theme/dark.ts
@@ -307,7 +307,11 @@ const controlBar = {
   shadow: shadows.large,
   bgd: colors.greys.grey100,
   border: 'none',
-  opacity: 1
+  opacity: 1,
+  selected: {
+    text: buttons.primary.selected.text,
+    bgd: buttons.primary.selected.bgd,
+  },
 };
 
 const roster = {
--- a/src/theme/light.ts
+++ b/src/theme/light.ts
@@ -307,7 +307,11 @@ const controlBar = {
   shadow: shadows.large,
   bgd: colors.greys.white,
   border: `0.03125rem solid ${colors.greys.grey20}`,
-  opacity: 0.95
+  opacity: 0.95,
+  selected: {
+    text: buttons.primary.selected.text,
+    bgd: buttons.primary.selected.bgd,
+  },
 };
 
 const roster = {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `useLocalAudioInputActivityPreview` hook for direct access to microphone input value
 - Add chat message component
+- Add `isSelected` prop to `ControlBarButton` component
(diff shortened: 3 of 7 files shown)


src/theme/dark.ts
```diff
@@ -307,7 +307,11 @@ const controlBar = {
   shadow: shadows.large,
   bgd: colors.greys.grey100,
   border: 'none',
-  opacity: 1
+  opacity: 1,
+  active: {
```
bradmanAWS, line 311: Is it confusing that sometimes `active` styling is applied when an element is actually in its active state (being clicked) and sometimes it means the property is selected? Maybe not, but just a thought.

davejlane: Fair point. Are you saying you'd prefer "selected'?

bradmanAWS: I like `selected`. Looks good otherwise!

davejlane: I like it too. Looking at the `theme` file you're absolutely correct; we use `active` for the CSS :active selector, so it is confusing to use it in this context. Updating.

src/theme/dark.ts
```diff
@@ -307,7 +307,11 @@ const controlBar = {
   shadow: shadows.large,
   bgd: colors.greys.grey100,
   border: 'none',
-  opacity: 1
+  opacity: 1,
+  selected: {
+    text: buttons.primary.active.text,
+    bgd: buttons.primary.active.bgd,
```
bradmanAWS, line 313: selected?

src/theme/light.ts
```diff
@@ -307,7 +307,11 @@ const controlBar = {
   shadow: shadows.large,
   bgd: colors.greys.white,
   border: `0.03125rem solid ${colors.greys.grey20}`,
-  opacity: 0.95
+  opacity: 0.95,
+  selected: {
+    text: buttons.primary.active.text,
+    bgd: buttons.primary.active.bgd,
```
bradmanAWS, line 313: selected?

src/theme/dark.ts
davejlane: That would help, wouldn't it? Updating.

src/theme/light.ts
davejlane: Updating.

src/theme/dark.ts
bradmanAWS: I do that all the time. It's specifically hard to compared updates here vs crux. I'm still getting used to it.

bradmanAWS (approved):

bradmanAWS (approved):

PR #226, opened 2020-10-06
change base prop "css" to "styles

davejlane:
**Description of changes:**
This change will solve a couple of issues found when using this library in certain projects.
- The `css` prop can become conflicted when a project is using other CSS-in-JS frameworks, such as emotion, as that also declares a `css` property, but the type is an object.
- We can now remove most prop declarations that involve the use of TS `Omit`, which is only available after v 3.5.1. 
- This also disambiguates the base `css` props from the `css` method supplied my `styled-components`.

**Testing**
1. Have you successfully run `npm run build:release` locally? No

2. How did you test these changes? Running tests locally. Building storybook and the demo app locally and inspecting components to ensure that additional styles are still being applied.

3. If you made changes to the component library, have you provided corresponding documentation changes? YEs


bradmanAWS (commented):
Can you confirm snapshot tests ran and passed?

bradmanAWS (changes requested):
You probably have old "css" snapshot image files you can delete, right?

davejlane:
> Can you confirm snapshot tests ran and passed?

Yes, I had to update some assertions, but that was enough to have the tests pass.

davejlane:
> You probably have old "css" snapshot image files you can delete, right?

Do these have to be done manually?

bradmanAWS (approved):

PR #229, opened 2020-10-07
Add MessageList component and refactor Message component

bradmanAWS:
**Description of changes:**
This change adds a styleMessageList function that can be used to construct the chat list including using ChatBubbles, styling each chat bubble according to UX designs relative to other chat bubbles in the list, and inserting DateHeaders into the list. This is demonstrated independently and integrated with the InfiniteList component in storybook.

**Testing**
Yes
`npm run test`, `npm run test:snapahots`, and visually tested in storybook
Yes


src/components/ui/Chat/MessageList/DateHeader.tsx
```diff
@@ -0,0 +1,34 @@
...
+    if (date === yesterday) {
+      return 'Yesterday'
+    } else if (date === today) {
+      return 'Today'
+    }
+    return (
+      date
+    )
+  }
+
```
davejlane, line 30: I'd recommend a few things here:
* This is a pretty useful utility, it should live somewhere that other components can access it
* Rather than hard coding words in a specific language, it might be more useful to make those strings arguments of the `formatDate` function (maybe with default args using the english words you have in this?)
* For even more flexibility, you could add a `locale` argument too. 
* You could memoize this function. I'd imagine that there are going to be many of these on a page.

src/components/ui/Chat/MessageList/styleMessageList.tsx
```diff
@@ -0,0 +1,111 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
davejlane, line 1: Nit on the name: usually we identify a styled component with a 'Styled' prefix. It feels a bit confusing to have this name that is similar.

davejlane: Also for some reason there's a lack of semicolons in this file, I won't point out each instance.

src/components/ui/Chat/MessageList/styleMessageList.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const createChatBubbles = (items: MessageProps[]) => {
+    return items.map((m: any, i: number) => {
+      if (!m.content) {
+        return m // not a message
+      } else {
+        let showTail = true;
+        if (
+          items[i].sender?.arn && // it is a message
+          items[i + 1]?.sender?.arn && // the item after is a message
+          items[i].sender.arn === items[i + 1].sender.arn // the item after is from the same sender
```
davejlane, line 73: You sometimes have optional chaining here, and sometimes not. I'd recommend choosing the most defensive option. Also, you're dipping into the `sender` object each time for the `arn`. It would be more performant to cache both of those values.

src/components/ui/Chat/MessageList/styleMessageList.tsx
```diff
@@ -0,0 +1,111 @@
...
+        dateCount++
+      }
+    })
+    return items
+  }
+
+  const createChatBubbles = (items: MessageProps[]) => {
+    return items.map((m: any, i: number) => {
+      if (!m.content) {
+        return m // not a message
```
davejlane, line 67: Just curious, if it's not a message, what could it be?

src/components/ui/Chat/MessageList/DateHeader.tsx
bradmanAWS: I thought about that. I wasn't sure what kind of options/use cases would be needed. I'll tinker around and add some robustness to it. And move it.

src/components/ui/Chat/MessageList/styleMessageList.tsx
bradmanAWS: Yeah, I will update the name and add in the semicolons.

src/components/ui/Chat/MessageList/styleMessageList.tsx
bradmanAWS: I don't have optional chaining where I have already checked the values. If `items[i].sender` is falsey it would fail out on line 71. I can cache the values early, though. I'll update.

src/components/ui/Chat/MessageList/styleMessageList.tsx
bradmanAWS: A date header. Once they get spliced into the list, they could be iterated here.

src/components/ui/Chat/MessageList/DateHeader.tsx
bradmanAWS: Also, I'm not sure about the memoization here, since pretty much every single message will have a unique timestamp, unless I'm missing something.

src/components/ui/Chat/MessageList/formatMessageList.tsx
```diff
@@ -0,0 +1,114 @@
...
+  channelArn?: string;
+  sender: {
+    /* The display name of the sender. */
+    name: string;
+    /* The unique identifier of the sender. */
+    arn: string;
+  }
+};
+
+export const formatMessageList = (messages: MessageProps[], memberArn: string, messageActions: PopOverItemProps[] | null) => {
```
p-foucht, line 31: This function feels a bit off. The naming `formatMessageList` sounds like a utility function for formatting the message data structure, but it's doing a bit of that and additionally returning an array of ChatBubble components

IMO, this should just be a component `<FormattedMessageList/>`, or something along those lines

I would also consider stripping out these utilities and exporting them individually, so that if a user doesn't want to use this component, they can still use the logic for inserting chat bubble tails/formatting time, etc

src/components/ui/Utilities/formatDate.ts
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+export interface DateOptions {
+  weekday?: string;
+  year?: string;
+  month?: string;
+  day?: string;
+}
+
+const formatDate = (dateStr: string, locale: string | undefined, dateOptions?: DateOptions, todayText?: string, yesterdayText?: string) => {
```
p-foucht, line 11: Nice, this seems like a pretty useful utility

src/components/ui/Chat/MessageList/DateHeader.tsx
davejlane: Those messages will re-render as you get more messages. If the timestamp is memoized then the browser doesn't have to re-calculate the formatted date on each render.

src/components/ui/Chat/MessageList/formatMessageList.tsx
```diff
@@ -0,0 +1,114 @@
...
+  /* Determines if a message was redacted (deleted) by a user. */
+  redacted: boolean;
+  /* The unique identifier of the channel the message was sent in. */
+  channelArn?: string;
+  sender: {
+    /* The display name of the sender. */
+    name: string;
+    /* The unique identifier of the sender. */
+    arn: string;
+  }
```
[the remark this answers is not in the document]
bradmanAWS: This data structure matches that of the messages API from the chat service. I thought it would be easy to plug and play this way. I could flatten it which would only require the mildest of data manipulation in the app.

src/components/ui/Chat/MessageList/formatMessageList.tsx
bradmanAWS: Dave and I talked quite a bit about how do to this, but it's still up in the air, and I was curious what you thought. I originally was going to make it a component, but I wanted something that could create styling but could also be used with the <InfiniteList/> component. With the utilities stripped out of <FormattedMessageList/> it wouldn't do anything. This was an attempt to strip these styling functions out of <InfiniteList />. I do see the value in separating them though.

src/components/ui/Chat/MessageList/InfiniteList.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+    }
+  };
+
+  const [isLoading, setIsLoading] = useState(false);
+  const [items, setItems] = useState([...createBatch()]);
+
+  useEffect(() => {
+    setBatchNumber(batchNumber + 1)
+  }, [])
+
```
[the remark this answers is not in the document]
bradmanAWS: This whole chunk? This is a storybook only functionality that replicates an API loading items 201-250, then 151-200, etc.

src/components/ui/Chat/MessageList/DateHeader.tsx
bradmanAWS: Oh, right! Nice call.

src/components/ui/Chat/MessageList/formatMessageList.tsx
p-foucht: Just to be clear, by stripping out the utilities I mean make them independently usable. So `FormattedMessageList` would still be responsible for importing and applying those utility functions and rendering the component list, but the utilities themselves can be used outside of the component if a developer doesn't want to use that List component

bradmanAWS: Yeah, I can split each function out to it's own utility, I think that makes sense. But I don't think `<FormattedMessageList />` should be a component. The component that should actually render `<li>` tags on the DOM already exists (`<InfiniteList/>`). We just need these functions to return an array. I don't think you could use a `<FormattedMessage/>` component and  an `<InfiniteList/>` component together.

src/components/ui/Chat/ChatBubble/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+    /* The name of the user that sent the message. */
+    senderName: string;
+    /* the unique identifier of the sender */
+    createdTimestamp: string;
+    /* The timestamp the message was last edited */
+    lastUpdatedTimestamp?: string;
+    /* The text content of the message. */
+    content: string;
+    /* The options available in the more options button. */
+    messageActions?: PopOverItemProps[] | null;
```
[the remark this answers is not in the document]
bradmanAWS: Ok, updated. I'm not 100% in love with a prop called `messageActions` being type `ReactNode[]` (since it's an element, not a function) but I think you definitely right about moving away from the strict `PopOverItem[]` type.

src/components/ui/Chat/MessageList/formatMessageList.tsx
p-foucht: Hmmm I see. One option would be to adjust InfiniteList so it's more flexible and takes away the "list" responsibility, just manages the infinite scrolling piece of it. An InfiniteList HOC or something like that would be more flexible for future use cases, and potentially more pleasant to work with

Otherwise, we could adjust the name of the function to better indicate what it's doing. `formatMessageList` sounds like it's formatting the data, not really returning an array of components. `createChatBubbleItems` (or something similar) might help people understand the functionality a bit better

src/components/ui/Chat/MessageList/InfiniteList.stories.tsx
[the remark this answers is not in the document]
bradmanAWS: Oh, I see. I think this was an incidental workaround for a setState race condition. I've fixed this.

src/components/ui/Chat/MessageList/InfiniteList.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+        isLoading={isLoading}
+      />
+    </Flex>
+  );
+};
+
+
+BasicInfiniteList.story = {
+  name: 'Basic Infinite List'
+};
```
wootsbot, line 76:
```suggestion
BasicInfiniteList.story = {
  name: 'BasicInfiniteList'
};
```

Maybe I don't understand if there is a convention on when when to do `UpperCamelCase` (`BasicInfiniteList`) or this `Basic Infinite List`

src/components/ui/Chat/MessageList/InfiniteList.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+        css='border: 1px solid #3f4149; width: 40rem; height: 20rem'
+        isLoading={isLoading}
+      />
+    </Flex>
+  );
+};
+
+FormattedMessageInfiniteList.story = {
+  name: 'Formatted Message Infinite List'
+};
```
wootsbot, line 136:
```suggestion
FormattedMessageInfiniteList.story = {
  name: 'FormattedMessageInfiniteList'
};
```

src/components/ui/Chat/MessageList/formatMessageList.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+  return (
+    <Flex css={`background-color: ${bgd};`}>
+      <ul>{formatMessageList(storyMessages, 'abc', null)}</ul>
+    </Flex>
+  );
+};
+
+FormattedMessageList.story = {
+  name: 'Formatted Message List'
+};
```
wootsbot, line 85:
```suggestion
FormattedMessageList.story = {
  name: 'FormattedMessageList'
};
```

src/components/ui/Utilities/memoize.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export const memoize = (fn: Function) => {
+  const map = {};
+  return (...args: any) => {
+    let n = args[0];
+    if (n in map) {
```
p-foucht, line 8: Seems a little strange to accept multiple args but then memoize based on only the first arg. This likely will lead to bugs in any function memoizes but takes in multiple args

We either need to restrict to a single argument, or add support for mluti args (such as multi-arg index, or build the cache around the # of arguments passed)

p-foucht (approved):

src/components/ui/Chat/ChatBubble/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+    content: string;
+    /* This allows for custom elements to be rendered into the optional popover menu for each message, such as an edit button. */
+    actions?: ReactNode | ReactNode[];
+    /* Determines if the name should be shown or not. */
+    showName?: boolean;
+    /* Adds the bubble tail style to a message. */
+    showTail?: boolean;
+    /* Determines if the message has been removed after being sent. */
+    redacted?: boolean;
+    /* Detmermines if the message has been edited after being sent. */
```
nicklera1, line 32: misspelled

bradmanAWS: Thanks, fixed and pushed

src/components/ui/Utilities/formatDate.ts
```diff
@@ -0,0 +1,28 @@
...
+import memoize from 'fast-memoize';
+
+export interface DateOptions {
+  weekday?: string;
+  year?: string;
+  month?: string;
+  day?: string;
+}
+
+const formatDateUnmemoized = (dateStr: string, locale: string | undefined, dateOptions?: DateOptions, todayText?: string, yesterdayText?: string) => {
```
nicklera1, line 13: if passed local is 'undefined' then calling methods on Date object will throw exeption!

bradmanAWS: It works as `undefined`. In fact, that is how we are calling it on the DateHeader story.

nicklera1 (approved):

PR #230, opened 2020-10-07
Use the 'children' prop in ControlBarItem to allow for custom PopOver…

davejlane:
… content

**Description of changes:**
The current props for a `ControlBarItem` provide a way to add a simple `PopOver` menu. This change allows the developer to optionally use the `children` prop to allow for any combination of components to exist in the `PopOver`. I also added a prop so that the placement of the `PopOver` can be controlled.

**Testing**

2. How did you test these changes? unit tests, storybook stories


bradmanAWS (approved):

src/components/ui/ControlBar/ControlBar.stories.tsx
```diff
@@ -82,7 +87,37 @@ export const ControlBarForDocs = () => {
...
+          as="button"
+          onClick={() => console.log('clicked')}
+          children={<span>This is more test content</span>}
+        />
+        <PopOverSubMenu text="This is a submenu">
+          <PopOverItem
+            as="button"
+            onClick={() => console.log('clicked')}
+            children={<span>This is also a submenu component</span>}
+          />
```
p-foucht, line 108: Nit - I know this is just a story, but can we use the implicit form of children instead of explicit?

```
<PopoverItem>
  <span>This is a submenu</span>
</PopoverItem>
```

src/components/ui/ControlBar/ControlBarItem.tsx
```diff
@@ -22,8 +22,12 @@ export interface ControlBarButtonProps
   label: string;
   /** The items to render in a popover menu. When passed, the button will render an arrow to open or close a popover menu. Refer to [PopOverItem](/?path=/docs/ui-components-popover--basic-pop-over-menu) */
   popOver?: PopOverItemProps[] | null;
+  /** Defines the placement of PopOver menu. */
+  popOverPlacement?: Placement;
   /**  Apply this prop to receive visual feedback that the button is 'active' */
   isSelected?: boolean;
+  /** Use children to define an alternative to popOver prop with a custom set of elements to be rendered into the popover */
+  children?: ReactNode | ReactNode[];
```
p-foucht, line 30: Should we just have this interface extend `React.FC` instead of explicitly defining children?

Or was this so we could explicitly call it out in the documentation?

davejlane: It was the latter. I could add something to that effect to the documentation, but my concern is that it'll make it more confusing.

src/components/ui/ControlBar/ControlBar.stories.tsx
davejlane: Fair enough, will update.

src/components/ui/ControlBar/ControlBarItem.tsx
p-foucht: Yeah that makes sense, that sounds good to me

p-foucht (approved):

PR #235, opened 2020-10-09
Fixed bug so user leaves a meeting when the use back button or naviga…

bradmanAWS:
…te to home

**Description of changes:**
This adds detection when a user navigates away from the /meeting page with the Back button in the browser or with the URL. 

**Testing**
Yes
Joined a meeting as two users. I confirmed a user can no longer show video, send audio, or be seen in the roster when the navigate away from the meeting. 
n/a

### DIFF
--- a/demo/meeting/src/providers/NavigationProvider.tsx
+++ b/demo/meeting/src/providers/NavigationProvider.tsx
@@ -8,6 +8,10 @@ import React, {
   useRef,
   ReactNode
 } from 'react';
+import { useLocation } from 'react-router-dom';
+import { useMeetingManager } from 'amazon-chime-sdk-component-library-react';
+
+import routes from '../constants/routes';
 
 export type NavigationContextType = {
   showNavbar: boolean;
@@ -37,6 +41,17 @@ const NavigationProvider = ({ children }: Props) => {
   const [showRoster, setShowRoster] = useState(() => isDesktop());
   const [showMetrics, setShowMetrics] = useState(false);
   const isDesktopView = useRef(isDesktop());
+
+  const location = useLocation();
+  const meetingManager = useMeetingManager();
+
+  useEffect(() => {
+    if (location.pathname.includes(routes.MEETING)) {
+      return () => { 
+        meetingManager.leave();
+      }
+    }
+  }, [location.pathname])
 
   useEffect(() => {
     const handler = () => {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Rename icon (ContentShare > ScreenShare) to fix conflicting names
 - Remove unused import in Select
 - Temporarily removed ChannelList component snapshot test
+- [Demo] Fixed so a user leaves a meeting when they press the back button in the browser or navigate back to home
 
 ### Added
 


demo/meeting/src/providers/NavigationProvider.tsx
```diff
@@ -38,6 +42,20 @@ const NavigationProvider = ({ children }: Props) => {
...
+    const handleBackOrForwardClick = (e: any) => {
+      history.push(routes.HOME);
+      meetingManager.leave();
+    }
+    window.onpopstate = (e: any) => handleBackOrForwardClick(e);
+  }, []);
+
+  if (!window.location.href.includes(routes.MEETING || routes.DEVICE)) {
+    meetingManager.leave()
+  };
```
p-foucht, line 57: I think we can make this a little simpler if we use the `useLocation` hook. Really we just want to do the following whenever the location changes
- check if we're on the home page
- decide whether or not to clear the meeting

bradmanAWS: I ran into a couple of problems when trying that way originally. Clearing the meeting in `AppState` doesn't actually cause the user to leave a meeting. Even when the meeting is cleared, the user can still be seen and/or heard in the meeting. We need to call `leave()` from the MeetingManager as far as I can tell. 

Also, when a user presses the back button in a meeting, they don't go to the home screen, they go to the device setup screen. We can't simple see if they are in this screen and clear the meeting, because on the way into a meeting they visit this screen and need the meeting data. Does that make sense?

p-foucht: Right, I meant end/clear the meeting session (calling leave or w/e the API is on MeetingManager)

Yeah, that makes sense. I think you can still achieve that with the `useLocation` hook

```
useEffect(() => {
  if (location.pathName === "MEETING" or w/e) {
  return () => { 
  // Our route has changed from meeting to something else, 
  // so we can do our logic here like stopping the meeting
}
}, [location])
```

Or we could just not stop the meeting on the device screen, since I think technically it "should" work even if you click back since it has to do with selecting devices and quality and such. Mainly I think it should be cleared on the home screen

bradmanAWS: I'm assuming in your snippet above it' supposed to be `if (location.pathName !== "MEETING" or w/e) {`???

I can switch it to use useLocation for sure, but I don't think the user clicking back, going to the device setup screen, and being in the meeting (including sending video and audio) is a good experience. You shouldn't ever send media to a meeting from that screen. I think we do need to hook up the back button to leaving the meeting.

p-foucht: No, the snippet I posted was intentionally `=== "MEETING"

We're returning a cleanup function we want to run when our state has changed FROM meeting to anything else

demo/meeting/src/providers/NavigationProvider.tsx
```diff
@@ -38,6 +42,20 @@ const NavigationProvider = ({ children }: Props) => {
...
+    let inMeeting = false;
+    if (location.pathname.includes(routes.MEETING)) {
+      inMeeting = true;
+    }
+    return () => { 
+      if (inMeeting) {
+        meetingManager.leave();
+      }
+    }
+  }, [location.pathname])
```
p-foucht, line 57: nit - I think you can just conditionally return the cleanup function

```
if (location.pathname.includes(routes.MEETING)) {
  return () =>  meetingManager.leave();
}
```

p-foucht (approved):

PR #246, opened 2020-10-20
Add support for optional icon property on Radio (and RadioGroup)

abey-george:
**Issue #:** 
We need to be able to create a Ratings component with a Like and Dislike actions. This can effectively be achieved if the Radio and RadioGroup controls could accept an optional "icon" property that can be used to pass in a JSX.Element that could then be used as an icon. In this mode, the control would not render the "label" visually. The "label" property should be still be used for a11y purposes.

**Description of changes:**
This PR adds support for an optional "icon" prop, of type Jsx.Element,  on the Radio and RadioGroup controls. 

**Testing**
Yes.

Manually via storybook.

Yes.


src/components/ui/Radio/index.tsx
```diff
@@ -40,17 +47,36 @@ export const Radio: FC<RadioProps> = props => {
...
+          <StyledRadioIcon
+            checked={checked}
+            className="ch-radio"
+            onClick={handleChange}
+            data-testid="styled-radio-icon"
+          >
+            <span className="ch-icon">
+              {icon}
+            </span>
+          </StyledRadioIcon>
```
p-foucht, line 65: LGTM - has the accessibility of this been verified? I'm assuming it's fine since you're adding the aria-label

abey-george: Verified via VoiceOver on the Mac and tabbing through and confirmed that the aria-label is read.

p-foucht (dismissed):

p-foucht (approved):

Issue #253, opened 2020-10-21
allow "ref" attribute support on elements


alpjor:
Along with the "Focusable" Interface suggested in https://github.com/aws/amazon-chime-sdk-component-library-react/issues/252, it would be nice if components supported adding of "ref" attributes so that parent components can call them for focusing and blurring

alpjor:
Also for things like attaching to events via .addEventListener( etc.

p-foucht:
For any others arriving at this ticket, this is an enhancement we have backlogged to add ref support to all elements. 

As an immediate workaround, you can always wrap the component with your own ref to access it

```
// use whatever browser API you prefer to grab the element
console.log(myRef.firstElementChild)

return (
 <div ref={myRef}>
   <SomeComponent />
 </div>
)
```

xuesichao:
Thank you for submitting a feature request. We will track your request via an entry on the [Amazon Chime SDK Project Board](https://github.com/orgs/aws/projects/253) for centralized visibility of the request and its status and close this GitHub issue.
Thank you for your feedback and for contributing to the future of the service.

Issue #252, opened 2020-10-21
new "Focusable" interface


alpjor:
In the design of the Roster lists, it's the RosterCell that's focused with the keyboard navigation, and not the options button, inside the RosterCell. The design requests accessible keyboard navigation of long Roster lists and wants to prevent having to tab through each entry 1-by-1 by supporting tabbing across sections (RosterGroups), and using arrow-key navigation to go through each entry (RosterCell). To support these features I propose a new "Focusable" Interface that can be given to components like RosterCell that gives access to "tabIndex", "onFocus", and "onBlur" methods. This interface can be implemented by other components like Button that also support these features, and we can remove "tabIndex" from BaseProps.

p-foucht:
[PR](https://github.com/aws/amazon-chime-sdk-component-library-react/pull/254)

PR #254, opened 2020-10-21
Fix roster button accessibility bug

p-foucht:
**Issue #:** 
Work in progress PR, just posting to get feedback and for client team visibility

Will update this PR to address issue #252, #253

**Description of changes:**
 Roster was rendering some broken HTML. Updating Popover to accept an alternate prop so you can control the full button rendering

**Testing**

2. How did you test these changes? unit, manually in demo

### DIFF
--- a/src/components/ui/Base/index.ts
+++ b/src/components/ui/Base/index.ts
@@ -10,6 +10,15 @@ export interface BaseProps extends SpaceProps, BaseSdkProps {
   tag?: any;
 }
 
+export interface FocusableProps {
+  /** Optional tab index for keyboard accessibility */
+  tabIndex?: number;
+  /** onFocus event handler */
+  onFocus?: (event: React.FocusEvent<any>) => void;
+  /** onBlur event handler */
+  onBlur?: (event: React.FocusEvent<any>) => void;
+}
+
 export const baseStyles = ({ css }: BaseProps) =>
   css ? `@media { ${css} };` : '';
 export const baseSpacing = (props: BaseProps) => space(props);
--- a/src/components/ui/PopOver/index.tsx
+++ b/src/components/ui/PopOver/index.tsx
@@ -6,7 +6,7 @@ import React, {
   createRef,
   useState,
   HTMLAttributes,
-  useEffect
+  useEffect,
 } from 'react';
 import { Manager, Reference, Popper } from 'react-popper';
 
@@ -32,8 +32,10 @@ export interface PopOverProps extends HTMLAttributes<HTMLSpanElement> {
   isSubMenu?: Boolean;
   /** Defines the placement of PopOver menu. */
   placement?: Placement;
-  /** Defines the function to render button(s). */
-  renderButton: (isActive: boolean) => {};
+  /** Defines the function to render the inner contents of the popover button element */
+  renderButton?: (isActive: boolean) => {};
+  /** Alternative to renderButton, defines the function to render the full popover button element (as opposed to just its contents). This is used if you want full control over the button rendering. The button must forwardRef */
+  renderButtonWrapper?: (isActive: boolean, props: any) => {};
   /** The label used for availability. */
   a11yLabel: string;
   children: any;
@@ -45,6 +47,7 @@ const getFocusableElements = (node: HTMLElement): NodeListOf<HTMLElement> => {
 
 export const PopOver: FC<PopOverProps> = ({
   renderButton,
+  renderButtonWrapper,
   children,
   isSubMenu = false,
   placement = 'bottom-start',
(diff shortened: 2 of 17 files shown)


src/components/ui/PopOver/index.tsx
```diff
@@ -33,7 +33,9 @@ export interface PopOverProps extends HTMLAttributes<HTMLSpanElement> {
   /** Defines the placement of PopOver menu. */
   placement?: Placement;
   /** Defines the function to render button(s). */
-  renderButton: (isActive: boolean) => {};
+  renderButton?: (isActive: boolean) => {};
+  /** Alternative to renderButton, lets you control the HTML button */
+  renderButtonWrapper?: (isActive: boolean, props: any) => {};
```
p-foucht, line 38: Open to alternative naming here. Ideally it would be `renderButtonContents` and `renderButton`, but that would be a breaking change sadly

abey-george: I tried this out, by modifying the story locally to set a IconButton using RenderButtonWrapper and it worked as expected. (I had to remember to pass the props down to the IconButton)

src/components/ui/Roster/RosterGroup.tsx
```diff
@@ -5,18 +5,18 @@ import React from 'react';
 
 import Badge from '../Badge';
 import Flex from '../Flex';
-import { BaseProps } from '../Base';
+import { BaseProps, FocusableProps } from '../Base';
 
 import { StyledGroupWrapper, StyledGroup, StyledTitle } from './Styled';
 
-interface Props extends BaseProps {
+export interface RosterGroupProps extends BaseProps, FocusableProps {
```
alpjor, line 12: FWIW, I don't think I need this. but it doesn't hurt anything to have it.

src/components/ui/Roster/RosterHeader.tsx
```diff
@@ -9,10 +9,10 @@ import SearchInput from '../Input/SearchInput';
 import { Search, Remove } from '../icons';
 import IconButton from '../Button/IconButton';
 import { StyledHeader } from './Styled';
-import { BaseProps } from '../Base';
 import { PopOverMenu } from './PopOverMenu';
+import { BaseProps, FocusableProps } from '../Base';
 
-interface RosterHeaderProps extends BaseProps {
+export interface RosterHeaderProps extends BaseProps, FocusableProps {
```
alpjor, line 15: same here, unneeded, but not harmful

src/components/ui/Roster/index.tsx
```diff
@@ -4,9 +4,11 @@
 import React from 'react';
 
 import { StyledRoster } from './Styled';
-import { BaseProps } from '../Base';
+import { BaseProps, FocusableProps } from '../Base';
 
-export const Roster: React.FC<BaseProps> = ({ children, ...rest }) => {
+export interface RosterProps extends BaseProps, FocusableProps {}
```
alpjor, line 9: same here, unneeded, but not harmful

src/components/ui/Roster/RosterGroup.tsx
p-foucht: I'll just keep in the chance that it's needed later - like you said, shouldn't hurt

bradmanAWS (commented):
Perhaps you could include an example using the `renderButtonWrapper` prop in PopOver.mdx? It's not super obvious what the difference is. Also, since `renderButton` and `renderButtonWrapper` are both optional now, do we need some kind of error handling if neither prop are provided?

src/components/ui/Base/index.ts
```diff
@@ -10,6 +10,15 @@ export interface BaseProps extends SpaceProps, BaseSdkProps {
...
 }
 
+export interface FocusableProps {
+  /** Optional tab index for keyboard accessibility */
+  tabIndex?: number;
+  /** onFocus event handler */
+  onFocus?: (event: React.FocusEvent<any>) => void;
+  /** onBlur event handler */
+  onBlur?: (event: React.FocusEvent<any>) => void;
+}
```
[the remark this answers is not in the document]
p-foucht: - Updated documentation to call out button contents vs button element
- Yeah we do handle that - if you don't pass either you'll get a pretty useless component, but it shouldn't break anything
- I don't think every component needs `FocusableProps`, but if others end up asking for the same functionality we can move it to `BaseProps`

alpjor:
FWIW, this all looks good to me. Looking forward to it!

bradmanAWS (approved):

nicklera1:
Changelog ?

p-foucht:
> Changelog ?

Ah oops, merged before this comment. Shouldn't the build tests failed?

Issue #286, opened 2020-11-16
Feature request - Simulcast configuration


alexander-teemyco:
Amazon Chime SDK JS allows to use Simulcast.

Here is a documentation: https://aws.github.io/amazon-chime-sdk-js/modules/simulcast.html

Please make possible to use Simulcast in Amazon Chime SDK Component Library React.

From simulcast documentation, we understand that we need to enable 2 options: 

- `MeetingSessionConfiguration.enableUnifiedPlanForChromiumBasedBrowsers`
-  `MeetingSessionConfiguration.enableSimulcastForUnifiedPlanChromiumBasedBrowsers`

We found that this configuration is set in Chime React Component library here:

https://github.com/aws/amazon-chime-sdk-component-library-react/blob/f302b2aeced728099e2f84b325f0491b5fc03d7a/src/providers/MeetingProvider/MeetingManager.ts#L169-L173

And that function is called from `meetingManager.join` method:

https://github.com/aws/amazon-chime-sdk-component-library-react/blob/f302b2aeced728099e2f84b325f0491b5fc03d7a/src/providers/MeetingProvider/MeetingManager.ts#L130

So looks like we cannot change configuration to enable Simulcast.

rishiraj824:
Hi there, I'm new here and wanted to contribute (been through contribution.md). Can I take a stab at it?

p-foucht:
Added in the above PR, it will be available in the next NPM release. Let me know if you have any comments on the PR

@rishiraj824 Feel free to post PRs for any issues or improvements that are open 👍 

Resolving, feel free to re-open

Issue #261, opened 2020-10-23
Question: Join meeting without turning on the camera?


LoganArnett:
Currently when I run `meetingManager.join()` the camera is turned on by default. Is it possible to to join without turning on the camera and allow the user to test their camera if they choose to?

devalevenkatesh:
Hello @LoganArnett ,

Yes, it is possible to join a meeting without turning the camera on. Are you talking about the [demo meeting app](https://github.com/aws/amazon-chime-sdk-component-library-react/tree/master/demo/meeting)? If so, the app goes through below steps where you get the experience of joining the meeting with video preview on, on the device setup page.

1. `meetingManager.join()` is called which routes the app to load the [`DeviceSetup`](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/98019825c2e84925ddecc9c046b320d1bd081697/demo/meeting/src/app.tsx#L37) component.
2. The `DeviceSetup` renders below tree:
2.1. Renders `DeviceSelection`.
2.2. Then it renders `CameraDevices`.
2.3. Then it renders `VideoPreview` which uses component library's [`PreviewVideo`](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/98019825c2e84925ddecc9c046b320d1bd081697/demo/meeting/src/components/DeviceSelection/CameraDevices/index.tsx#L30) component. If you check this component, it [chooses the video input](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/98019825c2e84925ddecc9c046b320d1bd081697/src/components/sdk/PreviewVideo/index.tsx#L39) and [starts the video preview for the provided video element](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/98019825c2e84925ddecc9c046b320d1bd081697/src/components/sdk/PreviewVideo/index.tsx#L41). Hence you get the video right away.

To control this you could simply conditionally render the `PreviewVideo` component in the [`CameraDevices`](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/98019825c2e84925ddecc9c046b320d1bd081697/demo/meeting/src/components/DeviceSelection/CameraDevices/index.tsx#L15) component.

> Note: You will need to update your demo app copy.

Example pseudo-code:
```js
const CameraDevices = () => {
  const [showPreview, setShowPreview] = useState(false);

  const startPreview = () => {
    setShowPreview(true);
  }
 return {
  <div>
[...]

PR #312, opened 2020-11-30
Remove needless camera selection when calling MeetingManager.join, al…

p-foucht:
**Issue #:** 
#261, #286

**Description of changes:**
- 261: We are preemptively calling `chooseVideoInputDevice` within `meetingManager.join`, which triggers the camera light to show regardless if we're showing video or not.

The [toggleVideo](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/providers/LocalVideoProvider/index.tsx#L47) function and the [PreviewVideo](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/components/sdk/PreviewVideo/index.tsx#L39) component already call `chooseVideoInputDevice`, so the initial call we make within `join` provides no value and needlessly triggers the camera light

- 286: Allow simulcast config to be set and used for creating a meeting session

**Testing**

- Removed the routing to stay on the home page when starting a meeting and verified no camera light
- Joined a meeting, verified there was no camera light on the meeting page when my video had not been started. Checked that preview and toggling video worked normally
- Passed simulcast config and joined a meeting, worked normally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed
+- Remove needless camera selection when joining meeting
 
 ### Added
 - [Demo] Add Chat Demo app
 
 ### Changed
 - Allow arbitrary to be passed to RosterProvider
+- Allow for simlucast configuration in MeetingProvider
 
 ### Removed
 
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -137,7 +137,7 @@ export { useBandwidthMetrics } from './hooks/sdk/useBandwidthMetrics';
 
 // Providers
 export { NotificationProvider } from './providers/NotificationProvider';
-export { MeetingProvider } from './providers/MeetingProvider';
+export { MeetingProvider, MeetingContext } from './providers/MeetingProvider';
```
devalevenkatesh, line 140: We have not exported the contexts, rather have the `use..` hooks exported that provide the context, so should we export all contexts or none?

p-foucht: Meant to remove this, thanks -- I think there's a use case for exporting the contexts, but not in this PR

devalevenkatesh (approved):

PR #343, opened 2020-12-21
Update ChatBubble and MessageAttachment

HLiu-amazon:
**Description of changes:**
  1.Add timestamp next to senderName in ChatBubble
  2.Use forwardRef for ChatBubbleContainer
  3.Add img to MessageAttachment
  4.Remove content in ChatBubble
  5.Remove showName in ChatBubble
  6.Change senderName to optional in ChatBubble

**Testing**
Yes

Tested manually and unit tests

Yes


src/components/ui/Chat/ChatBubble/index.tsx
```diff
@@ -24,11 +26,15 @@ export interface ChatBubbleProps extends Omit<HTMLAttributes<HTMLDivElement>, 'c
...
   return (
     <StyledChatBubble variant={variant} editable={false} {...rest} data-testid='chat-bubble'>
-      {showName && <div className='ch-sender-name'>{senderName}</div>}
-      <div>{content}</div>
+      {showName &&
+        <div className='ch-header'>
+          <div className='ch-sender-name'>{senderName}</div>
+          {timestamp && <div className='ch-header-timestamp'>{timestamp}</div>}
+        </div>}
+      {content && <div>{content}</div>}
```
bradmanAWS, line 37: I think if this is part of the breaking change we talked about, we should get rid of `content` all together an only use the `children` prop.

tst/components/ui/Chat/ChatBubble/index.test.tsx
```diff
@@ -42,6 +42,22 @@ describe('ChatBubble', () => {
...
 
+  it('should render timestamp in the ChatBubble component', () => {
+    const component = <ChatBubble content='test message' senderName='test user' timestamp='1:30 PM' variant='outgoing' />;
+    const { getByText } = renderWithTheme(lightTheme, component);
+    const timestamp = getByText('1:30 PM');
+
+    expect(timestamp).toBeInTheDocument();
+  });
+
+  it('should not render timestamp in the ChatBubble component if showName is false', () => {
```
bradmanAWS, line 53: I'm not sure I the relationship between `showName` and showing the `timestamp`. I think you should be able to see a timestamp even if in some cases you don't show the name.

src/components/ui/Chat/MessageAttachment/index.tsx
```diff
@@ -22,21 +30,33 @@ export const MessageAttachment: FC<MessageAttachmentProps> = ({
...
 
   return (
-    <StyledMessageAttachmentContent {...props}>
-      <div className="attachment-icon">
-        <Document className="document-icon" width="2rem" height="2rem" />
-      </div>
+    <StyledMessageAttachment {...props}>
+      <StyledMessageAttachmentContent {...props}>
+        <div className="attachment-icon">
+          <Document className="document-icon" width="2rem" height="2rem" />
```
bradmanAWS, line 39: `className`s should start with the `ch-` prefix.

bradmanAWS (commented):
Additionally, do we have design for what read receipts should look like? I think we could take a more flexible approach to read receipts and timestamps being populated in the `ChatBubbleContainer` and `ChatBubble` components.

src/components/ui/Chat/ChatBubble/index.tsx
```diff
@@ -12,10 +12,12 @@ export interface ChatBubbleProps extends Omit<HTMLAttributes<HTMLDivElement>, 'c
...
   /** The name of the user that sent the message. */
   senderName: string;
-  /** The text content of the message. */
-  content: string;
+  /** The timestamp of the message being sent. */
+  timestamp?: string;
   /** Determines if the name should be shown or not. */
   showName?: boolean;
+  /** Determines if timestamp should be shown or not. */
+  showTimestamp?: boolean;
```
bradmanAWS, line 20: I don't think we need this prop. If you don't want to show the timestamp, you can just not pass it.

src/components/ui/Chat/ChatBubble/index.tsx
```diff
@@ -12,10 +12,12 @@ export interface ChatBubbleProps extends Omit<HTMLAttributes<HTMLDivElement>, 'c
   variant: MessageVariant;
   /** The name of the user that sent the message. */
   senderName: string;
-  /** The text content of the message. */
-  content: string;
+  /** The timestamp of the message being sent. */
+  timestamp?: string;
   /** Determines if the name should be shown or not. */
   showName?: boolean;
```
bradmanAWS, line 18: I know you didn't write this code, but like below, I think we can get rid of this prop too.

src/components/ui/Chat/MessageAttachment/index.tsx
```diff
@@ -14,6 +14,14 @@ export interface MessageAttachmentProps
   name: string;
   /** The file URL string to download attachment. */
   downloadUrl: string;
+  /** Determines whether render image of the attachment. */
+  renderImg?: boolean;
+  /** The URL of image. */
+  imgUrl?: string;
+  /** The height of image. */
+  imgHeight?: number;
```
bradmanAWS, line 22: If we are asking for a number here, I think we should specify the unit. Alternatively, we could use styled components for this and pass a string here. Either way.

src/components/ui/Chat/ChatBubble/index.tsx
HLiu-amazon: But we want to hide the timestamp and mouseover the chat bubble will show it

src/components/ui/Chat/ChatBubble/index.tsx
HLiu-amazon: Do you mean we only need `timestamp`?

bradmanAWS: Yeah, just like with the name, I think instead of passing `showTimestamp={false}` you could just not pass a `timestamp` prop.

src/components/ui/Chat/ChatBubble/index.tsx
bradmanAWS: In that use case, we could still pass `null` or `undefined` to hide the name or timestamp.

bradmanAWS (commented):
Looks good to me, but let hold on this until we get a clearer plan on when these breaking changes will be released.

bradmanAWS (approved):

PR #364, opened 2021-01-07
optional button in notifications

davejlane:
**Description of changes:**
There are 3 changes in this PR:
- Add an optional button to accompany the notification message.
- Allow users to optionally override the default icon with a custom element.
- Add an option `children` slot for custom content

**Testing**

2. How did you test these changes? Manually, and with unit tests

3. If you made changes to the component library, have you provided corresponding documentation changes? Yes. The docs are updated as well as more storybook stories.

### DIFF
--- a/tst/components/ui/Notification/index.test.tsx
+++ b/tst/components/ui/Notification/index.test.tsx
@@ -3,15 +3,31 @@
 
 import React from 'react';
 import '@testing-library/jest-dom';
-import { fireEvent } from '@testing-library/dom';
+import { fireEvent, getByText } from '@testing-library/dom';
 
 import lightTheme from '../../../../src/theme/light';
 import { renderWithTheme } from '../../../test-helpers';
 import Notification from '../../../../src/components/ui/Notification';
 import { Severity } from '../../../../src/providers/NotificationProvider';
+import Echo from '../../../../src/components/ui/icons/Echo';
 
 const getNotificationComponent = () => (
   <Notification onClose={jest.fn()} severity={Severity.ERROR} message="Hello" />
+);
+
+const getNotificationWithOptionalProps = () => (
+  <Notification
+    icon={<Echo />} 
+    onClose={jest.fn()}
+    severity={Severity.SUCCESS}
+    message='Hello'
+    buttonProps={{ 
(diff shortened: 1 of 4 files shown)


tst/components/ui/Notification/index.test.tsx
```diff
@@ -3,17 +3,31 @@
...
   <Notification onClose={jest.fn()} severity={Severity.ERROR} message="Hello" />
 );
 
+const getNotificationWithButtonAndIconComponent = () => (
+  <Notification
+    icon={<Echo />} 
+    onClose={jest.fn()}
+    severity={Severity.SUCCESS}
+    message='Hello'
+    buttonProps={{
```
bradmanAWS, line 24: It's hard for me to not thing about having the option of using `children` here instead of a string `message` prop and `buttonProps`. Do you think that's overcomplicating it?

davejlane: I thought about that too. Given the nature of a component such as a notification, I feel like there's a more limited range of ways someone might want to modify it. Essentially we want a message to be delivered in an accessible way. The specific button styling for each notification severity level would potentially be more error-prone to apply with a more generic `children` prop.

That said, we could introduce an optional slot for `children`. I'd imagine that we would want to offer a variety of options for the the user, something like this:

```
<div className="ch-severity-icon" data-testid="severity-icon">
  {icon ? icon : iconMapping[severity]}
</div>
<output className="ch-message" data-testid="message" role={ariaRole}>
  {message}
</output>
{buttonProps && <StyledNotificationButton aria-hidden {...buttonProps}/>}
{children}
{onClose && (
  <StyledCloseIconButton
    label="close"
    icon={<Remove />}
    onClick={onClose}
  />
)}
```

What are your thoughts?

bradmanAWS: I like this design. Strongly encouraging the message to be type string while having children as a bail out seems right.

davejlane: I agree. I'll make the update.

bradmanAWS (approved):

bradmanAWS (approved):

PR #368, opened 2021-01-09
Optional prop to prevent dismissing the Modal component

davejlane:
**Description of changes:**
Sometimes a user flow might require a modal that needs to persist. Although there are a few ways to do that with the current modal props, it's a little convoluted, and requires the user to remember to do the following:

- pass `null` to the `onClose` prop
- set `displayClose` to false in the ModalHeader
- ensure that `ModalButtons` do not set `closesModal` to true

Introducing the `dismissible` prop will allow all of the above to be handled via a single prop. I also cleaned up the modal stories by using a `Heading` component.

**Testing**

2. How did you test these changes? manually and with unit tests

### DIFF
--- a/src/components/ui/Modal/Modal.stories.tsx
+++ b/src/components/ui/Modal/Modal.stories.tsx
@@ -14,6 +14,7 @@ import ModalButton from './ModalButton';
 import ModalButtonGroup from './ModalButtonGroup';
 import PrimaryButton from '../Button/PrimaryButton';
 import ModalDocs from './Modal.mdx';
+import Heading from '../Heading';
 
 export default {
   title: 'UI Components/Modal',
@@ -63,7 +64,7 @@ export const BasicExample = () => {
     <Flex layout="fill-space-centered">
       <div style={{ width: '30rem', textAlign: 'center' }}>
         <div>
-          <h1>Modal basic example</h1>
+          <Heading level={6}>Modal basic example</Heading>
           <PrimaryButton onClick={toggleModal} label="toggle modal" />
           {showModal && (
             <Modal
@@ -110,7 +111,7 @@ export const BasicExample = () => {
   );
 };
 
-BasicExample.story = 'Basic example';
+BasicExample.story = 'Basic Example';
 
 export const largeContent = () => {
   const [showModal, setShowModal] = useState(false);
@@ -120,7 +121,7 @@ export const largeContent = () => {
     <Flex layout="fill-space-centered">
       <div style={{ width: '30rem', textAlign: 'center' }}>
         <div>
-          <h1>Modal with scrollable content</h1>
+          <Heading level={6}>Modal with scrollable content</Heading>
           <PrimaryButton onClick={toggleModal} label="toggle modal" />
 
           {showModal && (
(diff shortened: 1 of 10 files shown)


bradmanAWS (approved):

src/components/ui/Modal/index.tsx
```diff
@@ -24,22 +24,25 @@ export interface ModalProps
...
 }) => {
   const labelID = useUniqueId();
   const contentEl = useRef<HTMLDivElement>(null);
   const mainEl = useRef<HTMLDivElement>(null);
 
-  const modalContext = { onClose, labelID };
+  const modalContext = { onClose, labelID, dismissible };
 
-  useClickOutside(mainEl, onClose);
+  !dismissible && useClickOutside(mainEl, onClose);
```
bradmanAWS, line 45: Shouldn't `useClickOutside` work if `dismissable` is true?

src/components/ui/Modal/index.tsx
```diff
@@ -24,22 +24,25 @@ export interface ModalProps
...
+  /** Optional prop to prevent the modal from closing. */
+  dismissible?: boolean;
 }
 
 export const Modal: FC<ModalProps> = ({
   size = 'md',
   onClose,
   children,
   rootId,
+  dismissible = false,
```
bradmanAWS, line 36: Wouldn't we want `dismissable` to be true to maintain default behavior?

src/components/ui/Modal/ModalHeader.tsx
```diff
@@ -35,7 +35,7 @@ export const ModalHeader: FC<ModalHeaderProps> = ({
         {title}
       </Tag>
 
-      {displayClose && (
+      {(displayClose && !context.dismissible) && (
```
bradmanAWS, line 38: I think we would want to display a close button in the modal *is* dismissable.

src/components/ui/Modal/Modal.stories.tsx
```diff
@@ -269,9 +270,67 @@ export const largeContent = () => {
...
+      <div style={{ width: 'auto', textAlign: 'center' }}>
+        <div>
+          <Heading level={6}>Persistent Modal Example</Heading>
+          <PrimaryButton onClick={toggleModal} label="open modal" />
+          {showModal && (
+            <Modal
+              size={select('size', ['md', 'lg', 'fullscreen'], 'md')}
+              onClose={null}
+              rootId="modal-root"
+              dismissible
```
bradmanAWS, line 291: Why is `dismissable` true here if the example is persistent?

tst/components/ui/Modal/index.test.tsx
```diff
@@ -47,4 +47,30 @@ describe('Modal', () => {
     });
     expect(component.props.onClose).toBeCalledTimes(0);
   });
+
+  it(`will not close the modal using ESC key if "dismissible" prop is set to "true"`, () => {
```
bradmanAWS, line 51: I think this should be the opposite.

src/components/ui/Modal/index.tsx
davejlane: Yep, I think I got everything turned around at the end of last week. Updating now.

src/components/ui/Modal/ModalHeader.tsx
davejlane: Agree, updating.

src/components/ui/Modal/index.tsx
davejlane: Updated

tst/components/ui/Modal/index.test.tsx
davejlane: I agree, it's update in the latest rev

tst/components/ui/Modal/ModalButtonGroup.test.tsx
```diff
@@ -111,4 +111,24 @@ describe('ModalButtonGroup', () => {
     expect(component.props.value.onClose).toHaveBeenCalled();
     expect(getByLabelText('close')).toBeInTheDocument();
   });
+
+  it('does not pass the onClose function when "dismissible" is false', () => {
```
bradmanAWS, line 115: I think `onClose` is still passable, but I think it should be. I don't think the test is needed.

tst/components/ui/Modal/ModalButtonGroup.test.tsx
```diff
@@ -111,4 +111,24 @@ describe('ModalButtonGroup', () => {
...
+        <ModalButtonGroup
+          primaryButtons={[<ModalButton label="close" closesModal />]}
+        />
+      </ModalContext.Provider>
+    );
+    const { getByTestId } = renderWithTheme(
+      lightTheme,
+      component
+    );
+    const modalButton = getByTestId('button');
```
bradmanAWS, line 130: Is this actually finding the button? I don't think it is.

davejlane: If we weren't finding the button, we'd get an error from this line, and we don't. Why do you think that it's not finding the button? The change I'm making will not remove the `ModalButtons` completely, it just doesn't inject the `oncClose` function to their click handler.

tst/components/ui/Modal/ModalButtonGroup.test.tsx
davejlane: I think I could benefit from renaming this. It should be something like "it doesn't inject 'onClose' into the click handler of the button when "dismissible" is false".

We're not removing the buttons here, just not injecting the onClose.

tst/components/ui/Modal/ModalButtonGroup.test.tsx
bradmanAWS: My bad. I missed that this `data-testid` was coming all the way from our base button.

bradmanAWS (approved):

Issue #280, opened 2020-11-09
useActiveSpeakersState is returning Empty Array


pawanl77:
**Describe the bug**
I am using useActiveSpeakersState to fetch the active speakers in the meeting. But it is returning an empty array irrespective of the users activated in the meeting. I have opened camera and audio of the user still they are not showing as active users.

**To Reproduce**
1. Create Meeting
2. Join 2-3 Attendees
3. Open their Camera and fetch activeUsers it will return null array.

**Expected behavior**
It should return the active users in the meeting.

bradmanAWS:
Hi @pawanl77 , can you please send the logs for this event?

64json:
I encountered this bug as well, and I believe it is related to https://github.com/aws/amazon-chime-sdk-js/issues/673#issuecomment-690054697.

Whenever the following happens in `useAttendeeAudioStatus` (e.g., `attendeeId` changes or a component using this hook unmounts), `useActiveSpeakersState` stops tracking of the active speaker.
https://github.com/aws/amazon-chime-sdk-component-library-react/blob/d0040ab0f0fa214dd0afdb19ebdc1d04d6afc732/src/hooks/sdk/useAttendeeAudioStatus.tsx#L34


It seems to be a bug on [aws/amazon-chime-sdk-js](https://github.com/aws/amazon-chime-sdk-js), and I wonder if there's any workaround for the time being.

p-foucht:
@64json Nice catch, that does look like a bug in the SDK. I'll look into getting that fixed

@pawanl77 Assuming the above is the issue you are running into, this would occur when you are unmounting a component relying on `useAttendeeStatus`/`useAttendeeAudioStatus`. Do you ever get a correct list of active speakers?

PR #373, opened 2021-01-13
Update realtimeUnsubscribeFromVolumeIndicator to receive callback param

bradmanAWS:
**Issue #:** 
https://github.com/aws/amazon-chime-sdk-component-library-react/issues/280

**Description of changes:**
- Update the `realtimeUnsubscribeFromVolumeIndicator` method to optionally receive a `callback` param to match the interface updated in the `amazon-chime-sdk-js@2.4`.
- Update the meeting demo application to use the new callback param which no longer causes active speaker detection to stop after closing the roster.
**Testing**
Yes
Updated the demo application and saw active speaker detection working after unsubscribing from volume indicators (closing the roster).
Yes

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "@types/testing-library__react": "^10.0.1",
     "@types/throttle-debounce": "^2.1.0",
     "add": "^2.0.6",
-    "amazon-chime-sdk-js": "^1.22.0",
+    "amazon-chime-sdk-js": "^2.4.x",
     "babel-loader": "^8.1.0",
     "css-loader": "^2.1.1",
     "jest": "^26.6.3",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -109,7 +109,7 @@
     "webpack-cli": "^3.3.2"
   },
   "peerDependencies": {
-    "amazon-chime-sdk-js": "1.x || 2.x",
```
p-foucht, line 112: Hm.. We don't want to force customers to use only 2.4. We really want 2.4 and up, but now that I think about it, I'm unsure if you can specify that. We may have to just do 2.x here

bradmanAWS: Good links! Updated.

p-foucht: `>=2.4.x` would upgrade to the next major version (3.x, 4.x, etc), which could break builders' applications as well

I think we want something like this? (Probably good to double check the link I sent you)

`>=2.4 <3.0`

Maybe @davejlane can confirm if this is the right approach

bradmanAWS: Ok, I'm surprised we would need to worry about 3.0 before it exists, but I can take a deeper look into the standards in those docs.

PR #381, opened 2021-01-19
Navbar refactor

davejlane:
**Description of changes:**
This would be a breaking change as it changes the API contract of the `NavBar` component. 

* Optionally allow the `IconButton` component to render another element via a `badge` prop.
* Refactor the `NavBarItem` component so that it use `IconButton` directly, and accept `children` as content for a `PopOver`. (this is the breaking change)
* Optionally prevent responsive styles for the `NavBar` by introducing a `responsive` prop that developers can use to keep the `NavBar` at a static width throughout different breakpoint sizes. 
* Include optional prop to allow `NavBarItem` labels to be visible at larger breakpoints.

**Testing**

2. How did you test these changes? Manually and with tests


src/components/ui/Navbar/index.tsx
```diff
@@ -7,22 +7,27 @@ import { StyledNavbar } from './Styled';
...
 import { FocusableProps, BaseProps } from '../Base';
 
+
 export interface NavbarProps extends FlexProps, BaseProps, FocusableProps {
   /** Classname to apply custom CSS styles */
   className?: string;
   /** Any react components or HTML elements */
   children?: any;
+  /** optionally render a responsive layout at mobile breakpoints  */
+  responsive?: boolean;
```
p-foucht, line 17: Should this prop be specific to mobile? In the chance we want to responsiveness at other breakpoints?

src/components/ui/Button/IconButton.tsx
```diff
@@ -3,12 +3,19 @@
...
+  badge?: React.ReactNode | React.ReactNode[];
+}
 
 export const IconButton = forwardRef(
   (props: IconButtonProps, ref: React.Ref<HTMLButtonElement>) => (
-    <Button ref={ref} variant="icon" {...props} />
+    <StyledIconButtonWrapper>
+      <Button ref={ref} variant="icon" {...props} />
+      {props.badge && props.badge}
+    </StyledIconButtonWrapper>
```
p-foucht, line 18: Do you foresee any isues with spreading props to the inner component and not the wrapper? Might get some weird behavior with something like `<IconButton className="pos-absolute" />`, since it'll be applied to the button but not the wrapper

Since it's pretty easy to work around, not a big dea but figured I'd bring it up

PR #382, opened 2021-01-19
refactor NavBar

davejlane:
**Description of changes:**

This would be a breaking change as it changes the API contract of the `NavBar` component.

Optionally allow the `IconButton` component to render another element via a `badge` prop.
Refactor the `NavBarItem` component so that it use `IconButton` directly, and accept `children` as content for a `PopOver`. (this is the breaking change)
Optionally prevent responsive styles for the `NavBa`r by introducing a responsive prop that developers can use to keep the `NavBa`r at a static width throughout different breakpoint sizes.
Testing

**Testing**

### DIFF
--- a/src/components/ui/Button/IconButton.tsx
+++ b/src/components/ui/Button/IconButton.tsx
@@ -3,12 +3,19 @@
 
 import React, { forwardRef } from 'react';
 import Button, { ButtonProps } from './';
+import { StyledIconButtonWrapper } from './Styled';
 
-export interface IconButtonProps extends ButtonProps {}
+export interface IconButtonProps extends ButtonProps {
+  /** Render a component to the top right area of the IconButton */
+  badge?: React.ReactNode | React.ReactNode[];
+}
 
 export const IconButton = forwardRef(
   (props: IconButtonProps, ref: React.Ref<HTMLButtonElement>) => (
-    <Button ref={ref} variant="icon" {...props} />
+    <StyledIconButtonWrapper>
+      <Button ref={ref} variant="icon" {...props} />
+      {props.badge}
+    </StyledIconButtonWrapper>
   )
 );
 
--- a/src/components/ui/ControlBar/Styled.tsx
+++ b/src/components/ui/ControlBar/Styled.tsx
@@ -120,9 +120,12 @@ export const StyledControlBarItem = styled.div<StyledControlBarItemProps>`
     }
   `};
 
-  .ch-control-bar-item-iconButton {
+  > :first-child {
     grid-column-start: ${({ layout, popOver, children }) =>
       isVertical(layout) && (popOver || children) ? '2' : '1'};
+  }
+
+  .ch-control-bar-item-iconButton {
 
     .ch-icon {
       width: 1.5rem;
--- a/src/components/ui/Modal/ModalHeader.tsx
+++ b/src/components/ui/Modal/ModalHeader.tsx
@@ -36,12 +36,13 @@ export const ModalHeader: FC<ModalHeaderProps> = ({
       </Tag>
 
       {(displayClose && context?.dismissible) && (
-        <IconButton
-          label="Close"
-          icon={<Remove />}
-          className="ch-close-button"
-          onClick={handleClick}
-        />
+        <span className="ch-close-button">
+          <IconButton
+            label="Close"
+            icon={<Remove />}
+            onClick={handleClick}
+          />
+        </span>
       )}
     </StyledModalHeader>
   );
--- a/src/components/ui/Navbar/index.tsx
+++ b/src/components/ui/Navbar/index.tsx
@@ -7,22 +7,27 @@ import { StyledNavbar } from './Styled';
 import { FlexProps } from '../Flex';
 import { FocusableProps, BaseProps } from '../Base';
 
+
 export interface NavbarProps extends FlexProps, BaseProps, FocusableProps {
   /** Classname to apply custom CSS styles */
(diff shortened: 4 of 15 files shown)


src/components/ui/Button/IconButton.tsx
```diff
@@ -3,12 +3,19 @@
...
+  badge?: React.ReactNode | React.ReactNode[];
+}
 
 export const IconButton = forwardRef(
   (props: IconButtonProps, ref: React.Ref<HTMLButtonElement>) => (
-    <Button ref={ref} variant="icon" {...props} />
+    <StyledIconButtonWrapper>
+      <Button ref={ref} variant="icon" {...props} />
+      {props.badge && props.badge}
+    </StyledIconButtonWrapper>
```
p-foucht, line 18: Do you foresee any isues with spreading props to the inner component and not the wrapper? Might get some weird behavior with something like <IconButton className="pos-absolute" />, since it'll be applied to the button but not the wrapper

Since it's pretty easy to work around, not a big dea but figured I'd bring it up

src/components/ui/Navbar/index.tsx
```diff
@@ -7,22 +7,27 @@ import { StyledNavbar } from './Styled';
...
 import { FocusableProps, BaseProps } from '../Base';
 
+
 export interface NavbarProps extends FlexProps, BaseProps, FocusableProps {
   /** Classname to apply custom CSS styles */
   className?: string;
   /** Any react components or HTML elements */
   children?: any;
+  /** optionally render a responsive layout at mobile breakpoints  */
+  responsive?: boolean;
```
p-foucht, line 17: Should this prop be specific to mobile? In the chance we want to responsiveness at other breakpoints?

src/components/ui/Button/IconButton.tsx
davejlane: It's a good point. In this case the wrapper element is purely for providing a new relative positioning context to allow for the button and the badge component to be positioned correctly. Ideally that never gets overridden.

One of the downsides to putting all of the props in the wrapper is that we've based a lot of our props on native HTML attributes, and if we were to do what you'd suggest, then `StyledIconButtonWrapper` would need to be able to accept `button` attributes, which i suspect will require a decent amount of TS wrangling, and ultimately result in us obfuscating which element is which.

src/components/ui/Navbar/index.tsx
davejlane: I'm completely open to a different name for this prop. I don't think that it should be specific to mobile. The use case that brought up this need was using the component in a setting where we *didn't* want the component to display its' styles at the mobile breakpoint. We wanted the component to keep the same layout throughout.

src/components/ui/Button/IconButton.tsx
p-foucht: Yeah that makes sense to me, and might just make the component more confusing if you try to split the two props up. Sounds good to me

p-foucht (approved):
LGTM!

src/components/ui/Button/IconButton.tsx
```diff
@@ -3,12 +3,19 @@
...
+  /** Render a component to the top right area of the IconButton */
+  badge?: React.ReactNode | React.ReactNode[];
+}
 
 export const IconButton = forwardRef(
   (props: IconButtonProps, ref: React.Ref<HTMLButtonElement>) => (
-    <Button ref={ref} variant="icon" {...props} />
+    <StyledIconButtonWrapper>
+      <Button ref={ref} variant="icon" {...props} />
+      {props.badge && props.badge}
```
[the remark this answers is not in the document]
davejlane: That's true. React is really good at just not rendering `null` expressions. I'll update.

src/components/ui/ControlBar/Styled.tsx
```diff
@@ -120,9 +120,15 @@ export const StyledControlBarItem = styled.div<StyledControlBarItemProps>`
...
-  .ch-control-bar-item-iconButton {
+  > :first-child {
     grid-column-start: ${({ layout, popOver, children }) =>
       isVertical(layout) && (popOver || children) ? '2' : '1'};
+  }
+
+
+  .ch-control-bar-item-iconButton {
+    /* grid-column-start: ${({ layout, popOver, children }) =>
+      isVertical(layout) && (popOver || children) ? '2' : '1'}; */
```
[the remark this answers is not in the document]
davejlane: It should, I was checking something and didn't remove this. I'll update.

src/components/ui/Modal/ModalHeader.tsx
```diff
@@ -36,12 +36,13 @@ export const ModalHeader: FC<ModalHeaderProps> = ({
       </Tag>
 
       {(displayClose && context?.dismissible) && (
-        <IconButton
-          label="Close"
-          icon={<Remove />}
-          className="ch-close-button"
-          onClick={handleClick}
-        />
+        <span className="ch-close-button">
```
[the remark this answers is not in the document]
davejlane: It is. Now that the `IconButton` is wrapped in a span that provides a new positioning context, we need to preserve the outer positioning context so that we can position the close button in the location that we want to for the `ModalHeader`. We can't set these styles to the `IconButton` as it needs to be relatively positioned.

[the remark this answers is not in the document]
bradmanAWS: makes sense

bradmanAWS (approved):

PR #388, opened 2021-01-21
Allow PopOver to receive multiple clicks before closing

bradmanAWS:
**Issue #:** 

**Description of changes:**
See title. This should solve the issues like being unable to select multiple emojis before closing. 

**Testing**
Yes
Unit and storybook
Yes

### DIFF
--- a/src/components/ui/PopOver/index.tsx
+++ b/src/components/ui/PopOver/index.tsx
@@ -44,6 +44,8 @@ export interface PopOverProps
   a11yLabel: string;
   /** The elements that populate the menu */
   children: any;
+  /** Allow the popover to stay open for multiple clicks. */
+  closeOnClick?: boolean;
 }
 
 const getFocusableElements = (node: HTMLElement): NodeListOf<HTMLElement> => {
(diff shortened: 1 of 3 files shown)


src/components/ui/PopOver/index.tsx
```diff
@@ -44,6 +44,8 @@ export interface PopOverProps
   a11yLabel: string;
   /** The elements that populate the menu */
   children: any;
+  /** Unlike the default behavior, this prop allows the popover to stay open for multiple selections. */
+  multiSelect?: boolean;
```
p-foucht, line 48: What are "selections" and should the `PopOver` component have direct knowledge of it?

I realize you mean clicks, but just pointing out that the prop name brings in assumptions about how this component is used. I would instead rename the prop to something that better describes what the prop is doing (which is preventing clicks from closing the popover)

bradmanAWS: Ok, I thought about that. I just thought the prop name "dontImmediatelyClose" was bad but you are right.

davejlane: I think that `multiselect` is a bit of a loaded term. Usually that is related to `select` menus and the ability to choose more than one value. How about `closeOnClick` as a`boolean`?

davejlane (dismissed):

p-foucht (dismissed):

p-foucht (approved):

PR #408, opened 2021-02-03
Set amazon-chime-sdk-js in package.json dependency

ltrung:
**Description of changes:**
- Specify amazon-chime-sdk-js in the component library package.json dependencies.

**Testing**

2. How did you test these changes? Manually run build and the two meeting demo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Refactored NavBarItem to use IconButton directly
 - Change control bar theme opacity to 1
 - [Chat Demo] Rename misspelled file names and imports
+- Set amazon-chime-sdk-js in package.json dependencies
 
 ### Removed
 
--- a/demo/chat/package.json
+++ b/demo/chat/package.json
@@ -3,20 +3,22 @@
   "version": "1.0.1",
   "description": "Demo app for for Amazon Chime Chat SDK",
   "scripts": {
-    "create": "webpack",
-    "start": "npm install && webpack-dev-server",
-    "test": "test"
+    "deps": "cd ../.. && npm run build && yalc publish",
+    "start": "npm run deps && yalc add amazon-chime-sdk-component-library-react && npm install && npm run start:fast",
+    "start:fast": "webpack-dev-server"
   },
   "dependencies": {
     "@aws-amplify/auth": "^3.4.17",
     "@aws-amplify/core": "^3.8.9",
     "@aws-amplify/storage": "^3.3.17",
+    "amazon-chime-sdk-component-library-react": "file:.yalc/amazon-chime-sdk-component-library-react",
     "aws-amplify": "^3.3.14",
     "babel-eslint": "^9.0.0",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
     "react-router-dom": "^5.2.0",
     "styled-components": "^5.2.0",
+    "styled-system": "^5.1.5",
     "uuid": "^8.3.1"
   },
   "devDependencies": {
--- a/demo/meeting/package.json
+++ b/demo/meeting/package.json
@@ -8,7 +8,7 @@
     "@types/webpack-env": "^1.15.1",
     "@typescript-eslint/eslint-plugin": "^2.19.2",
     "@typescript-eslint/parser": "^2.19.2",
-    "amazon-chime-sdk-js": "^2.4.x",
+    "amazon-chime-sdk-component-library-react": "file:.yalc/amazon-chime-sdk-component-library-react",
     "aws-sdk": "^2.731.0",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
@@ -18,11 +18,13 @@
     "throttle-debounce": "^2.3.0"
   },
   "scripts": {
-    "build": "npm install && npm run build:prod",
-    "build:prod": "webpack --config ./webpack.config.js --mode=production --devtool=false",
-    "start:client": "npm install && webpack-dev-server",
+    "deps": "cd ../.. && npm run build && yalc publish",
+    "build:fast": "webpack --config ./webpack.config.js --mode=production --devtool=false",
+    "build": "npm run deps && yalc add amazon-chime-sdk-component-library-react && npm install && npm run build:fast",
+    "start:client": "npm run build:fast && webpack-dev-server",
     "start:backend": "node server.js",
-    "start": "concurrently \"npm run start:client\" \"npm run start:backend\""
+    "start:fast": "concurrently \"npm run start:client\" \"npm run start:backend\"",
+    "start": "npm run deps && yalc add amazon-chime-sdk-component-library-react && npm install && npm run start:fast"
   },
   "devDependencies": {
     "@babel/core": "^7.11.1",
--- a/package.json
+++ b/package.json
@@ -69,8 +69,8 @@
     "@types/styled-system": "^5.1.9",
     "@types/testing-library__react": "^10.0.1",
     "@types/throttle-debounce": "^2.1.0",
+    "amazon-chime-sdk-js": "^2.4.1",
     "add": "^2.0.6",
-    "amazon-chime-sdk-js": "^2.4.x",
     "babel-loader": "^8.1.0",
     "css-loader": "^2.1.1",
     "jest": "^26.6.3",
@@ -104,7 +104,7 @@
     "webpack-cli": "^3.3.2"
   },
   "peerDependencies": {
-    "amazon-chime-sdk-js": "^2.4.0",
+    "amazon-chime-sdk-js": "^2.4.1",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
     "styled-components": "^5.0.0",
--- a/src/components/migrationToV2.stories.mdx
+++ b/src/components/migrationToV2.stories.mdx
@@ -4,7 +4,8 @@
 
 Version 2 of the Amazon Chime React Component Library makes a small number of interface changes, as well as removing some deprecated interfaces.
 
-### IMPORTANT - Version 2 of the library also requires a dependency to version 2.4 or higher of the amazon-chime-sdk-js.
+### IMPORTANT - Version 2 of the library also requires a dependency to version 2.4.1 or higher of the
+amazon-chime-sdk-js.
 
 In many cases you should not need to adjust your application code at all. This will be the case if:
 
@@ -18,10 +19,11 @@ If you do, read on.
 
 ## Installation
 
-Installation involves adjusting your package.json to depend on version 2.x.x.
+Installation involves adjusting your package.json to depend on version 2.x.x and its peer dependency amazon-chime-sdk-js
 
 ```
-npm install --save amazon-chime-sdk-component-library-react@2
+npm uninstall amazon-chime-sdk-js
+npm install --save amazon-chime-sdk-component-library-react@2 amazon-chime-sdk-js
 ```
 
 ## What’s New
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ coverage
 tst/snapshots/**/__diff_output__
 verdaccio_app
 .out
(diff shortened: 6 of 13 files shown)


CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Refactored NavBarItem to use IconButton directly
 - Change control bar theme opacity to 1
 - [Chat Demo] Rename misspelled file names and imports
+- Set amazon-chime-sdk-js in package.json dependencies
```
p-foucht, line 52: I think we'll need to update the docs to call this change out

- [In the introduction](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/components/introduction.stories.mdx#getting-started)
- [In the migration guide](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/components/migrationToV2.stories.mdx#important---version-2-of-the-library-also-requires-a-dependency-to-version-24-or-higher-of-the-amazon-chime-sdk-js)

demo/chat/package.json
```diff
@@ -3,14 +3,15 @@
...
-    "test": "test"
+    "deps": "cd ../.. && npm run build && npm pack",
+    "start": "npm run deps && npm install && webpack-dev-server",
+    "start:fast": "webpack-dev-server"
   },
   "dependencies": {
     "@aws-amplify/auth": "^3.4.17",
     "@aws-amplify/core": "^3.8.9",
     "@aws-amplify/storage": "^3.3.17",
+    "amazon-chime-sdk-component-library-react": "file:../../amazon-chime-sdk-component-library-react-1.6.0.tgz",
```
p-foucht, line 14: Is there a reason to reference the `.tgz` instead of the relative path to the module? Like we started doing with [the SDK repo demo](https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/browser/package.json#L34)

ltrung: When I tested it out, if we just pointed to the parent folder, it copies everything (parent's node_modules, demo files, etc.) so thats why I added npm pack to only copy over the lib folder.

michhyun1: That means we'll have to npm pack this module to use it?

ltrung: yes that is why I added npm run deps to go to the parent folder and packed it.

p-foucht: Does a customer need to know about that? I guess I'm wondering what the difference between this and a customer installing `amazon-chime-sdk-component-library-react` directly through npm. Are customers getting needless extra files?

nicklera1: Ideally we want an easy path for devs to depend on relative path of the components lib better dev experience and have live reloading. One way is to setup different web-pack configs one for release and one for development.

CHANGELOG.md
p-foucht: In the migration guide, could we add a note about builders having to uninstall amazon-chime-sdk-js from their direct deps?

demo/chat/package.json
simmkyu: > Does a customer need to know about that? I guess I'm wondering what the difference between this and a customer installing `amazon-chime-sdk-component-library-react` directly through npm. Are customers getting needless extra files?

@p-foucht  When a customer runs `npm install amazon-chime-sdk-component-library-react`, `node_modules/amazon-chime-sdk-component-library-react` includes only files and folders listed in the `files` field of `package.json` plus some mandatory files like `README.md`.
https://github.com/aws/amazon-chime-sdk-component-library-react/blob/aff5c53a266620825dfd5611ec6c4c272cbc96e2/package.json#L7-L9

Based on what Trung said, the demo's `node_modules/amazon-chime-sdk-component-library-react` includes all folders and files.

@ltrung Am I understanding correctly? Have you checked the NPM documentation about this behavior?

ltrung: @simmkyu Yes, it is documented here https://docs.npmjs.com/cli/v6/configuring-npm/package-json#files. From what I understand, it just copied everything that is not specified in .npmignore

simmkyu: > @simmkyu Yes, it is documented here https://docs.npmjs.com/cli/v6/configuring-npm/package-json#files. From what I understand, it just copied everything that is not specified in .npmignore

In the `package.json#files` doc, I can't find any description that says installing a local NPM folder as a dependency includes all files. I think `package.json#files` has the top priority, so, ideally, it should only include files and folders listed in `package.json#files`.

simmkyu: I just tried running `npm install` after specifying `file:../..`. In `node_modules`, it creates an alias to the folder, so you see all files and folders. What was the issue with `file:../..` again?

package.json
```diff
@@ -38,6 +38,7 @@
     "@popperjs/core": "^2.2.2",
     "@types/puppeteer": "^2.0.1",
     "@types/uuid": "^7.0.3",
+    "amazon-chime-sdk-js": "^2.4.0",
```
simmkyu, line 41: The [AudioVideoProvider Storybook](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/sdk-providers-audiovideoprovider--page) has an example of importing both JS SDK and the component library. A builder imports `DefaultMeetingSession` from JS SDK, instantiates, and passes it into `AudioVideoContext.Provider`.

```
import { AudioVideoContext } from 'amazon-chime-sdk-component-library-react';
import { DefaultMeetingSession } from 'amazon-chime-sdk-js';
```

I think we need to explain how to achieve this example in a new repo setup where a peer dependency is no more there.

CHANGELOG.md
ltrung: Updated both documents. I do not think we need to uninstall over version of amazon-chime-sdk-js

package.json
p-foucht: Does this example change? I thought Richard mentioned that builders can still access the transitive dependency directly

Otherwise, I would expect our current repo demo to break since it relies on the [SDK directly for the TestSound](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/demo/meeting/src/utils/TestSound.tsx#L7)

simmkyu: Cool. We just need an explanation for builders.

src/components/migrationToV2.stories.mdx
```diff
@@ -17,6 +17,12 @@ In many cases you should not need to adjust your application code at all. This w
 If you do, read on.
 
 ## Installation
+If you currently have amazon-chime-sdk-js as a dependency in your package.json, please also update it to v2.4.0 or later
```
simmkyu, line 20: From @p-foucht 
> Does this example change? I thought Richard mentioned that builders can still access the transitive dependency directly

  Otherwise, I would expect our current repo demo to break since it relies on the [SDK directly for the TestSound](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/demo/meeting/src/utils/TestSound.tsx#L7)

Customers do not have to install `amazon-chime-sdk-js` separately. Since the component library depends on `amazon-chime-sdk-js`, they can import it without re-installing the module.

demo/meeting/package.json
```diff
@@ -8,7 +8,7 @@
     "@types/webpack-env": "^1.15.1",
     "@typescript-eslint/eslint-plugin": "^2.19.2",
     "@typescript-eslint/parser": "^2.19.2",
-    "amazon-chime-sdk-js": "^2.4.x",
+    "amazon-chime-sdk-component-library-react": "file:../../amazon-chime-sdk-component-library-react-1.6.0.tgz",
```
richnew10, line 11: Shouldn't this be 1.6.1, at the very least?

But I think you should look at `yalc`.

https://www.npmjs.com/package/yalc

package.json
richnew10: We need to be in one of two positions:

* Customers install their own version of the SDK, it's a peer dependency of the component library, and the versions must match. They should use a broad caret, and we should tighten it in a peer dependency relationship.
* Customers should not install their own version of the SDK, and it should only be provided implicitly by the SDK.

The situation we want to avoid is that the module tree has two different versions of the SDK installed:

```
$ npm ls --all | fgrep amazon
├─┬ amazon-chime-sdk-component-library-react@1.6.1
│ ├── amazon-chime-sdk-js@2.4.0
└─┬ amazon-chime-sdk-js@1.22.0
```

src/components/migrationToV2.stories.mdx
richnew10: @simmkyu is right, but also see above. It is crucial that *if* they have their own dependency, it be the same version. It's possible that we can model this with a `peerDependency` restriction to prevent an incorrect sibling, because the two versions will not be interoperable.

src/components/migrationToV2.stories.mdx
```diff
@@ -17,6 +17,11 @@ In many cases you should not need to adjust your application code at all. This w
 If you do, read on.
 
 ## Installation
+If you currently have amazon-chime-sdk-js as a dependency in your package.json, please uninstall it first
```
richnew10, line 20: I would phrase this as:

> If you currently take a dependency on `amazon-chime-sdk-js`, you must either ensure that your explicit dependency is compatible with the requirements of the component library, or remove the explicit dependency and use the SDK dependency provided by the SDK. You can remove your explicit dependency by running:
>
> ```
> npm uninstall amazon-chime-sdk-js
> ```

package.json
richnew10: @ltrung in your testing were you able to get any conflicting versions with this approach? What happened for the three test cases we discussed?

src/components/migrationToV2.stories.mdx
ltrung: Updated

demo/meeting/package.json
ltrung: Update to yalc

package.json
ltrung: If JS SDK is a direct dependency in component library:
Case 1: If the package consumes component library does not have any existing dependency on JS SDK, installing component library would automatically pull in JS SDK with the correct version (v2.4.1).
Case 2: If the package consumes component library and has an existing dependency on JS SDK ^2.0.0, installing component library would automatically pull in JS SDK v2.4.1.
Case 3: If the package consumes component library and has an existing dependency on JS SDK ^1.x, installing component library would result in 2 different of JS SDK (v2.4.1 and v1.22.0)

If we also specify JS SDK as a peer dependency in component library, NPM gave a warning about peer dependency not met but still install both versions.

If we specify JS SDK as dev and peer dependency in component library, in all three cases, consumer packages would need to make sure the current installed version of JS SDK is compatible with component library. If not, npm would give a warning but does not auto download the compatible JS SDK.

src/components/migrationToV2.stories.mdx
ltrung: Added back peerDependency.

demo/chat/package.json
ltrung: Switch to using yalc to simulate installing a publish package so it is closer to how a real package will use the library

package.json
ltrung: From the test results above, to avoid the case where the module tree contains 2 different version of JS SDK, we will still use devDependency and peerDependency and customers would need to install their own version of JS SDK and make sure it matches the component library.

richnew10: And if we set it up with dev and peer dependencies, the three cases are:

* The customer manually installs a conflicting version. What happens? (I expect: error or warning about a peer dependency, but we need this to fail somehow.)
* The customer does not install anything. What happens? (I expect: build bustage.)
* The customer installs the right version.

ltrung: - If the customer installs the right version, npm would not get any error or warning.
- If the customer does not install anything, npm install would give a warning that the JS SDK version that component library requires is not installed and customers would need to install it themselves.
- If the customer installed a conflicting version of JS SDK, npm would install that version but also give a warning that the JS SDK version that component library requires is not installed and customers would need to install it themselves.

richnew10: OK, so we should do peer dep and explore what options we have for detecting conflicts more robustly here.

richnew10 (approved):

Issue #423, opened 2021-02-16
MicrophoneActivity doesn’t support className


arty-name:
### What are you trying to do?

I am trying to customize the look of `MicrophoneActivity`

### How can the documentation be improved to help your use case?

It could be updated to reflect that `MicrophoneActivity` doesn’t support `className` prop anymore, only `css`

### What documentation have you looked at so far?

https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/sdk-components-microphoneactivity--page

devalevenkatesh:
Hello @arty-name,

Thanks for reporting the issue. Upon checking, I found that the `MicrophoneActivity` renders `MicVolumeIndicator` component which is overriding the passed `className` prop with the `ch-mic-volume-indicator` class name, check [here](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/components/ui/MicVolumeIndicator/index.tsx#L33). We will work on a fix for this, until then, can you please check If you are able to customize using the `ch-mic-volume-indicator` class name.

devalevenkatesh:
We have merged in a fix this issue. Closing it, please feel free to re-open if you still face any issue.

PR #429, opened 2021-02-17
Override default className with prop className

devalevenkatesh:
**Issue #:** 
#423 

**Description of changes:**
- Fixed the `MicrophoneActivity` component's `className` prop overridden by the `MicVolumeIndicator` component `className`.

**Testing**

2. How did you test these changes? Tested by running and testing locally in the demo meeting app.

3. If you made changes to the component library, have you provided corresponding documentation changes? No update required.

### DIFF
--- a/src/components/ui/MicVolumeIndicator/index.tsx
+++ b/src/components/ui/MicVolumeIndicator/index.tsx
@@ -19,18 +19,19 @@ export interface MicVolumeIndicatorProps
 
 export const MicVolumeIndicator = forwardRef(
   (
-    { muted = false, signalStrength, ...rest }: MicVolumeIndicatorProps,
+    { muted = false, signalStrength, className: propClassName, ...rest }: MicVolumeIndicatorProps,
     bgRef: Ref<HTMLDivElement>
   ) => {
     const poorConnection =
       signalStrength !== undefined && signalStrength <= 0.5;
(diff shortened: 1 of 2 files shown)


src/components/ui/MicVolumeIndicator/index.tsx
```diff
@@ -27,10 +27,10 @@ export const MicVolumeIndicator = forwardRef(
 
     return (
       <StyledMicVolumeIndicator
-        {...rest}
```
p-foucht, line 30: Hmm, we might want to pull out classNames prop and combine it with `ch-mic-volume-indicator` explicitly

devalevenkatesh: okok yeah that makes sense. Good call out! So you mean this right?
```
className={`${className} ch-mic-volume-indicator`}
```

p-foucht: You'll want to explicitly check if className is defined, otherwise you'll get `undefined ch-mic-volume-indecator`. But other than that yep looks good

devalevenkatesh: Yes, will check that. Thanks!

p-foucht (approved):

p-foucht (approved):

PR #440, opened 2021-02-26
Update release script to update SDK dependencies on release

michhyun1:
**Description of changes:**
Update release script to update the dependency on the sdk to the latest version every release, and also added a npm script to allow us to run the demo without defaulting to using the yalc installed library every time.

**Testing**
 yes
ran release scipt locally
n/a


demo/meeting/package.json
```diff
@@ -26,7 +26,8 @@
     "start:client": "npm run build:fast && webpack-dev-server",
     "start:backend": "node server.js",
     "start:fast": "concurrently \"npm run start:client\" \"npm run start:backend\"",
-    "start": "npm run deps && yalc add amazon-chime-sdk-component-library-react && npm install && npm run start:fast"
+    "start": "npm run deps && yalc add amazon-chime-sdk-component-library-react && npm install && npm run start:fast",
+    "start:no-yalc": "npm run deps && npm install && npm run start:fast"
```
ltrung, line 30: hmm we should not need to run deps in this case as it just does yalc publish. 
And can we have better name (like start:npm or start:release) instead of start:no-yalc?
Also need to do a build:release command too.

scripts/release.js
```diff
@@ -111,6 +110,29 @@ fs.writeFileSync(
 spawnOrFail('npm', [`version ${versionString} --no-git-tag-version`]);
 logger.log(`Updated package.json version to ${versionString}`);
 
+
+// Udpate meeting demo to the most up to date version of the SDK
+process.chdir(path.join(__dirname, '../demo/meeting'));
```
ltrung, line 115: Need to update chat demo too so this should be a loop. We may also want to update the MCM to verify the chat functionalities.

scripts/release.js
```diff
@@ -111,6 +110,29 @@ fs.writeFileSync(
 spawnOrFail('npm', [`version ${versionString} --no-git-tag-version`]);
```
ltrung, line 110: Should we also update the peer dependency in the main package.json automatically? We can get the latest version with npm show amazon-chime-sdk-js I think.

ltrung (changes requested):

scripts/release.js
devalevenkatesh: So we decided to not bump the JS SDK version but test the react demo with the latest right?

demo/meeting/package.json
michhyun1: The build happens in the start:fast script, so we dont need to. But I'll update the script names.

devalevenkatesh: Would it be better to have a script with `yalc` separately and not edit the existing script triggers? If existing scripts are changed we need to surface that as we may run into any errors as I did yesterday.

scripts/release.js
michhyun1: Okay, I'll update the sdk version in the chat demo also

scripts/release.js
michhyun1: I thought we said we should update it during each release since we only want to support the latest version of the SDK?

devalevenkatesh: Hmm, sorry I am not remembering it correctly, well then in that case we need to update the peer dependency then as Trung mentioned.

demo/meeting/package.json
michhyun1: I'm not changing any existing scripts, just adding start:release and deps:release, which dont use yalc

ltrung: We do have a build command though. What happen if I just want to check whether it build or not?

scripts/release.js
```diff
@@ -111,6 +110,34 @@ fs.writeFileSync(
...
+spawnOrFail('npm', ['install amazon-chime-sdk-js']);
+
+// Update the peer dependency to the most updated version of the SDK
+let demoPackageJson = JSON.parse(fs.readFileSync('package.json', 'utf-8'));
+const updatedSdkVersion = demoPackageJson.dependencies['amazon-chime-sdk-js'];
+
+logger.log(`Installing SDK Version: ${updatedSdkVersion} into the meeting demo, chat demo, and as a peerDependency and devDependency of the react library.`);
+
+process.chdir(path.join(__dirname, '..'));
+let componentsPackageJson = JSON.parse(fs.readFileSync('package.json', 'utf-8'));
```
ltrung, line 131: Also want to update devDependency too

scripts/release.js
```diff
@@ -111,6 +110,34 @@ fs.writeFileSync(
...
+spawnOrFail('npm', ['install amazon-chime-sdk-js']);
+
+// Update the chat demo to the most up to date version of the SDK
+process.chdir(path.join(__dirname, '../demo/chat'));
+spawnOrFail('npm', ['uninstall amazon-chime-sdk-js']);
+spawnOrFail('npm', ['install amazon-chime-sdk-js']);
+
+// Update the peer dependency to the most updated version of the SDK
+let demoPackageJson = JSON.parse(fs.readFileSync('package.json', 'utf-8'));
+const updatedSdkVersion = demoPackageJson.dependencies['amazon-chime-sdk-js'];
```
ltrung, line 126: Does this actually update the version in package.json from ^2.4.1 to ^2.5.0?

demo/chat/package.json
```diff
@@ -4,7 +4,9 @@
   "description": "Demo app for for Amazon Chime Chat SDK",
   "scripts": {
     "deps": "cd ../.. && npm run build && yalc publish",
+    "deps:release": "cd ../.. && npm run build ",
     "start": "npm run deps && yalc add amazon-chime-sdk-component-library-react && npm install && npm run start:fast",
+    "start:release": "npm run deps && yalc add amazon-chime-sdk-component-library-react && npm install && npm run start:fast",
```
devalevenkatesh, line 9: Remove `yalc` call from here, like is the demo/meeting/package.json?

demo/meeting/package.json
michhyun1: I think normally we should be defaulting to use the yalc generated module, but I can see a scenario where we're not doing a release where we might want to build a different version of the react library into the meeting demo. Ive added `build:release' so that we can skip the yalc step

scripts/release.js
michhyun1: The release script will update the peer dependency of the main package.json every release.

scripts/release.js
michhyun1: Yup, we update the devDep in this script. I added it

scripts/release.js
michhyun1: Updated it, we use writeFileSync to write to the file

demo/chat/package.json
michhyun1: Oops, yes this shouldnt use yalc. Thanks

scripts/release.js
```diff
@@ -111,14 +110,41 @@ fs.writeFileSync(
...
+// Update the peer dependency to the most updated version of the SDK
+const updatedSdkVersion = spawnOrFail('npm', [`show amazon-chime-sdk-js version`]).trim();
+
+logger.log(`Installing SDK Version: ${updatedSdkVersion} into the meeting demo, chat demo, and as a peerDependency and devDependency of the react library.`);
+
+let componentsPackageJson = JSON.parse(fs.readFileSync('package.json', 'utf-8'));
+const oldSdkVersion = componentsPackageJson.peerDependencies['amazon-chime-sdk-js'];
+let replacedPeerDependency = fs.readFileSync('package.json').toString();
+// Replace first instance
+replacedPeerDependency = replacedPeerDependency.replace(`"amazon-chime-sdk-js": "${oldSdkVersion}"` , `"amazon-chime-sdk-js": "^${updatedSdkVersion}"`);
```
ltrung, line 122: Can we pass this out to JSON and set devDependency and peerDependency correctly? :D

scripts/release.js
```diff
@@ -111,14 +110,41 @@ fs.writeFileSync(
...
+  'package.json',
+  replacedPeerDependency
+);
+
+logger.log("NPM Installing component library...")
+spawnOrFail('npm', [`install`]);
+
+// Udpate meeting demo to the most up to date version of the SDK
+process.chdir(path.join(__dirname, '../demo/meeting'));
+spawnOrFail('npm', [`install amazon-chime-sdk-js@${updatedSdkVersion}`]);
```
ltrung, line 135: What is the outcome of this command? Does it update the dependency to ^2.5.0 from ^2.4.1?
Also I think we want to run npm run build here, otherwise the deployment script may pick up the old dist.

michhyun1: The deploy script always run npm run build before deploying, so I dont think we need to build here. The updatedSdkVersion should be the newest version that currently exists in NPM.

scripts/release.js
michhyun1: Done, thanks

scripts/release.js
ltrung: Ok did not know that deploy script in component is different from js sdk. 
> What is the outcome of this command? Does it update the dependency to ^2.5.0 from ^2.4.1?

What about this?

michhyun1: > Does it update the dependency to ^2.5.0 from ^2.4.1?

Yes, it will overwrite it.

ltrung (approved):

PR #442, opened 2021-03-01
Housekeeping.

richnew10:
This commit:

* Updates a bunch of components, including React.
* Fixes the incorrect type definitions in date formatting.
* Corrects the detection of 'yesterday' in said code.
* Fixes the Jest configuration to not use the deprecated 'tsConfig' spelling of 'tsconfig'.
* Wraps actions in tests with 'act', as React requests.
* Adds some key properties to 'ModalButtonGroup', which quiets some React warnings.

Tests pass, though I have not manually checked the behavior of any demo apps or storybooks.

---


package.json
```diff
@@ -35,10 +35,8 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@popperjs/core": "^2.2.2",
-    "@types/puppeteer": "^2.0.1",
-    "@types/uuid": "^7.0.3",
```
richnew10, line 39: Types never need to be build dependencies, AIUI, so I moved these to devDependencies.

src/components/ui/Utilities/formatDate.ts
```diff
@@ -4,25 +4,32 @@
...
 export interface DateOptions {
-  weekday?: string;
-  year?: string;
-  month?: string;
-  day?: string;
+  weekday?: 'long' | 'short' | 'narrow';
+  year?: 'numeric' | '2-digit';
+  month?: 'long' | 'short' | 'narrow' | 'numeric' | '2-digit';
+  day?: 'numeric' | '2-digit';
 }
```
richnew10, line 11: Strictly speaking this is a change in our published types. Some customers might have to fix their type definitions.

This is OK for two reasons:

* It only narrows the expected types to what they should have been. Most customers will not be affected.
* Any customer who supplied different values and gets a build error as a result of this change **will have just discovered an actual bug**.

siddhartharao:
Thanks for the cleanup @richnew10.

package.json
```diff
@@ -54,35 +52,37 @@
...
     "postcss-loader": "^3.0.0",
     "postcss-preset-env": "^6.6.0",
     "prompt-sync": "^4.2.0",
     "puppeteer": "^2.1.1",
-    "react": "^16.8.6",
+    "react": "^17.0.1",
     "react-docgen-typescript-loader": "^3.7.2",
-    "react-dom": "^16.13.1",
-    "react-is": "^16.13.1",
+    "react-dom": "^17.0.1",
```
devalevenkatesh, line 84: Should we check running the demo app and storybook since we are bumping react from v16 -> v17 which a major version change?

bradmanAWS (approved):

package.json
richnew10: Yes — or at least, do so before a release. I was doing this as a Sunday favor, so I did not.

richnew10:
@devalevenkatesh @michhyun1 please test this PR when you next prep a release.

devalevenkatesh:
> @devalevenkatesh @michhyun1 please test this PR when you next prep a release.

Sure thing!

ldai1 (dismissed):

README.md
```diff
@@ -25,7 +25,8 @@ Documentation on these components and how to use them can be found on [Amazon Ch
 
 [Amazon Chime SDK for JavaScript Documentation](https://aws.github.io/amazon-chime-sdk-js/#amazon-chime-sdk-for-javascript)
 
-[Migration to V2](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/docs/migration-to-v2--page)
+[Chime SDK UI Component Library Documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/)
+[Migration from V1 to V2](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/docs/migration-to-v2--page)
```
devalevenkatesh, line 29: We have already updated the README, @ldai1 could you please check once if these are now needed?

ldai1 (approved):

README.md
ldai1: You are right. I checked the commit: https://github.com/aws/amazon-chime-sdk-component-library-react/pull/450 and it already removed the [Chime SDK UI Component Library Documentation]. Will remove it in this housekeeping.

devalevenkatesh: Yeah right, Vidya also added some steps around installing and running the library. It would be great to just compare and add those back and modify few links to be on new lines if not. Thanks for taking care of this.

PR #463, opened 2021-03-22
Add a step in release script to create a test react app for dependency check

ldai1:
**Issue #:** 

**Description of changes:**
Add a step in release script to create a test react app, install latest version of component app as dependency and build the test app.

**Testing**

2. How did you test these changes? Manually run the release script with the code change snippet.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -156,6 +156,19 @@ logger.warn(`
 
 shouldContinuePrompt();
 
+// Creates a new React test app, install the component app as dependency and build the test app to ensure no errors or build issues
+process.chdir(path.join(__dirname, '..'));
+spawnOrFail('yalc', ['publish']);
+process.chdir(path.join(__dirname, '../..'));
+spawnOrFail('npx', ['create-react-app dependency-check-app']);
+process.chdir(path.join(__dirname, '../../dependency-check-app'));
+spawnOrFail('yalc', ['add amazon-chime-sdk-component-library-react']);
+spawnOrFail('npm', [`install amazon-chime-sdk-js@${updatedSdkVersion}`]);
+spawnOrFail('npm', ['install --save styled-components styled-system']);
+spawnOrFail('npm', ['run build']);
+process.chdir(path.join(__dirname, '../..'));
+spawnOrFail('rm', ['-rf dependency-check-app']);
+
 if (release_option === '5') {
   spawnOrFail('git', ['push origin HEAD:hotfix -f']);
 } else {
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 const prompt = require('prompt-sync')();
-const { logger, spawnOrFail, process, shouldContinuePrompt } = require('./utilities');
+const { logger, spawnOrFail, process, shouldContinuePrompt, checkWarning } = require('./utilities');
 const fs = require('fs');
 const path = require('path');
 const util = require('util');
@@ -156,16 +156,18 @@ logger.warn(`
 
 shouldContinuePrompt();
 
-// Creates a new React test app, install the component app as dependency and build the test app to ensure no errors or build issues
+// Creates a new React test app, install component and sdk and build the test app to ensure no peer dependency warnings, errors or build issues
+logger.log('Create a new npm package and install amazon-chime-sdk-component-library-react and amazon-chime-sdk-js sdk as dependencies, check if there is peer dependency warning for amazon-chime-sdk-component-library-react');
 process.chdir(path.join(__dirname, '..'));
 spawnOrFail('yalc', ['publish']);
 process.chdir(path.join(__dirname, '../..'));
-spawnOrFail('npx', ['create-react-app dependency-check-app']);
+spawnOrFail('mkdir', ['dependency-check-app']);
 process.chdir(path.join(__dirname, '../../dependency-check-app'));
+spawnOrFail('npm', ['init -y']);
+spawnOrFail('npm', ['install react react-dom']);
+spawnOrFail('npm', [`install amazon-chime-sdk-js@${updatedSdkVersion} styled-components styled-system`]);
 spawnOrFail('yalc', ['add amazon-chime-sdk-component-library-react']);
-spawnOrFail('npm', [`install amazon-chime-sdk-js@${updatedSdkVersion}`]);
-spawnOrFail('npm', ['install --save styled-components styled-system']);
-spawnOrFail('npm', ['run build']);
+checkWarning('npm', ['install -q'], null, 'amazon-chime-sdk-component-library-react');
 process.chdir(path.join(__dirname, '../..'));
 spawnOrFail('rm', ['-rf dependency-check-app']);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added a step in release script to create a test react app, install latest version of component app as dependency and build the test app.
 - Added some key properties to `ModalButtonGroup`, which quiets some React warnings.
 - Added an optional prop `onPopOverClick` to pass a callback function to the `PopOver` UI component. This callback will be called when the `PopOver` UI component is clicked.
 - Added `--no-fail-on-empty-changeset` flag in deploy script to not fail for empty changeset.
--- a/scripts/utilities.js
+++ b/scripts/utilities.js
@@ -36,6 +36,35 @@ const spawnOrFail = (command, args, options) => {
   return output;
 };
 
+// automatically catch peer dependency warnings and error out
+const checkWarning = (command, args, options, dependency) => {
+  options = {
+    ...options,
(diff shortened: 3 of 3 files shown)


scripts/release.js
```diff
@@ -156,6 +156,15 @@ logger.warn(`
 
 shouldContinuePrompt();
 
+// Creates a new React test app, install the component app as dependency and build the test app to ensure no errors or build issues
+process.chdir(path.join(__dirname, '../..'));
+spawnOrFail('npx', ['create-react-app test-app']);
+process.chdir(path.join(__dirname, '../../test-app'));
+spawnOrFail('npm', [`install amazon-chime-sdk-js@${updatedSdkVersion}`]);
```
ltrung, line 163: We need to install the component library too right? And the hard part is how we validate that there is no error. This does not include any validation steps.

scripts/release.js
```diff
@@ -156,6 +156,15 @@ logger.warn(`
 
 shouldContinuePrompt();
 
+// Creates a new React test app, install the component app as dependency and build the test app to ensure no errors or build issues
+process.chdir(path.join(__dirname, '../..'));
+spawnOrFail('npx', ['create-react-app test-app']);
```
michhyun1, line 161: lets call this something else - dependency-check-app

ltrung (commented):
Could you fix the PR description "…st version of component app as dependency and build the test app"

scripts/release.js
```diff
@@ -156,6 +156,19 @@ logger.warn(`
...
 
+// Creates a new React test app, install the component app as dependency and build the test app to ensure no errors or build issues
+process.chdir(path.join(__dirname, '..'));
+spawnOrFail('yalc', ['publish']);
+process.chdir(path.join(__dirname, '../..'));
+spawnOrFail('npx', ['create-react-app dependency-check-app']);
+process.chdir(path.join(__dirname, '../../dependency-check-app'));
+spawnOrFail('yalc', ['add amazon-chime-sdk-component-library-react']);
+spawnOrFail('npm', [`install amazon-chime-sdk-js@${updatedSdkVersion}`]);
+spawnOrFail('npm', ['install --save styled-components styled-system']);
```
ltrung, line 167: This can be done with the previous install of js sdk.

scripts/release.js
```diff
@@ -156,6 +156,19 @@ logger.warn(`
...
+spawnOrFail('npm', ['run build']);
```
ltrung, line 168: Is there a way to validate the result?

scripts/release.js
ldai1: updated. thanks!

scripts/release.js
ldai1: Do you mean verify the package version of component and js sdk? And also verify the build result?

scripts/release.js
ldai1: You mean the previous steps (https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/scripts/release.js#L125) already install the js sdk as peer and dev dependency of component library? We don't need to install the js sdk here?

scripts/release.js
ltrung: I meant check if there were any error during npm install.

scripts/release.js
ltrung: No i just mean you can combine the two commands to npm install --save amazon-chime-sdk-js@${updateSdkVersion}  styled-components styled-system

ldai1: Gotcha. Will combine them into one. Thanks

scripts/release.js
ldai1: so you are asking if there is error with install command, what will be the behavior and how do we know if there are errors right? I have tested that if installed version is not available, the install step will be terminated with errors.

scripts/release.js
```diff
@@ -156,6 +156,19 @@ logger.warn(`
 
 shouldContinuePrompt();
```
michhyun1, line 158: Can we add some logging here so that when we run this we can see whats happening

scripts/release.js
```diff
@@ -156,6 +156,19 @@ logger.warn(`
...
 
+// Creates a new React test app, install the component app as dependency and build the test app to ensure no errors or build issues
+process.chdir(path.join(__dirname, '..'));
+spawnOrFail('yalc', ['publish']);
+process.chdir(path.join(__dirname, '../..'));
+spawnOrFail('npx', ['create-react-app dependency-check-app']);
+process.chdir(path.join(__dirname, '../../dependency-check-app'));
+spawnOrFail('yalc', ['add amazon-chime-sdk-component-library-react']);
+checkWarning('npm', [`install amazon-chime-sdk-js@1.22.0 styled-components styled-system -q`], null, "amazon-chime-sdk-component-library-react");
+// checkWarning('npm', [`install amazon-chime-sdk-component-library-react@2.2.0 -q &> log.txt`]);
```
michhyun1, line 167: Is there a reason why we install the 1.22 SDK version?

ldai1: Oh so this commit is only for testing purpose. What I am trying to do here is install the latest version of component library and the earliest version of js sdk and use the method I add in the utilities to catch the warnings. But problem here is if we use yalc instead of npm to install the component as dependency, there will be no warnings for the incompatible peer dependency. So I upload the changes here for others to try and to brainstorm solutions

scripts/release.js
```diff
@@ -156,6 +156,19 @@ logger.warn(`
 
 shouldContinuePrompt();
 
+// Creates a new React test app, install the component app as dependency and build the test app to ensure no errors or build issues
+process.chdir(path.join(__dirname, '..'));
+spawnOrFail('yalc', ['publish']);
+process.chdir(path.join(__dirname, '../..'));
+spawnOrFail('npx', ['create-react-app dependency-check-app']);
```
ltrung, line 163: Could we switch to use 'npm init -y' to create a dummy package? create-react-app took a while to complete.

scripts/release.js
```diff
@@ -156,6 +156,19 @@ logger.warn(`
 
 shouldContinuePrompt();
 
+// Creates a new React test app, install the component app as dependency and build the test app to ensure no errors or build issues
+process.chdir(path.join(__dirname, '..'));
+spawnOrFail('yalc', ['publish']);
```
ltrung, line 161: Make sure to verify the content of yalc publish. You need to run npm install && npm run build before yalc publish otherwise it would be a empty package. We may already do this above but just make sure.

scripts/release.js
```diff
@@ -156,6 +156,19 @@ logger.warn(`
...
+process.chdir(path.join(__dirname, '../..'));
+spawnOrFail('npx', ['create-react-app dependency-check-app']);
+process.chdir(path.join(__dirname, '../../dependency-check-app'));
+spawnOrFail('yalc', ['add amazon-chime-sdk-component-library-react']);
```
ltrung, line 165: This does not actually install the package so you need to manually call npm install amazon-chime-sdk-component-library-react

scripts/release.js
ldai1: yes, added in the lasted commit.

scripts/release.js
ldai1: yes, added in the latest commit.

scripts/release.js
ldai1: Will do. Thanks for pointing this out.

scripts/release.js
ldai1: Can I just do a "npm install" ? It think it will work in the same way,

scripts/release.js
```diff
@@ -156,6 +156,21 @@ logger.warn(`
 
 shouldContinuePrompt();
 
+// Creates a new React test app, install component and sdk and build the test app to ensure no peer dependency warnings, errors or build issues
+logger.log('Create a new npm package and installl component and js sdk as dependencies, check if there is peer dependency warning for component library');
```
michhyun1, line 160: nit: install typo

ldai1: Good catch. Corrected in the lasted commit.

scripts/release.js
michhyun1: doing 'npm install amazon-chime-sdk-component-library-react' would try to install it from NPM right? I think once you add it, you can just do npm install.

ldai1: yes, in my pr, I use npm install

scripts/release.js
```diff
@@ -156,6 +156,21 @@ logger.warn(`
 
 shouldContinuePrompt();
 
+// Creates a new React test app, install component and sdk and build the test app to ensure no peer dependency warnings, errors or build issues
+logger.log('Create a new npm package and install component and js sdk as dependencies, check if there is peer dependency warning for component library');
```
anuranduttaroy, line 160: Would it be better to write down the full name of the repositories? e.g

```
logger.log(
    'Create a new NPM package and install amazon-chime-sdk-component-library-react 
    and amazon-chime-sdk-js as dependencies, check if there is peer dependency warning for 
    amazon-chime-sdk-component-library-react'
);
```

ldai1: I like the idea. will add. Thanks!

ltrung (approved):

Issue #445, opened 2021-03-05
Tooltips dont work


OriginLive:
**Describe the bug**
Using tooltipable items or wrapping a component itself does not actually show the tooltip.

Even worse, ControlBarItem was renamed to ControlBarButton, but the filename and documentation still points to ControlBarItem

src/components/ui/ControlBar/ControlBarItem.tsx
https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/ui-components-controlbar--control-bar

If you click on the Documentation button above, you can see it says
"ControlBars support tooltips on any buttons by adding a data-tooltip property. Adding a data-tooltip-position property controls the tooltip position".

If you wrap the abovementioned item, you already have a label prop which prevents you from using the tooltip.

Nor is there any documentation on using Tooltips anywhere in the storybook.

**To Reproduce**
1. Import any element and wrap it with WithTooltip.
2. const YourTooltipElement = WithTooltip(YourElement)
2. Apply label and/or other props
3. See that there is no tooltip

**Expected behavior**
Expecting a tooltip to show,
trying to implement a tooltip on ControlBarButton (ex Item) in ControlBar

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser latest chrome
 - "amazon-chime-sdk-component-library-react": "^2.1.0",
 - "amazon-chime-sdk-js": "^2.5.0",

**Additional context**
Please update your documentation, typescript definition of files and whether they're tooltippable or not and fix the tooltipping, thanks!.

ltrung:
@OriginLive Thanks for reporting this. I have been able to reproduce the tooltip does not show in ControlBarButton. However, could you clarify what you meant by the following:
> If you wrap the abovementioned item, you already have a label prop which prevents you from using the tooltip.

OriginLive:
Hi,

so for ControlBarButton

it has the following props:
```
export const ControlBarButton: FC<ControlBarButtonProps> = ({
  icon,
  onClick,
  label,
  isSelected = false,
  popOver = null,
  popOverPlacement,
  popOverLabel,
  children,
  ...rest
}) => {
```

while the wrapper for tooltip has the following props
```
export interface ToolTipProps {
  label: string;
  tooltipPosition?: ToolTipPositionType;
}
```

both have a prop named label and you can't change the label on the tooltip without affecting what's written bellow the controllbar item, which is not desirable!

[...]

PR #464, opened 2021-03-22
Add docs, minor flexibility to WithTooltip HOC

p-foucht:
**Issue #:** 
#445 

**Description of changes:**
The `WithTooltip` has a lot of flaws. The HOC may share the label prop with other components, and other components internally relying on it don't pass the container ID to the HOC.

This change...
- adds a `tooltipText` prop that can alternatively be used in place of `label`. Label will continue to be used as a fallback to avoid breaking changes
- Logs a warning if no container element is found
- Fixes some `WithTooltip` consumers that didn't allow a way for to pass a container ID
- Adds some docs around `WithTooltip`

**Testing**

Manually, and by syncing the changes to Chime and seeing if there was any breakage. I did not see any

Yeah

### DIFF
--- a/src/components/ui/Chat/ChatBubble/ChatBubbleContainer.tsx
+++ b/src/components/ui/Chat/ChatBubble/ChatBubbleContainer.tsx
@@ -1,7 +1,7 @@
 // Copyright 2020-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import React, { FC, HTMLAttributes, ReactNode, Ref } from 'react';
+import React, { FC, useMemo, HTMLAttributes, ReactNode, Ref } from 'react';
 
 import { BaseProps } from '../../Base';
 import { StyledChatBubbleContainer, StyledChatBubbleInfo } from './Styled';
@@ -10,8 +10,6 @@ import { Dots } from '../../icons';
 import { IconButton } from '../../../..';
 import { IconButtonProps } from '../../Button/IconButton';
 import { Tooltipable, WithTooltip } from '../../WithTooltip';
-
-const IconButtonWithToolTip = WithTooltip(IconButton);
 
 export type Message = {
   /** The displayed text of the message sent. */
@@ -47,9 +45,14 @@ export const ChatBubbleContainer: FC<ChatBubbleContainerProps> = React.forwardRe
     const {
       timestamp,
       actions,
+      tooltipContainerId,
       a11yLabel = 'Open channel options',
       ...rest
     } = props;
+
+    const IconButtonWithToolTip = useMemo(
+      () => WithTooltip(IconButton, tooltipContainerId),
+    [tooltipContainerId]);
 
     const ButtonComponent = !!rest['data-tooltip']
       ? IconButtonWithToolTip
--- a/src/components/ui/ControlBar/ControlBar.mdx
+++ b/src/components/ui/ControlBar/ControlBar.mdx
@@ -2,7 +2,7 @@ import { Preview, Props } from '@storybook/addon-docs/blocks';
 import { ThemeProvider } from 'styled-components';
 
 import ControlBar from '.';
-import ControlBarItem from './ControlBarItem';
+import ControlBarButton from './ControlBarButton';
 import { ControlBarForDocs } from './ControlBar.stories';
 import { lightTheme } from '../../../theme/';
 import { GlobalStyles } from '../../../theme/GlobalStyles';
@@ -13,16 +13,16 @@ The ControlBar component displays a section of controls with icons and labels.
 
 You can place ControlBar into the parent element according to the `layout` prop, e.g. `top`, `bottom`, `right`, `left`, `undocked-horizontal`, and `undocked-vertical`.
 
-The ControlBar component contains one or more `ControlBarItem` components, which can show a label and icon. You can add a dropdown list to each ControlBarItem by setting the [popOver props](/?path=/docs/ui-components-popover--basic-pop-over-menu).
+The ControlBar component contains one or more `ControlBarButton` components, which can show a label and icon. You can add a dropdown list to each ControlBarButton by setting the [popOver props](/?path=/docs/ui-components-popover--basic-pop-over-menu).
 
-ControlBars support tooltips on any buttons by adding a `data-tooltip` property. Adding a `data-tooltip-position` property controls the tooltip position
+`ControlBarButton` support tooltips on any buttons by adding a `data-tooltip` property. Adding a `data-tooltip-position` property controls the tooltip position. The tooltip will render the `label` passed, unless a `tooltipContent` prop is passed.
 
 ## Importing
 
 ```javascript
 import {
   ControlBar,
-  ControlBarItem
+  ControlBarButton
 } from 'amazon-chime-sdk-component-library-react';
 ```
 
@@ -99,12 +99,12 @@ const laptopButtonProps = {
 
 return (
   <ControlBar showLabels layout="left">
-    <ControlBarItem {...microphoneButtonProps} />
-    <ControlBarItem {...volumeButtonProps} />
-    <ControlBarItem {...cameraButtonProps} />
-    <ControlBarItem {...dialButtonProps} />
-    <ControlBarItem {...laptopButtonProps} />
-    <ControlBarItem {...hangUpButtonProps} />
+    <ControlBarButton {...microphoneButtonProps} />
+    <ControlBarButton {...volumeButtonProps} />
+    <ControlBarButton {...cameraButtonProps} />
+    <ControlBarButton {...dialButtonProps} />
+    <ControlBarButton {...laptopButtonProps} />
+    <ControlBarButton {...hangUpButtonProps} />
   </ControlBar>
 );
 ```
@@ -114,5 +114,5 @@ return (
 #### ControlBar Component
 <Props of={ControlBar} />
 
-#### ControlBarItem Component
-<Props of={ControlBarItem} />
+#### ControlBarButton Component
+<Props of={ControlBarButton} />
--- a/src/components/ui/Navbar/Navbar.mdx
+++ b/src/components/ui/Navbar/Navbar.mdx
@@ -14,7 +14,7 @@ import { GlobalStyles } from '../../../theme/GlobalStyles';
 
 The Navbar component displays a vertical menu with navigation icon buttons and labels. Navbar is constructed using the NavbarHeader and NavbarItem components.
 
-Navbars support tooltips on any buttons by adding a `data-tooltip` property. Adding a `data-tooltip-position` property controls the tooltip position
+`NavbarItem` support tooltips by adding a `data-tooltip` property. Adding a `data-tooltip-position` property controls the tooltip position. It will render the `label` as the tooltip, unless a `tooltipContent` prop is passed.
 
 ## Importing
 
--- a/src/components/ui/WithTooltip/WithTooltip.stories.mdx
+++ b/src/components/ui/WithTooltip/WithTooltip.stories.mdx
@@ -1,0 +1,58 @@
+import { WithTooltip } from '.'
+
+<Meta title="UI Components/WithTooltip" />
+
+# WithTooltip
+
+Returns a higher order component (HOC) that renders a tooltip. Make sure to pass an existing element container ID and spread props of the wrapped component.
+
+The `WithTooltip` function takes a component to be wrapped, and an optional container ID used to insert the tooltip into.
+If no container ID is passed, it will fallback to `Tooltip__container`. If no portal is found, it will log a warning.
+
+The wrapped component takes a `tooltipPosition` prop for placing the tooltip, and a label prop for the tooltip text.
+
+```javascript
+  type `tooltipPosition` = 'top' | 'bottom' | 'right' | 'left';
+```
+
+**Please note** - This HOC will eventually be deprecated, as we plan to create a `<Tooltip/>` component in its place. We do not recommend using this HOC.
+
+## Importing
+
+```javascript
+import { WithTooltip } from 'amazon-chime-sdk-component-library-react';
+```
+
+## Example
+
+```jsx
+const MyComponent = (props) => {
+  return <div {...props}> hello world </div>;
+}
+
+const WrappedComponent = WithTooltip(MyComponent, "a-tooltip-container");
+
+// usage
+
+const App = () => (
+  <>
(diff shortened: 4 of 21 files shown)


src/components/ui/WithTooltip/WithTooltip.stories.mdx
```diff
@@ -0,0 +1,46 @@
+import { WithTooltip } from '.'
+
+<Meta title="UI Components/WithTooltip" />
+
+# WithTooltip
+
+Returns an HOC that renders a tooltip. Make sure to pass an existing element container ID and spread props of the wrapped component.
```
ltrung, line 7: stupid question. What is a HOC?

src/components/ui/ControlBar/ControlBarItem.tsx
```diff
@@ -44,16 +42,22 @@ export const ControlBarButton: FC<ControlBarButtonProps> = ({
...
+  const IconButtonWithToolTip = useMemo(
+    () => WithTooltip(IconButton, tooltipContainerId),
+  [tooltipContainerId]);
+
   const ButtonComponent = !!rest['data-tooltip']
     ? IconButtonWithToolTip
     : IconButton;
   const buttonComponentProps = !!rest['data-tooltip-position']
-    ? { tooltipPosition: rest['data-tooltip-position'] }
+    ? { tooltipPosition: rest['data-tooltip-position'], tooltipText }
```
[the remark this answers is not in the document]
p-foucht: Yeah, that's intentional. Some of these components aren't visually displaying text, so they can specify whatever label they want and I don't imagine most builders will run into issues with it

If we get feedback that this isn't the case, we can add them pretty easily. But if it's unecessary, I'd like to avoid it because of this scenario

RosterHeader renders a few components

- `searchLabel` prop - [here](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/components/ui/Roster/RosterHeader.tsx#L150)
- an accessibility label - [here](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/components/ui/Roster/RosterHeader.tsx#L160)
- `closeLabel` - [here](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/components/ui/Roster/RosterHeader.tsx#L167)

Each of these labels would require an additional `labelTooltipText` prop, which just adds to the confusion. So if builders don't run into issues relying on the label field because they aren't rendered visually, I think we should wait

src/components/ui/WithTooltip/WithTooltip.stories.mdx
p-foucht: It's a generally used React term -- essentially a function that takes in a component and returns a new wrapped component with additional functionality

[From the docs](https://reactjs.org/docs/higher-order-components.html#:~:text=A%20higher%2Dorder%20component%20(HOC,React%20for%20reusing%20component%20logic.&text=They%20are%20a%20pattern%20that,and%20returns%20a%20new%20component )

src/components/ui/ControlBar/ControlBarItem.tsx
```diff
@@ -44,16 +42,22 @@ export const ControlBarButton: FC<ControlBarButtonProps> = ({
   popOver = null,
   popOverPlacement,
   popOverLabel,
+  tooltipContainerId,
+  tooltipText,
```
davejlane, line 46: 9 times out of 10, we end up eventually loosening up a prop like this to accept a `ReactNode` along with `string`. It might be worth adding this now.

davejlane (changes requested):
I think that these changes will at least get customers unblocked. I'm wary of adding more code to a component that we don't want to continue using, but as long as we document our intentions well, we can navigate to a less intrusive component.

We should go ahead and make the `toolTipText` prop more flexible, unless I'm missing something there.

src/components/ui/Chat/ChatBubble/ChatBubbleContainer.tsx
```diff
@@ -47,10 +45,15 @@ export const ChatBubbleContainer: FC<ChatBubbleContainerProps> = React.forwardRe
     const {
       timestamp,
       actions,
+      tooltipContainerId,
```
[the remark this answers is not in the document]
p-foucht: Yeah, it's not ideal. Just a workaround, unless we want to enforce a single tooltip ID and rely on documentation. In which case, I could get rid of this prop

src/components/ui/ControlBar/ControlBarItem.tsx
p-foucht: Yeah, good call. I'll update

src/components/ui/ControlBar/ControlBarItem.tsx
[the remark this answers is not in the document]
ltrung: Thanks for the thorough explanation :)

src/components/ui/ControlBar/ControlBarItem.tsx
```diff
@@ -1,7 +1,7 @@
 // Copyright 2020-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import React, { FC, ReactNode } from 'react';
+import React, { FC, ReactNode, useMemo } from 'react';
```
ltrung, line 4: Should we update tooltip section in the storybook for all these components.

p-foucht: The related tooltip props should be showing up in the props table of each component. Do you want an additional section detailing just the tooltip piece?

ltrung: No i meant the existing docs: https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/docs/ui-components-controlbar--control-bar
```
ControlBars support tooltips on any buttons by adding a data-tooltip property. Adding a data-tooltip-position property controls the tooltip position
```
We should remove this and redirect to the new story you wrote below or update it to include passing the containerId nd tooltipText

src/components/ui/WithTooltip/WithTooltip.stories.mdx
davejlane: To clarify, `HOC` is s term commonly used in functional programming, not just React.

p-foucht: I think you mean `HOF` (higher order function). But yeah, HOC isn't specific to React but it is a fairly common term for those working with the framework

src/components/ui/ControlBar/ControlBarItem.tsx
p-foucht: ah didn't see those, I'll update

src/components/ui/Navbar/Navbar.mdx
```diff
@@ -14,7 +14,7 @@ import { GlobalStyles } from '../../../theme/GlobalStyles';
 
 The Navbar component displays a vertical menu with navigation icon buttons and labels. Navbar is constructed using the NavbarHeader and NavbarItem components.
 
-Navbars support tooltips on any buttons by adding a `data-tooltip` property. Adding a `data-tooltip-position` property controls the tooltip position
+Navbars support tooltips on any buttons by adding a `data-tooltip` property. Adding a `data-tooltip-position` property controls the tooltip position. It will render the `label` as the tooltip, unless a `tooltipContent` prop is passed.
```
ltrung, line 17: Navbar does not support tooltipContent right?

p-foucht: Yeah that should be `NavbarItem`, not `Navbar`

src/components/ui/Chat/ChatBubble/ChatBubbleContainer.tsx
```diff
@@ -47,10 +45,15 @@ export const ChatBubbleContainer: FC<ChatBubbleContainerProps> = React.forwardRe
...
       a11yLabel = 'Open channel options',
       ...rest
     } = props;
 
+    const IconButtonWithToolTip = useMemo(
```
devalevenkatesh, line 53: We are repeating `useMemo` internally in the components specifically for the `WithTooltip(component, tooltipContainerId)`. Should we have `useMemoWithToolTip` as a utility function or more generally memoize component passed to `WithToolTip` component itself?

p-foucht: If this was something we would continually reuse, I think it'd be worth it. But since this is a fix until a better solution comes along, I think it's a little overkill. But if you guys prefer, I can split it out

devalevenkatesh: Yeah I agree, seems like an overkill as anyway this is going to be re-worked. Np.

src/components/ui/WithTooltip/index.tsx
```diff
@@ -11,12 +11,17 @@ export type ToolTipPositionType = 'top' | 'bottom' | 'right' | 'left';
 export interface Tooltipable {
   /* Put tooltips on any buttons */
   ['data-tooltip']?: boolean;
-  /* Tooltips locaiton on any buttons */
+  /* Tooltips location on any buttons */
   ['data-tooltip-position']?: ToolTipPositionType;
+  /* The element ID of an existing element to render tooltips into */
+  tooltipContainerId?: string;
+  /* Text content for tooltip. Falls back to the label */
+  tooltipContent?: React.ReactNode;
```
devalevenkatesh, line 19: Since, this is typed as `ReactNode` the description needs to change right? It will not be just Text content only right?

src/components/ui/WithTooltip/index.tsx
```diff
@@ -55,7 +61,19 @@ export const WithTooltip = <P extends object>(
...
+    if (!container) {
+      console.warn(`
+        Attempted to use 'WithTooltip' but could not find container element.
+        Pass a valid element ID or add 'Tooltip__container' ID to existing element
+      `);
+      return;
+    }
+
+    setContainer(container)
+  }, []);
```
devalevenkatesh, line 76: This `useEffect` needs to be dependent on the `container_id` prop right? Like if `container_id` changes, the component should get re-rendered.

src/components/ui/WithTooltip/WithTooltip.stories.mdx
```diff
@@ -0,0 +1,46 @@
...
+```
+
+## Example
+
+```jsx
+const MyComponent = (props) => {
+  return <div {...props}> hello world </div>;
+}
+
+const WrappedComponent = WithTooltip(MyComponent, "a-tooltip-container");
```
ltrung, line 33: Do we want to update this will the tooltip content too?

src/components/ui/Navbar/Navbar.mdx
ltrung: I actually mean NavbarItem does not support tooltipContent right?

src/components/ui/WithTooltip/WithTooltip.stories.mdx
```diff
@@ -0,0 +1,46 @@
+import { WithTooltip } from '.'
+
+<Meta title="UI Components/WithTooltip" />
+
+# WithTooltip
+
+Returns a higher order component (HOC) that renders a tooltip. Make sure to pass an existing element container ID and spread props of the wrapped component.
+
+The `WithTooltip` function takes a component to be wrapped, and an optional container ID used to insert the tooltip into.
```
devalevenkatesh, line 9: Should we explicitly call which components can get wrapped here, since not all support using this component right?

src/components/ui/ControlBar/ControlBar.mdx
```diff
@@ -15,7 +15,7 @@ You can place ControlBar into the parent element according to the `layout` prop,
 
 The ControlBar component contains one or more `ControlBarItem` components, which can show a label and icon. You can add a dropdown list to each ControlBarItem by setting the [popOver props](/?path=/docs/ui-components-popover--basic-pop-over-menu).
```
ltrung, line 16: Should we update ControlBarItem reference in this doc to ControlBarButton?

src/components/ui/Navbar/Navbar.mdx
p-foucht: It does. We're spreading props, so we don't need to explicitly call it out

src/components/ui/WithTooltip/index.tsx
p-foucht: Yep! Thanks for catching that

src/components/ui/WithTooltip/WithTooltip.stories.mdx
p-foucht: I don't think that will be very maintainable, and will quickly get out of sync. The individual components should be repsonsible for that documentation IMO

devalevenkatesh: Okok, yeah makes sense, as long as individual components specify their support for WithToolTip we should be good.

src/components/ui/ControlBar/ControlBar.mdx
p-foucht: It'll make this PR much busier, but sure

devalevenkatesh:
The build fails, I think we need to also update the test file name to `ControlBarButton`:
```
tst/components/ui/ControlBar/ControlBarItem.test.tsx does not have a corresponding source file
Ensure that a test file has a corresponding source file.
For example, the test file (tst/components/ui/ControlBar/ControlBarItem.test.tsx) should test the corresponding source file (src/components/ui/ControlBar/ControlBarItem.tsx)
src/components/ui/ControlBar/ControlBarButton.tsx:1 header does not include correct copyright
```

devalevenkatesh (changes requested):
The build failed. We need to update the test file name to `ControlBarButton` from `ControlBarItem` :)
```
tst/components/ui/ControlBar/ControlBarItem.test.tsx does not have a corresponding source file
Ensure that a test file has a corresponding source file.
For example, the test file (tst/components/ui/ControlBar/ControlBarItem.test.tsx) should test the corresponding source file (src/components/ui/ControlBar/ControlBarItem.tsx)
src/components/ui/ControlBar/ControlBarButton.tsx:1 header does not include correct copyright
```

src/components/ui/ControlBar/ControlBar.mdx
ltrung: Sorry, i just meant to update it in documentation (the mdx file and not everything) :)

ltrung (dismissed):

davejlane (approved):

devalevenkatesh (approved):

PR #473, opened 2021-04-02
Support extra properties attendee response

xuesichao:
**Issue #:** 
Reflect the support for allowing arbitrary to be passed to RosterProvider.
 
**Description of changes:**
Add index signature in AttendeeResponse.

**Testing**
Yes, passed all the test.

Return flatten and nested arbitrary from api.ts in meeting demo, these arbitrary can be access by RosterProvider, and be displayed in RosterCell UI component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added an optional prop `onPopOverClick` to pass a callback function to the `PopOver` UI component. This callback will be called when the `PopOver` UI component is clicked.
 - Added `--no-fail-on-empty-changeset` flag in deploy script to not fail for empty changeset.
 - Add `WithTooltip` docs, warning log if no container is found, and some additional flexibility.
+- Add index signature to `AttendeeResponse` to reflect the support of arbitrary.
 
 ### Changed
 - Bumped react and react-dom version to 17.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
 ### Added
 
-- Add support for optional keys to pass extra data in the `AttendeeResponse`.
+- Added support for optional keys to pass extra data in the `AttendeeResponse`.
+
+### Changed
+
+### Removed
 
 ## [2.2.0] - 2021-03-23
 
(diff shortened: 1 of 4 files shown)


xuesichao:
Do you have any better naming for index signature? Maybe 'other(s)', 'optional', 'extra', 'attribute(s)', e.g.

src/providers/MeetingProvider/types.ts
```diff
@@ -17,6 +17,7 @@ export interface MeetingJoinData {
 
 export interface AttendeeResponse {
   name?: string;
+  [attribute: string]: any;
```
devalevenkatesh, line 20: Could you please check if this also affects the [`RosterAttendeeType`](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/types/index.ts#L48). Because, the `RosterProvider` provides `RosterType` state. This `RosterType` holds mapping from a string `attendeeId` to `RosterAttendeeType`. The extra keys now the `AttendeeResponse` provides may not be set (may be errors out) since the `RosterAttendeeType` does not have those keys. Just making sure if you have tested this?

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added an optional prop `onPopOverClick` to pass a callback function to the `PopOver` UI component. This callback will be called when the `PopOver` UI component is clicked.
 - Added `--no-fail-on-empty-changeset` flag in deploy script to not fail for empty changeset.
 - Add `WithTooltip` docs, warning log if no container is found, and some additional flexibility.
+- Add index signature to `AttendeeResponse` to reflect the support of arbitrary.
```
devalevenkatesh, line 26: Would it be helpful to say: "Add support for optional keys to pass extra data in the AttendeeResponse" ?

We also need to update the documentation wherever the `AttendeeResponse` is mentioned.

xuesichao: Yes, that would be better! I will update it soon.

src/providers/MeetingProvider/types.ts
xuesichao: Yes, with only the change of `AttendeeResponse`, the extra data can be received in `RosterProvider` and be displayed using `RosterCell`, no error is out.

src/providers/MeetingProvider/docs/MeetingManager.stories.mdx
```diff
@@ -84,6 +84,7 @@ For example - you may want to fetch the attendee from the database, or get the n
 
 interface AttendeeResponse {
   name?: string;
+  [attribute: string]: any;
```
michhyun1, line 87: Should this be an optional field?

xuesichao: It is an optional field right now. User could only put `name` to `AttendeeResponse`. 
I do think it would be better to somehow mark it as optional. I tried to add a question mark like this `[attribute: string]?: any;`, but it's invalid to do that. Do you know any way to achieve this?

michhyun1: Actually, this should be fine. I don't think you need that '?' - it should be optional already

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added an optional prop `onPopOverClick` to pass a callback function to the `PopOver` UI component. This callback will be called when the `PopOver` UI component is clicked.
 - Added `--no-fail-on-empty-changeset` flag in deploy script to not fail for empty changeset.
 - Add `WithTooltip` docs, warning log if no container is found, and some additional flexibility.
+- Add support for optional keys to pass extra data in the `AttendeeResponse`.
```
devalevenkatesh, line 26: You need to add an [Unreleased] section and then add this to the new [Added] section up at the top.

xuesichao: Updated the CHANGELOG.

devalevenkatesh (approved):

PR #477, opened 2021-04-09
Set node resolve to browser

ltrung:
**Issue #:** 
The component library depends on uuid package which in turn depends on crypto but rollup did not know how to resolve this dependency (whether it comes from browser or node).
Currently, we use external to not include crypto in the result rollup but it causes uuid to use node implementation of crypto in the bundle file (The code copied from index.esm.js below use randomFillSync which is a method in node).
```
  if (poolPtr > rnds8Pool.length - 16) {
    crypto.randomFillSync(rnds8Pool);
    poolPtr = 0;
  }
```
In webpack version previous to 5, this is fine because they include polyfills for Node but it is no longer the case for Webpack 5.
Using component library with webpack would result in this error:
```
Module not found: Error: Can't resolve 'crypto' in '<your folder>/node_modules/amazon-chime-sdk-component-library-react/lib'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.
```
**Description of changes:**
Set the browser property in node resolve so that it uses the browser alternatives instead.

**Testing**

2. How did you test these changes? Manually build the library and compare the bundle size before and after.
Also verify the meeting and chat demo size and make sure there are no changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- Add browser flag for node resolve in rollup config.
 
 ### Added
 
--- a/demo/meeting/package.json
+++ b/demo/meeting/package.json
@@ -22,7 +22,7 @@
     "deps": "cd ../.. && npm run build && yalc publish",
     "build:fast": "webpack --config ./webpack.config.js --mode=production --devtool=false",
     "build": "npm run deps && yalc add amazon-chime-sdk-component-library-react && npm install && npm run build:fast",
-    "build:release": "npm run deps:release && npm install && npm run build:fast",
+    "build:release": "npm install && npm run build:fast",
     "postbuild": "node check-version.js",
     "start:client": "npm run build:fast && webpack-dev-server",
     "start:backend": "node server.js",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -19,10 +19,11 @@ export default {
       sourcemap: true
     }
   ],
-  external: ['crypto'],
   plugins: [
     peerDepsExternal(),
-    resolve(),
+    resolve({
+        browser: true, // This instructs the plugin to use the "browser" field in a dependency's package.json
+    }),
     commonjs(),
     typescript({ useTsconfigDeclarationDir: true }),
   ]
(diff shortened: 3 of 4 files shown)


rossng:
This fix is quite critical.

rossng:
This should fix the issue I'm experiencing.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- Add browser flag for node resolve in rollup config
```
devalevenkatesh, line 13: Nit: End the changelog with a ".".

devalevenkatesh (approved):

devalevenkatesh (changes requested):

rollup.config.js
```diff
@@ -19,11 +19,13 @@ export default {
       sourcemap: true
     }
   ],
-  external: ['crypto'],
   plugins: [
     peerDepsExternal(),
-    resolve(),
+    resolve({
+        browser: true, // This instructs the plugin to use
+                       // the "browser" property in package.json
```
simmkyu, line 26: Where is the `browser` property?

ltrung: https://github.com/uuidjs/uuid/blob/master/package.json#L35 I think this is the one that causes issues for us (rng module)

simmkyu: How about specifying a dependency?
- The "browser" field in a dependency's package.json

simmkyu (approved):

demo/meeting/package.json
```diff
@@ -22,7 +22,7 @@
     "deps": "cd ../.. && npm run build && yalc publish",
     "build:fast": "webpack --config ./webpack.config.js --mode=production --devtool=false",
     "build": "npm run deps && yalc add amazon-chime-sdk-component-library-react && npm install && npm run build:fast",
-    "build:release": "npm run deps:release && npm install && npm run build:fast",
+    "build:release": "npm install && npm run build:fast",
```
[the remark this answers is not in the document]
ltrung: build release meant to test release version so we do not want to build deps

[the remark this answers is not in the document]
devalevenkatesh: Yeah its for the demo, np!

devalevenkatesh (approved):

PR #478, opened 2021-04-09
Allowing localisation of labels within Microphone, Share Screen and Video Quality components

onepiecejigsaw:
**Issue #:** 

**Description of changes:**
Allow localisation of the following properties:
- Microphone icon
    - tooltip when muted
    - tooltip when unmuted
- Screen Share icon
    - tooltip
- Video Quality dropdown
    - label for unselected video quality

**Testing**
Yes

All the tests passed from the above script, and we manually verified that when we overrode the configuration in a project that uses these components, it had the desired effect.

We've added comments alongside the new properties, which are written consistently with existing comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added optional props to specify icon titles in the `AudioInputControl` and `ContentShareControl` components.
+- Added optional props to specify the dropdown text that shows when no video quality is selected, in the `QualitySelection` component.
+
+### Changed
+
+### Removed
+
+
+## [Unreleased]
+
+### Fixed
+
+### Added
+
 - Added support for optional keys to pass extra data in the `AttendeeResponse`.
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added support for optional keys to pass extra data in the `AttendeeResponse`.
+
 ### Changed
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [2.3.0] - 2021-04-12
+## [Unreleased]
 
 ### Fixed
 
@@ -20,25 +20,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
-
-## [Unreleased]
+## [2.3.0] - 2021-04-12
 
 ### Fixed
 - Add browser flag for node resolve in rollup config.
 
 ### Added
-
-### Changed
-
-### Removed
-
-## [Unreleased]
-
-### Fixed
-
-### Added
-
-- Added support for optional keys to pass extra data in the `AttendeeResponse`.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
 ### Added
 
-- Add support for optional keys to pass extra data in the `AttendeeResponse`.
+- Added support for optional keys to pass extra data in the `AttendeeResponse`.
+
+### Changed
+
+### Removed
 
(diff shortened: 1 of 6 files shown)


src/components/sdk/MeetingControls/AudioInputControl.tsx
```diff
@@ -17,11 +17,17 @@ interface Props {
   muteLabel?: string;
   /** The label that will be shown when microphone is unmuted, it defaults to `Unmute`. */
   unmuteLabel?: string;
+  /** Title attribute for the icon when muted, it defaults to `Muted microphone` */
+  mutedTitle?: string;
```
devalevenkatesh, line 21: Should we have "icon" in the props names here? Like `mutedIconTitle`, `unMutedIconTitle` and so on. This would help builders understand the prop more easily apart from the description.

src/components/sdk/MeetingControls/ContentShareControl.tsx
```diff
@@ -16,12 +16,15 @@ interface Props {
   pauseLabel?: string;
   /** The label that will be shown for unpausing content share button in content share control, it defaults to `Unpause`. */
   unpauseLabel?: string;
+  /** Title attribute for the icon, it defaults to `Screen share`. */
+  title?: string;
```
devalevenkatesh, line 20: Would it be better to rename this prop to either `contentShareIconTitle` or `iconTitle`?

devalevenkatesh (changes requested):
Thanks for your PR. Could you please add a CHANGELOG entry for this change?

src/components/sdk/MeetingControls/AudioInputControl.tsx
onepiecejigsaw: Yep, good suggestion, will make this change

src/components/sdk/MeetingControls/ContentShareControl.tsx
onepiecejigsaw: Yep - I prefer iconTitle, I think - I'll make this change

onepiecejigsaw:
I've updated the CHANGELOG, and made the suggested changes to the prop names. Thanks!

CHANGELOG.md
```diff
@@ -7,13 +7,37 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
devalevenkatesh, line 10: Could you please update this changelog with only one [Unreleased] section? We just released one version so you may have to pull in and provide updated changelog.

src/components/sdk/MeetingControls/AudioInputControl.tsx
```diff
@@ -39,7 +45,7 @@ const AudioInputControl: React.FC<Props> = ({
 
   return (
     <ControlBarButton
-      icon={<Microphone muted={muted} />}
+      icon={<Microphone muted={muted} mutedTitle={mutedIconTitle} unmutedTitle={unmutedIconTitle} />}
```
devalevenkatesh, line 48: Just checking if we do not pass any prop value to these titles here from `AudioInputControl` then are the values in the `Microphone` component `undefined` overriding the default ones? Or should we have default values for the `mutedIconTitle` and `unmutedIconTitle` in the `AudioInputControl`?

devalevenkatesh (changes requested):

CHANGELOG.md
onepiecejigsaw: Yep, have done this now

src/components/sdk/MeetingControls/AudioInputControl.tsx
onepiecejigsaw: If we don't pass any prop values to the `AudioInputControl`, the default ones on `Microphone` work ok. They don't get overridden by the `undefined` ones.

We decided to keep the defaults only at one level, because the default values are very much tied to the fact that it's a Microphone. However, if you'd prefer that we had some defaults set in the `AudioInputControl` as well, I can add them in. What do you think?

devalevenkatesh: Thanks for clarification, if it is not getting override and taking the icon title default then we should be fine.

CHANGELOG.md
```diff
@@ -7,15 +7,26 @@ All notable changes to this project will be documented in this file.
...
+### Removed
+
 ## [2.3.0] - 2021-04-12
 
 ### Fixed
 - Add browser flag for node resolve in rollup config.
 
 ### Added
 
-- Added support for optional keys to pass extra data in the `AttendeeResponse`.
```
devalevenkatesh, line 17: This should not be removed right? Rest all looks good!

devalevenkatesh (changes requested):
Just the changelog fix, rest looks good!

onepiecejigsaw:
Great, thanks! I've updated the CHANGELOG to add that line back in - sorry about that, but should all be sorted now.

devalevenkatesh (approved):

Issue #476, opened 2021-04-08
el.current is null on component unmount


schuettc:
@mcosicaws

**Describe the bug**
After the component has been unmounted, the useFocusIn and useMouseMove hooks will fail.  

```TypeError: el.current is null
useFocusIn/</<
src/hooks/useFocusIn/index.tsx:34

  31 |     if (timeoutRef.current) {
  32 |       clearTimeout(timeoutRef.current);
  33 |     }
> 34 |     el.current.removeEventListener('focusin', onFocusIn);
     | ^  35 |     el.current.removeEventListener('focusout', onFocusOut);
  36 |   };
  37 | }, [el]);
```

**To Reproduce**
1. Deploy React demo and run locally
2. Start meeting
3. Join meeting
4. End meeting

**Expected behavior**
Component unmounts correctly

**Screenshots**

**Desktop (please complete the following information):**
- OS: Mac 11.2.3
- yarn: 1.22.10
- "amazon-chime-sdk-component-library-react": "^2.2.0",
- "amazon-chime-sdk-js": "^2.6.2",
- "react": "^17.0.1",
- "react-dom": "^17.0.1",
- "react-router-dom": "^5.2.0",
- "react-scripts": "4.0.3",

**Additional context**
Possible fix:

if el.current is null, return instead of attempting to removeEventListener or set

```function useFocusIn(el, delay) {
    if (delay === void 0) { delay = 3000; }
    var timeoutRef = React.useRef(null);
    var _a = React.useState(false), isFocused = _a[0], setIsFocused = _a[1];
    React.useEffect(function () {
        if (!el.current) {
            return;
        }
        var onFocusIn = function () {
            clearTimeout(timeoutRef.current);
            **if (!el.current) return;**
            setIsFocused(true);
        };
        var onFocusOut = function () {
            clearTimeout(timeoutRef.current);
            timeoutRef.current = setTimeout(function () {
                **if (!el.current) return;**
                setIsFocused(false);
            }, delay);
        };
        el.current.addEventListener('focusin', onFocusIn);
        el.current.addEventListener('focusout', onFocusOut);
        return function () {
            if (timeoutRef.current) {
                clearTimeout(timeoutRef.current);
            }
            **if (!el.current) return;**
            el.current.removeEventListener('focusin', onFocusIn);
            el.current.removeEventListener('focusout', onFocusOut);
        };
    }, [el]);
    return { isFocused: isFocused };
}

[...]

PR #491, opened 2021-04-27
Check for valid current reference to add or remove event listeners

devalevenkatesh:
**Issue #:** 
#476

**Description of changes:**
- Checks for validity of elements current ref in `useFocusIn` and `useMouseMove` while adding / removing event listeners.
- Removes snapshot testing workflow.

**Testing**

- Run the `demo/meeting` app.
- Join a meeting.
- Go away from the screen until the meeting control (options of mic, speaker etc) disappear.
- Come back in, they should re-appear now.
- Leave the meeting and checked and found no browser console errors related to `useFocusIn`, `useFocusOut` and `useMouseMove`.


src/hooks/useFocusIn/index.tsx
```diff
@@ -24,15 +24,15 @@ export function useFocusIn(el: RefObject<any>, delay = 3000) {
       }, delay);
     };
 
-    el.current.addEventListener('focusin', onFocusIn);
-    el.current.addEventListener('focusout', onFocusOut);
+    el.current?.addEventListener('focusin', onFocusIn);
```
ltrung, line 27: Is this needed? Given that we already return in line 11. I think this is just needed for the removeEventListener

ltrung (commented):
Can you double check the changes in package-lock.json? Seem like a lot of changes so make sure it is not reverting back to v1.

ltrung (commented):
Can you add the steps on how you tested the apps and verify there is no console error log?

src/hooks/useFocusIn/index.tsx
devalevenkatesh: Yeah right not needed thanks for the catch!

devalevenkatesh:
> Can you double check the changes in package-lock.json? Seem like a lot of changes so make sure it is not reverting back to v1.

Yup, have verified they are `v2` changes only.

devalevenkatesh:
> Can you add the steps on how you tested the apps and verify there is no console error log?

Yup done.

ltrung:
> > Can you double check the changes in package-lock.json? Seem like a lot of changes so make sure it is not reverting back to v1.
> 
> Yup, have verified they are `v2` changes only.

Just curious what are those changes? Seem like a lot of changes for a couple line changes.

devalevenkatesh:
> > > Can you double check the changes in package-lock.json? Seem like a lot of changes so make sure it is not reverting back to v1.
> > 
> > 
> > Yup, have verified they are `v2` changes only.
> 
> Just curious what are those changes? Seem like a lot of changes for a couple line changes.

So, the lockfile version is still v2, but I had deleted node_modules and package-lock.json and regenerated them while testing and running the demo meeting app. Hence, I think a lot of dependency updates in the package-lock.json. I have anyways removed that file from the commit.

ltrung (approved):

Issue #314, opened 2020-12-02
Mediastream is still working after using meetingManager.leave


Catastropha:
**Describe the bug**
On Chrome after finishing the meeting with meetingManager.leave Mediastream is still working. It keeps the camera busy so if we do another meeting right after that the camera cannot be accessed.
The solution we came up with is to refresh the page after ending the meeting.

**Expected behavior**
Stop mediastream when doing meetingManager.leave

**Desktop (please complete the following information):**
 - OS: Ubuntu 20
 - Browser: Chrome

bradmanAWS:
Hi @Catastropha `meetingManager.leave()` should stop your video. Can you provide your logs when you are experiencing this issue?

Catastropha:
Logs show no errors

This is the code we use:
```
const endMeeting = async () => {
    if (meetingManager) {
      await meetingManager.leave().then(() => {
        // redirect goes here
      })
    }
  }
```
We tried without async await, with promises. Also tried timeout on redirect so it has time to finish whatever it needs to do.
The camera was still active.
On Firefox works fine. Found it only in Chrome.

Is this helpful?

firdausmaarof:
I have the same issue. The camera is still active even after calling meetingManager.leave()

p-foucht:
Is this something either of you can reproduce in the repo demo? I tried to repo with the following

- [commented out the route change](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/demo/meeting/src/containers/EndMeetingControl/index.tsx#L38) when ending a meeting
- Joined a meeting, turned on camera 
- With the camera light and my video on, I ended the meeting which calls `meetingManager.leave`

But I see my Mac's camera light turned off, indicating it released the mediastream successfully. 

Can you provide any additional details about how you're using the components? Also would help if you could provide info logs [by passing a log level prop to MeetingProvider](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/sdk-providers-meetingprovider--page)

firdausmaarof:
This is how i called the `meetingManager.leave`.

```
import {
  AudioInputControl,
  AudioOutputControl,
  ContentShareControl,
  ControlBar,
  ControlBarButton,
  Phone,
  useMeetingManager,
  VideoInputControl,
} from 'amazon-chime-sdk-component-library-react';
import { useRouter } from 'next/router';
import React, { useState } from 'react';

const MeetingControl = () => {
[...]

PR #494, opened 2021-04-30
Fix videopreview component bug

michhyun1:
**Issue #:** 
https://github.com/aws/amazon-chime-sdk-component-library-react/issues/314

**Description of changes:**
This fixes a bug where the VideoPreview component would not stop the media stream on the video preview video element because a ref was being unset before the unmount method could use the value of the ref.

Instead of using the ref in the unmount method - we just save the ref value locally, and then use the variable when we unmount. 

**Testing**
yes

Join a meeting, verify that the light next to the webcam indicating an active media stream is turned off when we join the meeting. Then, turn the camera on while in the meeting, and verify that the light next to the webcam is turned on. Then, leave the meeting and verify that the light is now turned off. This indicates all media streams are not active. 

yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Add or remove event listener only when `el.current` reference is valid
   in `useFocusIn` and `useMouseMove` hooks.
+- Fix bug related to `PreviewVideo` component not releasing media stream when unmounted.
 
 ### Added
 
--- a/src/components/sdk/PreviewVideo/index.tsx
+++ b/src/components/sdk/PreviewVideo/index.tsx
@@ -28,8 +28,8 @@ export const PreviewVideo: React.FC<BaseSdkProps> = (props) => {
     if (!audioVideo || !selectedDevice || !videoEl.current) {
       return;
     }
-
     let mounted = true;
+    const videoElement = videoEl.current;
 
     async function startPreview() {
       if (!audioVideo) {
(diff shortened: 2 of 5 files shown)


src/components/sdk/PreviewVideo/index.tsx
```diff
@@ -28,8 +28,8 @@ export const PreviewVideo: React.FC<BaseSdkProps> = (props) => {
     if (!audioVideo || !selectedDevice || !videoEl.current) {
       return;
     }
-
     let mounted = true;
+    let videoElement = videoEl.current;
```
ltrung, line 32: Should this be const?

ltrung:
You may want to update to Node 16 with NPM 7. It seems you changes package-lock.json to v1 so the build failed.

ltrung:
>This fixes a bug where the VideoPreview component would not stop the media stream on the video preview video element because a ref was being unset before the unmount method could use the value of the ref.

Can we move this under Description of changes?

src/components/sdk/PreviewVideo/index.tsx
michhyun1: yup,

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Add or remove event listener only when `el.current` reference is valid
   in `useFocusIn` and `useMouseMove` hooks.
+- Fix bug related to PreviewVideo component not releasing media stream when unmounted
```
devalevenkatesh, line 15: Nit: End with a ".".
Can we change PreviewVideo -> `PreviewVideo` ?

devalevenkatesh (commented):
Fix looks good to me.

michhyun1:
> You may want to update to Node 16 with NPM 7. It seems you changes package-lock.json to v1 so the build failed.

done. thanks

ltrung (approved):

Issue #379, opened 2021-01-18
Suppress device permission request


henlamk:
Whenever a user joins a meeting, he is asked to grant permission for microphone and camera. 

Is it possible to suppress this behavior, e. g. when the user is not supposed to use his microphone and webcam during the whole meeting and just to join as a viewer?

richnew10:
This behavior is triggered in order to get device labels:

https://github.com/aws/amazon-chime-sdk-component-library-react/blob/0d84475db0eb74da5acb9f5746a0d77c563c4141/src/providers/MeetingProvider/MeetingManager.ts#L265

I don't see a good approach to controlling that through the component library interface, only through the lower-level SDK interface, but I'll follow up with someone closer to the code to see if they have a suggestion.

p-foucht:
This is a shortcoming of the library, as this isn't something you can easily do. We'll take note of this critique and try to put flexibility/extensibility onto our roadmap

There are a few ways you could try to workaround this, but they are not the most straightfoward or have been thoroughly tested

1. Mock the `getUserMedia` API: `navigator.mediaDevices.getUserMedia = () => true` which will remove the device permission prompt
2. Override the `MeetingManager`'s `join` method ([seen here](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/providers/MeetingProvider/MeetingManager.ts#L129)) with your own implementation, removing the device calls that trigger the `getUserMedia` call in the first place

jrowny:
This is also biting us. We have users who enter as spectators and don't expect to have to enabled audio/video. You could have a simple flag to pass into join that disables the calls to listAndSelectDevices();

xuesichao:
Hi guys, we raise this PR [#499](https://github.com/aws/amazon-chime-sdk-component-library-react/pull/499) to fix this issue. 

We add an optional parameter `isDevicePreviewSkipped` in `meetingManager.join()` and a new function `meetingManager.invokeDevicePermissionRequests()` . 

* To suppress the device permission request, you could set `isDevicePreviewSkipped` to true when user joins a meeting (calling `meetingManager.join()`)
* Later If user wants to grant device permission, you could call `meetingManager.invokeDevicePermissionRequests()` function. If the device permission is unset, the device permission request should pop up

Feel free to try it and give us some feedback!

vidya-mohan:
Resolving as issue fixed and released

PR #499, opened 2021-05-10
Delay device permission request

xuesichao:
**Issue #379 :**

**Description of changes:**
- Add controls to device permission request in `DeviceProvider` and `MeetingManager` to suppress the permission request, and invoke it.
- Add a checkbox in `MeetingForm` to allow use join the meeting  without device preview process.
- Add a `DevicePermissionControl` component, which is used in `MeetingControls`.
- Change the `MeetingControls` to be rendered conditionally based on `devicePermissionStatus`.

**Testing**
Yes
    #### Normal Meeting
    * User joins a meeting without selecting "Join w/o Audio and Video"
    * Device permission prompt pops up for audio and video device
    * Device setup page works as usual and join the meeting
    * Check `MeetingControl` works as usual
    #### View-only Meeting
    * User joins a meeting with selecting "Join w/o Audio and Video"
    * No permission prompt pops up
    * Device setup page is skipped, redirect to Meeting page directly
    * `MeetingControl` display only `DevicePermissionControl` and `EndMeetingControl` like below:
    image

    * Click 'Device' button
    * Device permission prompt for audio and video device pops up
    * Click 'Allow' and `MeetingControl` transforms to the original one like below:
    image
    * Check `MeetingControl` works as usual
1. If you made changes to the component library, have you provided corresponding documentation changes?
Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `sourceId` as an optional parameter in the `toggleContentShare` function to specify which screen to share.
 - [Doc] Add a new page, a "How-tos" section in the story book to show use-cases with sample code.
+- Added optional parameter `deviceLabels: DeviceLabels | DeviceLabelTrigger` in `meetingManager.join()`, and `invokeDeviceProvider(deviceLabels: DeviceLabels)` function to control the device permission request. Builder could pass a `deviceLabels` of type `DeviceLabels` to select the devices from which the browser requests permission when joining the meeting. Builder could also pass a `deviceLabels` of `DeviceLabelTrigger` type, to set their customized `deviceLabelTrigger` which is triggered to get the device info. Builder could call `invokeDeviceProvider(deviceLabels: DeviceLabels)` to trigger the device permission prompts. For example, builder wants to implement a view-only mode and no device permission prompts are triggered during the whole process. Builder could just call `meetingManager.join(DeviceLabels.None)` to join a meeting. Later they trigger the device permission prompts by calling `meetingManager.invokeDeviceProvider(DeviceLabels.AudioAndVideo)` to get the full access to devices.
 
 ### Changed
 
@@ -28,8 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- [Doc] Add documentaion in introduction on how to use `MeetingSessionPostLogger` to post
-  Amazon Chime JS SDK logs.
+- [Doc] Add documentation in introduction on how to use `MeetingSessionPostLogger` to post Amazon Chime JS SDK logs.
 - Add `useMediaStreamMetrics` hook to expose audio, video and bandwidth data.
 
 ### Changed
@@ -40,8 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Add or remove event listener only when `el.current` reference is valid
-  in `useFocusIn` and `useMouseMove` hooks.
+- Add or remove event listener only when `el.current` reference is valid in `useFocusIn` and `useMouseMove` hooks.
 - Fix bug related to `PreviewVideo` component not releasing media stream when unmounted.
 - Correct `IconButton` border in dark theme.
 
@@ -54,8 +53,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Change `package-lock` to V2 to support NPM 7.
-- Update `engines` field in `package.json` to include Node 16
-- Bump `hosted-git-info` version
+- Update `engines` field in `package.json` to include Node 16.
+- Bump `hosted-git-info` version.
 
 ### Removed
 
--- a/demo/meeting/src/containers/DynamicMeetingControls/index.tsx
+++ b/demo/meeting/src/containers/DynamicMeetingControls/index.tsx
@@ -1,0 +1,65 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+import {
+  ControlBar,
+  AudioInputControl,
+  VideoInputControl,
+  ContentShareControl,
+  AudioOutputControl,
+  ControlBarButton,
+  useUserActivityState,
+  Dots,
+  useDevicePermissionStatus,
+  DevicePermissionStatus,
+} from 'amazon-chime-sdk-component-library-react';
+
+import EndMeetingControl from '../EndMeetingControl';
+import { useNavigation } from '../../providers/NavigationProvider';
+import { StyledControls } from './Styled';
+import DevicePermissionControl from '../DevicePermissionControl/DevicePermissionControl';
+import { DeviceLabels } from 'amazon-chime-sdk-component-library-react/lib/types';
+
+const DynamicMeetingControls = () => {
+  const { toggleNavbar, closeRoster, showRoster } = useNavigation();
+  const { isUserActive } = useUserActivityState();
+  const permission = useDevicePermissionStatus();
+
+  const handleToggle = () => {
+    if (showRoster) {
+      closeRoster();
+    }
+
+    toggleNavbar();
+  };
+
+  return (
+    <StyledControls className="controls" active={!!isUserActive}>
+      <ControlBar
+        className="controls-menu"
+        layout="undocked-horizontal"
+        showLabels
+      >
+        <ControlBarButton
+          className="mobile-toggle"
+          icon={<Dots />}
+          onClick={handleToggle}
+          label="Menu"
+        />
+        {permission === DevicePermissionStatus.GRANTED ?
+          <>
+            <AudioInputControl />
+            <VideoInputControl />
+            <ContentShareControl />
+            <AudioOutputControl />
+          </>
+          : <DevicePermissionControl deviceLabels={DeviceLabels.AudioAndVideo} />}
+
+        <EndMeetingControl />
+      </ControlBar>
+    </StyledControls>
+  );
+};
+
+export default DynamicMeetingControls;
--- a/demo/meeting/src/containers/MeetingForm/index.tsx
+++ b/demo/meeting/src/containers/MeetingForm/index.tsx
@@ -5,6 +5,7 @@ import React, { useState, useContext, ChangeEvent } from 'react';
 import { useHistory } from 'react-router-dom';
 import {
   Input,
+  Checkbox,
   Flex,
   Heading,
   FormField,
@@ -23,6 +24,8 @@ import DevicePermissionPrompt from '../DevicePermissionPrompt';
 import RegionSelection from './RegionSelection';
 import { fetchMeeting, createGetAttendeeCallback } from '../../utils/api';
 import { useAppState } from '../../providers/AppStateProvider';
+import { MeetingMode } from '../../types';
+import { DeviceLabels } from 'amazon-chime-sdk-component-library-react/lib/types';
 
 const MeetingForm: React.FC = () => {
   const meetingManager = useMeetingManager();
@@ -37,8 +40,10 @@ const MeetingForm: React.FC = () => {
   const [nameErr, setNameErr] = useState(false);
   const [region, setRegion] = useState(appRegion);
   const [isLoading, setIsLoading] = useState(false);
+  const [isSpectatorModeSelected, setIsSpectatorModeSelected] = useState(false)
   const { errorMessage, updateErrorMessage } = useContext(getErrorContext());
   const history = useHistory();
+  const { setMeetingMode } = useAppState();
 
   const handleJoinMeeting = async (e: React.FormEvent) => {
     e.preventDefault();
@@ -66,11 +71,19 @@ const MeetingForm: React.FC = () => {
 
       await meetingManager.join({
         meetingInfo: JoinInfo.Meeting,
-        attendeeInfo: JoinInfo.Attendee
+        attendeeInfo: JoinInfo.Attendee,
+        deviceLabels: isSpectatorModeSelected === true ? DeviceLabels.None : DeviceLabels.AudioAndVideo,
       });
 
       setAppMeetingInfo(id, attendeeName, region);
-      history.push(routes.DEVICE);
+      if (isSpectatorModeSelected === true) {
+        setMeetingMode(MeetingMode.Spectator);
+        await meetingManager.start();
+        history.push(`${routes.MEETING}/${meetingId}`);
+      } else {
+        setMeetingMode(MeetingMode.Attendee);
+        history.push(routes.DEVICE);
+      }
     } catch (error) {
       updateErrorMessage(error.message);
     }
@@ -124,6 +137,15 @@ const MeetingForm: React.FC = () => {
         }}
       />
       <RegionSelection setRegion={setRegion} region={region} />
+      <FormField
+        field={Checkbox}
+        label="Join w/o Audio and Video"
+        value=""
+        checked={isSpectatorModeSelected}
+        onChange={() => (
+          setIsSpectatorModeSelected(!isSpectatorModeSelected)
+        )}
+      />
       <Flex
         container
         layout="fill-space-centered"
--- a/demo/meeting/src/types/index.ts
+++ b/demo/meeting/src/types/index.ts
@@ -34,3 +34,8 @@ export type ContentShareControlContextType = {
   toggleContentShare: () => Promise<void>;
   togglePauseContentShare: () => void;
 };
+
+export enum MeetingMode {
+  Spectator,
+  Attendee,
+};
--- a/demo/meeting/src/views/Meeting/index.tsx
+++ b/demo/meeting/src/views/Meeting/index.tsx
@@ -13,8 +13,10 @@ import { useNavigation } from '../../providers/NavigationProvider';
 import MeetingDetails from '../../containers/MeetingDetails';
 import MeetingControls from '../../containers/MeetingControls';
 import useMeetingEndRedirect from '../../hooks/useMeetingEndRedirect';
+import DynamicMeetingControls from '../../containers/DynamicMeetingControls';
+import { MeetingMode } from '../../types';
 
-const MeetingView = () => {
+const MeetingView = (props: { mode: MeetingMode, }) => {
   useMeetingEndRedirect();
   const { showNavbar, showRoster } = useNavigation();
 
@@ -26,7 +28,10 @@ const MeetingView = () => {
             className="videos"
             noRemoteVideoView={<MeetingDetails />}
           />
-          <MeetingControls />
+          {props.mode === MeetingMode.Spectator ?
+            <DynamicMeetingControls /> :
+            <MeetingControls />
+          }
         </StyledContent>
         <NavigationControl />
       </StyledLayout>
--- a/src/providers/DevicesProvider/AudioInputProvider.tsx
+++ b/src/providers/DevicesProvider/AudioInputProvider.tsx
@@ -104,11 +104,18 @@ const AudioInputProvider: React.FC = ({ children }) => {
       }
     }
 
+    const callback = (): void => {
+      initAudioInput();
+    };
+
+    meetingManager.subscribeToDeviceLabelTriggerChange(callback);
+
     initAudioInput();
 
     return () => {
       isMounted = false;
       audioVideo?.removeDeviceChangeObserver(observer);
+      meetingManager.unsubscribeFromDeviceLabelTriggerChange(callback);
     };
   }, [audioVideo]);
 
@@ -135,7 +142,7 @@ const useAudioInputs = (props?: DeviceConfig): DeviceTypeContext => {
   let { devices } = context;
   const { selectedDevice } = context;
   const { selectDeviceError } = context;
-  
+
   if (needAdditionalIO) {
     const additionalAudioInputs = getFormattedDropdownDeviceOptions(
       AUDIO_INPUT
--- a/src/providers/DevicesProvider/AudioOutputProvider.tsx
+++ b/src/providers/DevicesProvider/AudioOutputProvider.tsx
@@ -59,11 +59,18 @@ const AudioOutputProvider: React.FC = ({ children }) => {
       }
     }
 
+    const callback = (): void => {
+      initAudioOutput();
+    };
+
+    meetingManager.subscribeToDeviceLabelTriggerChange(callback);
+
     initAudioOutput();
 
     return () => {
       isMounted = false;
       audioVideo?.removeDeviceChangeObserver(observer);
+      meetingManager.unsubscribeFromDeviceLabelTriggerChange(callback);
     };
   }, [audioVideo]);
 
--- a/src/providers/DevicesProvider/VideoInputProvider.tsx
+++ b/src/providers/DevicesProvider/VideoInputProvider.tsx
@@ -76,11 +76,18 @@ const VideoInputProvider: React.FC = ({ children }) => {
       }
     }
 
+    const callback = (): void => {
+      initVideoInput();
+    };
+
+    meetingManager.subscribeToDeviceLabelTriggerChange(callback);
+
     initVideoInput();
 
     return () => {
       isMounted = false;
       audioVideo?.removeDeviceChangeObserver(observer);
+      meetingManager.unsubscribeFromDeviceLabelTriggerChange(callback);
     };
   }, [audioVideo]);
 
@@ -108,7 +115,7 @@ const useVideoInputs = (props?: DeviceConfig): DeviceTypeContext => {
   let { devices } = context;
   const { selectedDevice } = context;
   const { selectDeviceError } = context;
-  
+
   if (needAdditionalIO) {
     const additionalVideoInputs = getFormattedDropdownDeviceOptions(
       additionalIOJSON
--- a/src/providers/MeetingProvider/MeetingManager.ts
+++ b/src/providers/MeetingProvider/MeetingManager.ts
@@ -13,7 +13,8 @@ import {
   MeetingSessionStatusCode,
   AudioVideoObserver,
   MultiLogger,
-  MeetingSessionPOSTLogger
+  MeetingSessionPOSTLogger,
+  TimeoutScheduler
 } from 'amazon-chime-sdk-js';
 
 import {
@@ -21,7 +22,7 @@ import {
   supportsSetSinkId,
   videoInputSelectionToDevice
 } from '../../utils/device-utils';
-import { MeetingStatus } from '../../types';
+import { DeviceLabels, DeviceLabelTrigger, MeetingStatus } from '../../types';
 import {
   DevicePermissionStatus,
   MeetingJoinData,
@@ -30,6 +31,10 @@ import {
   PostLogConfig,
   ManagerConfig
 } from './types';
+
+function noOpDeviceLabelHook(): Promise<MediaStream> {
+  return Promise.resolve(new MediaStream());
+}
 
 export class MeetingManager implements AudioVideoObserver {
   meetingSession: DefaultMeetingSession | null = null;
@@ -72,6 +77,8 @@ export class MeetingManager implements AudioVideoObserver {
   selectVideoInputDeviceError: Error | null = null;
 
   selectVideoInputDeviceErrorObservers: ((selectVideoInputDeviceError: Error | null) => void)[] = [];
+
+  deviceLabelTriggerChangeObservers: (() => void)[] = [];
 
   audioInputDevices: MediaDeviceInfo[] | null = null;
 
@@ -134,7 +141,7 @@ export class MeetingManager implements AudioVideoObserver {
     this.audioVideoObservers = {};
   }
 
-  async join({ meetingInfo, attendeeInfo }: MeetingJoinData) {
+  async join({ meetingInfo, attendeeInfo, deviceLabels = DeviceLabels.AudioAndVideo, }: MeetingJoinData) {
     this.configuration = new MeetingSessionConfiguration(
       meetingInfo,
       attendeeInfo
@@ -147,7 +154,7 @@ export class MeetingManager implements AudioVideoObserver {
 
     this.meetingRegion = meetingInfo.MediaRegion;
     this.meetingId = this.configuration.meetingId;
-    await this.initializeMeetingSession(this.configuration);
+    await this.initializeMeetingSession(this.configuration, deviceLabels);
   }
 
   async start(): Promise<void> {
@@ -183,7 +190,8 @@ export class MeetingManager implements AudioVideoObserver {
   }
 
   async initializeMeetingSession(
-    configuration: MeetingSessionConfiguration
+    configuration: MeetingSessionConfiguration,
+    deviceLabels: DeviceLabels | DeviceLabelTrigger = DeviceLabels.AudioAndVideo,
   ): Promise<any> {
     const logger = this.createLogger(configuration);
     const deviceController = new DefaultDeviceController(logger);
@@ -196,7 +204,7 @@ export class MeetingManager implements AudioVideoObserver {
 
     this.audioVideo = this.meetingSession.audioVideo;
     this.setupAudioVideoObservers();
-    this.setupDeviceLabelTrigger();
+    this.setupDeviceLabelTrigger(deviceLabels);
     await this.listAndSelectDevices();
     this.publishAudioVideo();
     this.setupActiveSpeakerDetection();
@@ -269,31 +277,52 @@ export class MeetingManager implements AudioVideoObserver {
       (await this.audioVideo?.listAudioOutputDevices()) || [];
   }
 
-  setupDeviceLabelTrigger(): void {
-    const callback = async (): Promise<MediaStream> => {
-      this.devicePermissionStatus = DevicePermissionStatus.IN_PROGRESS;
-      this.publishDevicePermissionStatus();
-      try {
-        const devices = await navigator.mediaDevices.enumerateDevices();
-        const hasVideoInput = devices.some(value => {
-          return value.kind === 'videoinput';
-        });
+  setupDeviceLabelTrigger(deviceLabels: DeviceLabels | DeviceLabelTrigger = DeviceLabels.AudioAndVideo): void {
+    let callback: DeviceLabelTrigger;
 
-        const stream = await navigator.mediaDevices.getUserMedia({
-          audio: true,
-          video: hasVideoInput
-        });
+    if (typeof deviceLabels === 'function') {
+      callback = deviceLabels;
+    } else if (deviceLabels === DeviceLabels.None) {
+      callback = noOpDeviceLabelHook;
+    } else {
+      const constraints: MediaStreamConstraints = {};
 
-        this.devicePermissionStatus = DevicePermissionStatus.GRANTED;
-        this.publishDevicePermissionStatus();
-        return stream;
-      } catch (error) {
-        console.error('Failed to get device permissions');
-        this.devicePermissionStatus = DevicePermissionStatus.DENIED;
-        this.publishDevicePermissionStatus();
-        throw new Error(error);
+      switch (deviceLabels) {
+        case DeviceLabels.Audio:
+          constraints.audio = true;
+          break;
+        case DeviceLabels.Video:
+          constraints.video = true;
+          break;
+        case DeviceLabels.AudioAndVideo:
+          constraints.audio = true;
+          constraints.video = true;
+          break;
       }
-    };
+
+      callback = async (): Promise<MediaStream> => {
+        this.devicePermissionStatus = DevicePermissionStatus.IN_PROGRESS;
+        this.publishDevicePermissionStatus();
+        try {
+          const devices = await navigator.mediaDevices.enumerateDevices();
+          const hasVideoInput = devices.some(value => value.kind === 'videoinput');
+
+          const stream = await navigator.mediaDevices.getUserMedia({
+            audio: constraints.audio,
+            video: constraints.video && hasVideoInput,
+          });
+
+          this.devicePermissionStatus = DevicePermissionStatus.GRANTED;
+          this.publishDevicePermissionStatus();
+          return stream;
+        } catch (error) {
+          console.error('Failed to get device permissions');
+          this.devicePermissionStatus = DevicePermissionStatus.DENIED;
+          this.publishDevicePermissionStatus();
+          throw new Error(error);
+        }
+      };
+    }
 
     this.audioVideo?.setDeviceLabelTrigger(callback);
   }
@@ -314,7 +343,19 @@ export class MeetingManager implements AudioVideoObserver {
 
   async listAndSelectDevices(): Promise<void> {
     await this.updateDeviceLists();
+
+    let hasAudioInput = false, hasAudioOutput = false, hasVideoInput = false;
+    const devices = await navigator.mediaDevices.enumerateDevices();
+    for (const device of devices) {
+      if (device.label) {
+        if (device.kind === "audioinput") hasAudioInput = true;
+        if (device.kind === "audiooutput") hasAudioOutput = true;
+        if (device.kind === "videoinput") hasVideoInput = true;
+      }
+    }
+
     if (
+      hasAudioInput &&
       !this.selectedAudioInputDevice &&
       this.audioInputDevices &&
       this.audioInputDevices.length
@@ -330,6 +371,7 @@ export class MeetingManager implements AudioVideoObserver {
       this.publishSelectedAudioInputDevice();
     }
     if (
+      hasAudioOutput &&
       !this.selectedAudioOutputDevice &&
       this.audioOutputDevices &&
       this.audioOutputDevices.length
@@ -347,6 +389,7 @@ export class MeetingManager implements AudioVideoObserver {
       this.publishSelectedAudioOutputDevice();
     }
     if (
+      hasVideoInput &&
       !this.selectedVideoInputDevice &&
       this.videoInputDevices &&
       this.videoInputDevices.length
@@ -414,6 +457,11 @@ export class MeetingManager implements AudioVideoObserver {
     }
     this.publishSelectedVideoInputDevice();
     this.publishSelectVideoInputDeviceError();
+  };
+
+  invokeDeviceProvider = (deviceLabels: DeviceLabels) => {
+    this.setupDeviceLabelTrigger(deviceLabels);
+    this.publishDeviceLabelTriggerChange();
   };
 
   /**
@@ -613,6 +661,27 @@ export class MeetingManager implements AudioVideoObserver {
       callback(this.selectVideoInputDeviceError);
     }
   };
+
+  subscribeToDeviceLabelTriggerChange = (
+    callback: () => void
+  ): void => {
+    this.deviceLabelTriggerChangeObservers.push(callback);
+  };
+
+  unsubscribeFromDeviceLabelTriggerChange = (
+    callbackToRemove: () => void
+  ): void => {
+    this.deviceLabelTriggerChangeObservers = this.deviceLabelTriggerChangeObservers.filter(
+      callback => callback !== callbackToRemove
+    );
+  };
+
+  private publishDeviceLabelTriggerChange = (): void => {
+    for (const callback of this.deviceLabelTriggerChangeObservers) {
+      callback();
+    }
+  };
+
 }
 
 export default MeetingManager;
--- a/src/providers/MeetingProvider/types.ts
+++ b/src/providers/MeetingProvider/types.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { LogLevel } from 'amazon-chime-sdk-js';
+import { DeviceLabels, DeviceLabelTrigger } from '../../types';
 
 export enum DevicePermissionStatus {
   UNSET = 'UNSET',
@@ -13,6 +14,7 @@ export enum DevicePermissionStatus {
 export interface MeetingJoinData {
   meetingInfo: any;
   attendeeInfo: any;
+  deviceLabels?: DeviceLabels | DeviceLabelTrigger;
 }
 
 export interface AttendeeResponse {
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -61,4 +61,13 @@ export enum DevicePermissionStatus {
   IN_PROGRESS = 'IN_PROGRESS',
   GRANTED = 'GRANTED',
   DENIED = 'DENIED',
-}
+};
+
+export enum DeviceLabels {
+  None,
+  Audio,
+  Video,
+  AudioAndVideo,
+};
+
+export type DeviceLabelTrigger = () => Promise<MediaStream>;
--- a/demo/meeting/src/app.tsx
+++ b/demo/meeting/src/app.tsx
@@ -39,7 +39,7 @@ const App: FC = () => (
                   </Route>
                   <Route path={routes.MEETING}>
                     <NoMeetingRedirect>
-                      <Meeting />
+                      <MeetingModeSelector />
                     </NoMeetingRedirect>
                   </Route>
                 </Switch>
@@ -60,6 +60,14 @@ const Theme: React.FC = ({ children }) => {
       <GlobalStyles />
       {children}
     </ThemeProvider>
+  );
+};
+
+const MeetingModeSelector: React.FC = () => {
+  const { meetingMode } = useAppState();
+
+  return (
+    <Meeting mode={meetingMode} />
   );
 };
 
--- a/demo/meeting/src/containers/DevicePermissionControl/DevicePermissionControl.tsx
+++ b/demo/meeting/src/containers/DevicePermissionControl/DevicePermissionControl.tsx
@@ -1,0 +1,35 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { ControlBarButton, Cog, useMeetingManager, Camera, Sound, Dots, } from 'amazon-chime-sdk-component-library-react';
+import { DeviceLabels } from 'amazon-chime-sdk-component-library-react/lib/types';
+import React from 'react';
+import DevicePermissionPrompt from '../DevicePermissionPrompt';
+
+const DevicePermissionControl = (props: { deviceLabels: DeviceLabels }) => {
+  const meetingManager = useMeetingManager();
+
+  const handleClick = async () => {
+    await meetingManager.invokeDeviceProvider(props.deviceLabels);
+  }
+
+  const label = props.deviceLabels === DeviceLabels.AudioAndVideo ? "Device" :
+    props.deviceLabels === DeviceLabels.Audio ? "Audio" :
+      props.deviceLabels === DeviceLabels.Video ? "Video" :
+        "None"
+
+  const icon = props.deviceLabels === DeviceLabels.AudioAndVideo ? <Cog /> :
+    props.deviceLabels === DeviceLabels.Audio ? <Sound /> :
+      props.deviceLabels === DeviceLabels.Video ? <Camera /> :
+        <Dots />
+
+  return (
+    props.deviceLabels === DeviceLabels.None ? null :
+      <>
+        <ControlBarButton icon={icon} onClick={handleClick} label={label} />
+        <DevicePermissionPrompt />
+      </>
+  );
+}
+
+export default DevicePermissionControl;
--- a/demo/meeting/src/containers/DevicePermissionControl/Styled.tsx
+++ b/demo/meeting/src/containers/DevicePermissionControl/Styled.tsx
@@ -1,0 +1,8 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import styled from 'styled-components';
+
+export const StyledP = styled.p`
+  padding: 1rem 1rem 1rem 0;
+`;
--- a/demo/meeting/src/containers/DynamicMeetingControls/Styled.ts
+++ b/demo/meeting/src/containers/DynamicMeetingControls/Styled.ts
@@ -1,0 +1,22 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import styled from 'styled-components';
+
+interface StyledProps {
+  active: boolean;
+}
+
+export const StyledControls = styled.div<StyledProps>`
+  opacity: ${props => (props.active ? '1' : '0')};
+  transition: opacity 250ms ease;
+
+  @media screen and (max-width: 768px) {
+    opacity: 1;
+  }
+
+  .controls-menu {
+    width: 100%;
+    position: static;
+  }
+`;
--- a/demo/meeting/src/providers/AppStateProvider.tsx
+++ b/demo/meeting/src/providers/AppStateProvider.tsx
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React, { useContext, useState, ReactNode } from 'react';
+import { MeetingMode } from '../types';
 
 type Props = {
   children: ReactNode;
@@ -12,8 +13,10 @@ interface AppStateValue {
   localUserName: string;
   theme: string;
   region: string;
+  meetingMode: MeetingMode;
   toggleTheme: () => void;
   setAppMeetingInfo: (meetingId: string, name: string, region: string) => void;
+  setMeetingMode: (meetingMode: MeetingMode) => void;
 }
 
 const AppStateContext = React.createContext<AppStateValue | null>(null);
@@ -33,6 +36,7 @@ const query = new URLSearchParams(location.search);
 export function AppStateProvider({ children }: Props) {
   const [meetingId, setMeeting] = useState(query.get('meetingId') || '');
   const [region, setRegion] = useState(query.get('region') || '');
+  const [meetingMode, setMeetingMode] = useState(MeetingMode.Attendee);
   const [localUserName, setLocalName] = useState('');
   const [theme, setTheme] = useState(() => {
     const storedTheme = localStorage.getItem('theme');
@@ -64,8 +68,10 @@ export function AppStateProvider({ children }: Props) {
     localUserName,
     theme,
     region,
+    meetingMode,
     toggleTheme,
-    setAppMeetingInfo
+    setAppMeetingInfo,
+    setMeetingMode,
   };
 
   return (
--- a/src/index.ts
+++ b/src/index.ts
@@ -168,7 +168,12 @@ export { lightTheme, darkTheme, GlobalStyles, StyledReset } from './theme';
 export { VideoQuality } from './hooks/sdk/useSelectVideoQuality';
 
 // enums
-export { MeetingStatus, DevicePermissionStatus } from './types/index';
+export {
+  MeetingStatus,
+  DevicePermissionStatus,
+  DeviceLabels,
+  DeviceLabelTrigger,
+} from './types/index';
 export { Severity, ActionType } from './providers/NotificationProvider';
 
 // Class


ltrung:
Rebase into one commit please

CHANGELOG.md
```diff
@@ -19,12 +20,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added optional props to specify icon titles in the `AudioInputControl` and `ContentShareControl` components.
 - Added optional props to specify the dropdown text that shows when no video quality is selected, in the `QualitySelection` component.
 - Added observer to `selectAudioInputDeviceError` and `selectVideoInputDeviceError` to deliver the error from SDK level to client application level.
-  
+- Added optional parameter `isDevicePreviewSkipped` in `meetingManager.join()` to suppress device permission request.
```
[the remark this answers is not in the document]
xuesichao: Hi Richard, thank you so much for giving me these comments. I really had a hard time naming them. Your suggestion is right what I need.
The main goal of this PR is to provide our customers with a simple way to achieve the view-only mode, that is, users join a meeting without audio input device and video input device, and no device permission request appears during the joining process.
At the beginning this optional parameter is called `joinAsSpectator`, and we find it too specific, then we change the name to 'skipDevicePreview` to make it more generic.
Do you have any better idea for its naming?

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -101,6 +101,18 @@ export class MeetingManager implements AudioVideoObserver {
...
+
+  suppressAudioOutputDevicePermissionRequest: (() => void) | null = null;
+
+  suppressVideoInputDevicePermissionRequest: (() => void) | null = null;
+
+  invokeAudioInputDevicePermissionRequest: (() => void) | null = null;
+
+  invokeAudioOutputDevicePermissionRequest: (() => void) | null = null;
+
+  invokeVideoInputDevicePermissionRequest: (() => void) | null = null;
```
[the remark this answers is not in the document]
xuesichao: This is a very good suggestion, I will change them to an enum. Could you please give me an example here, how should I phrase these function? From my shallow understanding, I find every word is necessary 😅.

The permission only works when permission is unset. If the permission has been already granted before user joins a meeting, they will joins with null devices, and they can change the selected device through the `MeetingControls`.

These function is to help our customer to build the view-only mode easily. But it should be much great if we can have a complete set of mechanisms to manage device permission.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -134,7 +146,7 @@ export class MeetingManager implements AudioVideoObserver {
     this.audioVideoObservers = {};
   }
 
-  async join({ meetingInfo, attendeeInfo }: MeetingJoinData) {
+  async join({ meetingInfo, attendeeInfo, isDevicePreviewSkipped = false }: MeetingJoinData) {
```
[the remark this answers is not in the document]
xuesichao: This parameter is to help user switch between normal mode and view-only mode when joining a meeting. But we don't want to make the case too specific, so we change the name as `skipDevicePreview`. It's more like delay the device permission request than suppress it. So we also provide a `invokeDevicePermissionRequests()` to let user grant the permission whenever they want.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -183,7 +195,8 @@ export class MeetingManager implements AudioVideoObserver {
   }
 
   async initializeMeetingSession(
-    configuration: MeetingSessionConfiguration
+    configuration: MeetingSessionConfiguration,
+    isDevicePermissionRequestAllowed: boolean,
```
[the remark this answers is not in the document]
xuesichao: Yes, we actually control the device permission here.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -197,7 +210,13 @@ export class MeetingManager implements AudioVideoObserver {
...
     this.setupAudioVideoObservers();
     this.setupDeviceLabelTrigger();
-    await this.listAndSelectDevices();
+
+    if (isDevicePermissionRequestAllowed === true) {
+      await this.listAndSelectDevices();
+    } else {
+      this.suppressDevicePermissionRequests();
+    }
+
```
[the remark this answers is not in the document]
xuesichao: The `setupDeviceLabelTrigger` create a callback function to change the `devicePermissionStatus` while user granting permission, and assign this callback to `DefaultDeviceController.deviceLabelTrigger` in JS SDK. It won't call `getUserMedia`, it just overwrites the default `deviceLabelTrigger` function.

Later when user calls `invokeDevicePermissionRequests()`, the `devicePermissionStatus` will be changed as well, which remain consistency with the actual status.

I think it won't be an issue here. I tested it in the meeting demo for many times.

xuesichao:
> This needs a big pile of developer docs explaining how a builder (a) joins a meeting without requesting audio or video permissions, (b) joins a meeting requesting only camera (silent), (c) joins a meeting requesting only microphone (no video), and (d) how they need to control when permissions are requested in order to perform certain actions, like listing devices or choosing a non-synthetic audio device.
> 
> You also need to explain how they can supply a synthetic non-silent audio stream in the case of joining muted.

I will write a how-to doc later to explain these contents to our users in detail. Thank you so much for pointing out the key points 👍.

CHANGELOG.md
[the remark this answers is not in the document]
xuesichao: The `enum DeviceLables` is really a great idea. We could use it to have more flexible controls over the device permission. I will make the improvement!

For second issue, I add a checkbox in `MeetingForm` in meeting demo. So user could select between normal mode and view-only mode when they join a meeting. 

image

I made the `MeetingControls` to be rendered conditionally according to the `devicePermissionStatus` exposed by the `useDevicePermissionStatus`.

For normal mode, everything works as before.

For view-only mode, user skips the device setting page, and join the meeting with `null` devices, the speaker works as usual. The `MeetingControls` displays only `EndMeetingControl` and `DevicePermissionControl`. 

image


'DevicePermissionControl` can trigger `audioVideo.listXxxxxXxxxxDevices()` function, user could grant device permission through this component. 

image

After the permission is granted, the `MeetingControls` is rendered like normal way. User could select and set the device through Device Controls in `MeetingControls`. And at this time everything works like the normal mode.

image

I think it would be more practical to do it like this, device permission is delayed instead of suppressed.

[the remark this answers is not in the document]
richnew10: It's not really "Skip device preview", it's "join without audio or video", no?

Be careful about the behavior when the user chooses Deny, or doesn't save permissions — test in Firefox rather than Chrome.

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
richnew10: I would try to avoid defining these functions if at all possible; they don't mean what they seem to mean. For example, if I call

```
mm.setSuppressAudioInputDevicePermissionRequest(() => {});
mm.chooseAudioInputDevice('default');
```

I might expect there to be no permissions dialog… but there will be one as a side effect of the `getUserMedia` call.

Permissions dialogs are a side-effect of `getUserMedia`, and having been granted permissions is a precondition of `enumerateDevices`. Those are the ground truths of this system.

As such, there is only one thing you need to be able to do: prior to the builder listing devices for the first time, they need to be able to invoke the device label trigger that the device controller ordinarily invokes for you. 

As a builder you might know that you don't need to (_e.g._, because you're running in Electron, or because you have already chosen a device), or you might want control of exactly when.

Just as we will add a `deviceLabels` option to `join`, we can add a single method `invokeDeviceLabelTrigger` that takes `DeviceLabels` as input.

Does that make sense?

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
richnew10: This is in the component library, not the demo — my app might not have a device preview at all! This is the point where you would take `deviceLabels` as an argument.

CHANGELOG.md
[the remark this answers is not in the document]
xuesichao: You're right. I renamed the checkbox label.

For the behavior here, there are two issues right now. 
* The first one is https://github.com/aws/amazon-chime-sdk-js/issues/474. I gonna work on it after this PR being merged.

* The second is that we use the state of `useDevicePermissionStatus` hook to render the `MeetingControls` conditionally. But the it cannot be synchronized with the actual device permission status, if the browser saves the device permission. e.g. User gave "Always Allow" device permission to browser when using the app last time. This time user opens the Application, the actual permission status is "Granted", but the state returned by the hook is "Unset". Since we gonna add controls to device permission, it should be much better to solve this issue. But currently I can't find some good way to detect the device permission at the very beginning. 

Any suggestion would be helpful!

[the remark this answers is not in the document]
richnew10: This is why I suggest not trying. Build two experiences, handle failures, and to transition between the two, invoke the device label function. Do not try to represent permission status in React state; that hook probably should not exist.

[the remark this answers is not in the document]
richnew10: To be clear: you _cannot_ find out whether the user has granted permissions without enumerating devices and checking to see whether they have labels.

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
richnew10: In either case you'd be setting the device label trigger to a new function — either one that depends on the enum value, or to the exact provided function.

Something a bit like this:

```typescript
  async join({ meetingInfo, attendeeInfo, deviceLabels = DeviceLabels.AudioAndVideo }: MeetingJoinData) {
    …
    await this.initializeMeetingSession(this.configuration, deviceLabels);
  }

  async initializeMeetingSession(
    configuration: MeetingSessionConfiguration,
    deviceLabels: DeviceLabels | DeviceLabelTrigger,
  ): Promise<any> {
    const logger = this.createLogger(configuration);
    const deviceController = new DefaultDeviceController(logger);

    this.meetingSession = new DefaultMeetingSession(
      configuration,
      logger,
      deviceController
    );

    this.audioVideo = this.meetingSession.audioVideo;
    this.setupAudioVideoObservers();

    // Amend `sDLT` accordingly: it now takes an optional trigger argument!
    this.setupDeviceLabelTrigger(deviceLabels);

    // … and now you can do this. It's a bit imprecise, because
    // the builder might want to skip this even if they provide a function,
    // but that's OK.
    if (deviceLabels !== DeviceLabels.None) {
      // TODO: selecting and listing devices should NOT be combined!
      await this.listAndSelectDevices();
    }

    this.publishAudioVideo();
    this.setupActiveSpeakerDetection();
    this.meetingStatus = MeetingStatus.Loading;
    this.publishMeetingStatus();
  }
```

CHANGELOG.md
```diff
@@ -19,12 +20,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added optional props to specify icon titles in the `AudioInputControl` and `ContentShareControl` components.
 - Added optional props to specify the dropdown text that shows when no video quality is selected, in the `QualitySelection` component.
 - Added observer to `selectAudioInputDeviceError` and `selectVideoInputDeviceError` to deliver the error from SDK level to client application level.
+- Added optional parameter `deviceLabels` in `meetingManager.join()`. The device that the browser will request permission can be set through this parameter. The customized `deviceLabelTrigger` function could also be set through it.
```
ltrung, line 23: Can we just have one succinct summary here about the changes?

src/hooks/sdk/docs/useDevicePermissionStatus.stories.mdx
```diff
@@ -44,9 +44,9 @@ const App = () => (
 );
 
 const MyChild = () => {
-  const meetingStatus = useMeetingStatus();
+  const devicePermissionStatus = useDevicePermissionStatus();
```
ltrung, line 47: This is fixed already. You should rebase :)

src/types/index.ts
```diff
@@ -61,4 +61,19 @@ export enum DevicePermissionStatus {
...
+};
+
+export enum DeviceLabels {
+  None,
+  Audio,
+  Video,
+  AudioAndVideo,
+};
+
+export enum Device {
```
ltrung, line 73: nit: Should this be deviceType?

xuesichao: `DeivceType` is a better naming for sure. The question is, here it's an `enum`, and I found in this repo, anything with a 'Type' suffix, is a `type` type. Besides, we already have a `DeviceType`, see below.
```ts
export type DeviceType = {
  deviceId: string;
  label: string;
};
```
Considering these, do we still need to rename it?

ltrung: How about DeviceKind? That aligns with webRTC MediaStream.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -312,47 +377,58 @@ export class MeetingManager implements AudioVideoObserver {
...
+          !this.selectedVideoInputDevice &&
+          this.videoInputDevices &&
+          this.videoInputDevices.length
+        ) {
+          this.selectedVideoInputDevice = this.videoInputDevices[0].deviceId;
+          this.publishSelectedVideoInputDevice();
+        }
+        break;
+      default:
+        break;
```
ltrung, line 431: nit: You do not need break for default. Also default should never happen right. If so, you should log or throw an error here.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -269,33 +311,56 @@ export class MeetingManager implements AudioVideoObserver {
...
-        const devices = await navigator.mediaDevices.enumerateDevices();
-        const hasVideoInput = devices.some(value => {
-          return value.kind === 'videoinput';
-        });
-
-        const stream = await navigator.mediaDevices.getUserMedia({
-          audio: true,
-          video: hasVideoInput
-        });
+  async updateDeviceList(device: Device): Promise<void> {
```
ltrung, line 314: Isn't this a breaking changes? Before builders can call `meetingManager.updateDeviceList()` but not it would fail.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -269,33 +311,56 @@ export class MeetingManager implements AudioVideoObserver {
...
+    }
+  }
 
-        this.devicePermissionStatus = DevicePermissionStatus.GRANTED;
-        this.publishDevicePermissionStatus();
-        return stream;
-      } catch (error) {
-        console.error('Failed to get device permissions');
-        this.devicePermissionStatus = DevicePermissionStatus.DENIED;
+  setupDeviceLabelTrigger(deviceLabels: DeviceLabels | DeviceLabelTrigger): void {
```
ltrung, line 333: Same here. These should all be optional.

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: No, it's not a breaking change. This is a new function, the original one is called `updateDeviceLists()`, it's unchanged. 
I was wondering should I merge these two functions? e.g. make the updateDeviceLists() as default behavior in switch case, or something else.

src/types/index.ts
xuesichao: Sounds great!

src/providers/DevicesProvider/AudioOutputProvider.tsx
```diff
@@ -59,7 +67,11 @@ const AudioOutputProvider: React.FC = ({ children }) => {
       }
     }
 
-    initAudioOutput();
+    meetingManager.setInvokeDeviceProviderInit(Device.AudioOutput, initAudioOutput);
+
+    if (isInitAllowed === true) {
```
ltrung, line 72: Don't need to compare to `true` here.

src/providers/MeetingProvider/types.ts
```diff
@@ -13,6 +14,7 @@ export enum DevicePermissionStatus {
 export interface MeetingJoinData {
   meetingInfo: any;
   attendeeInfo: any;
+  deviceLabels?: DeviceLabels | DeviceLabelTrigger;
```
ltrung, line 17: Could we not add here? deveiceLabels seems not related to meeting join data at all? You should put it as another optional parameter.
Also should we pass it through join? I think a better place would be meetingConfig when you first create meetingManager.

src/providers/MeetingProvider/MeetingManager.ts
ltrung: This is confusing in API levels to have 2 API so similar name. You should rewrite the old one and add device as an optional parameter

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -101,6 +101,18 @@ export class MeetingManager implements AudioVideoObserver {
 
   simulcastEnabled: boolean = false;
 
+  skipAudioInputProviderInit: (() => void) | null = null;
```
ltrung, line 104: Do we need to do it this way? 
Can we just have one parameter such as setupDeviceConfig of type `DeviceLabels` which builders can pass via meetingConfig.  We can have a function in `meetingManager` such as `initDevices(deviceLabels: deviceLabels | deviceLabelTrigger)`
`AudioInputProvider`, `VideoInputProvider` can watch `setupDeviceConfig` flag. if it is None then do not do anything. If it changes to the corresponding value, then init.

src/providers/MeetingProvider/types.ts
xuesichao: For the first question, I agree with you. I will move the `deviceLabels` out of the `MeetingJoinData` and just make it as an optional parameter.

For the second question, I think the key point here is the timing of setting device permission request.

We could pass the `DeviceLabel` via `meetingCongfig` at the very beginning. But the `DeviceLabel` could be changed , right?
 
Let's say, we set the `AudioAndVideo` as default value in `meetingConfig`. So the default setting is for the **Attendee Mode**. When user wants to switch to the **Spectator mode**, they need do something to change the device permission setting before joining the meeting. So here is the timing. It's best to setup the device permission right before they joining the meeting. That's why I think it's a good idea to pass the parameter in the `meetingManager.join()`.

How do you think?

xuesichao: If you agree that the timing before joining is a key point, and we must do something at this moment. Then the pass the `DeviceLabel` via `meetingConfig` seems redundant. Cause we need to setup the device permission before joining the meeting anyway.

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: Throw error could be a breaking change here. I would prefer log something here. Could I just use `console.log`?

ltrung: Sure you can use console.error

src/providers/MeetingProvider/types.ts
ltrung: Sure we can just have one function to setDeviceLabelMode or something.

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: If we don't need to pass `DeviceLabel` via `MeetingConfig`, we can't use the subscribe pattern. Cause we have nothing to subscribe to. If we add something to subscribe, it could be quite similar with the these functions. Any better idea?

src/types/index.ts
richnew10: Just to clarify: a `Type` suffix does not require that it be a TypeScript `type` alias. `Kind` is often a good word, but wanted to avoid an accidental rule forming.

src/providers/MeetingProvider/MeetingManager.ts
richnew10: Or make the new one supersede the old one. Getting in the habit of returning `Promise` is a good thing.

xuesichao: The `DeviceKind` enum only has `AudioInput`, `AudioOutput` and `VideoInput` now. So the `updateDeviceList()` is designed to support device specific operation. If I need to merge the `updateDeviceLists()`(the original one) in to this one, it should be better to have some other option in this enum.

How about add `AuidoAndVideo` and `None` in `DeviceKind` enum? And if we do so, the `DeviceKind` enum will almost be a superset of `DeviceLabels`. Should we keep both of them?

```ts
export enum DeviceLabels {
  None,
  Audio,
  Video,
  AudioAndVideo,
};

export enum DeviceKind {
  AudioInput,
  AudioOutput,
  VideoInput,
}
```

src/providers/MeetingProvider/types.ts
richnew10: I want to reiterate: the _browser_ prompts for permissions, and not all applications (a) need to enumerate devices, (b) need device labels, or (c) need permissions at all (e.g., Electron apps).

Nor do all applications have a linear procession through creating a meeting manager, joining a meeting, prompting for labels, etc. — the same window might join multiple calls, and permissions or labels are already a solved problem for subsequent sessions.

xuesichao: @richnew10 Yes, so I think it's good to have the optional parameter in `meetingManager.join()` function to let use choose the preset settings for `None`, `Audio`, `Video` and `AudioAndVideo`, and expose the functions like below to control specific type of device, how do you think about this?
```ts
  skipDeviceProviderInit(deviceKind: DeviceKind): void {
    switch (deviceKind) {
      case DeviceKind.AudioInput:
        if (this.skipAudioInputProviderInit) {
          this.skipAudioInputProviderInit();
        }
        break;
      case DeviceKind.AudioOutput:
        if (this.skipAudioOutputProviderInit) {
          this.skipAudioOutputProviderInit();
        }
        break;
      case DeviceKind.VideoInput:
        if (this.skipVideoInputProviderInit) {
          this.skipVideoInputProviderInit();
        }
        break;
      default:
        break;
    }
  };

  invokeDeviceProviderInit(deviceKind: DeviceKind): void {
    switch (deviceKind) {
      case DeviceKind.AudioInput:
        if (this.invokeAudioInputProviderInit) {
          this.invokeAudioInputProviderInit();
        }
        break;
      case DeviceKind.AudioOutput:
        if (this.invokeAudioOutputProviderInit) {
          this.invokeAudioOutputProviderInit();
        }
        break;
      case DeviceKind.VideoInput:
        if (this.invokeVideoInputProviderInit) {
          this.invokeVideoInputProviderInit();
        }
        break;
      default:
        break;
    }
  };

  invokeDeviceLabelTrigger(deviceLabels: DeviceLabels): void {
    this.setupDeviceLabelTrigger(deviceLabels);

    switch (deviceLabels) {
      case DeviceLabels.AudioAndVideo:
        this.invokeDeviceProviderInit(DeviceKind.AudioInput);
        this.invokeDeviceProviderInit(DeviceKind.AudioOutput);
        this.invokeDeviceProviderInit(DeviceKind.VideoInput);
        break;
      case DeviceLabels.Audio:
        this.invokeDeviceProviderInit(DeviceKind.AudioInput);
        this.invokeDeviceProviderInit(DeviceKind.AudioOutput);
        break;
      case DeviceLabels.Video:
        this.invokeDeviceProviderInit(DeviceKind.VideoInput);
        break;
      default:
        break;
    }
  };
```

richnew10: I don't really understand the `skip*ProviderInit` functions, or why the name `skip` should be used. All you're really trying to do, right, is to control a particular hook that's run during initialization of this provider. It seems like the way to do that is a constructor or init argument. What am I missing?

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -101,6 +101,18 @@ export class MeetingManager implements AudioVideoObserver {
...
+  skipAudioOutputProviderInit: (() => void) | null = null;
+
+  skipVideoInputProviderInit: (() => void) | null = null;
+
+  invokeAudioInputProviderInit: (() => void) | null = null;
+
+  invokeAudioOutputProviderInit: (() => void) | null = null;
+
+  invokeVideoInputProviderInit: (() => void) | null = null;
+
```
xuesichao, line 115: @richnew10 These functions is designed to receive the callbacks from those `DeviceProvider`s to control the device permission request. The logic is quite straightforward now, but it looks bad. Do you think I need to avoid them or improve them to subscribe pattern? Or do you have any better idea?

richnew10: I don't see why those providers need to know anything about the device label hook. They don't use labels, right?

richnew10: I'm sure I've written this before, but: there are only two things that trigger a permissions request — choosing a device and the device label hook. The only thing we need to care about here is the device label hook, and that's controlled by the MeetingManager. The only thing you need to do in this PR is provide a way to configure the MeetingManager to control the device label hook.

xuesichao: @richnew10 For example, there is a `initAudioInput()` function in `AudioInputProvider`, and it's called after rendering the `AudioInputProvider`. `initAudioInput` function calls `audioVideo.listAudioInputDevices()`, which will call `deviceLabelTrigger` in JS SDK. 

So I have to block/skip the `initAudioInput()` calling, and invoke it when user wants to grant permission to browser.

There could be another way, we just mock an empty `deviceLabelTrigger` funcion via `setupDeviceLabelTrigger`. So even the `initAudioInput()` is called, it won't trigger any permission request. I'm not sure how to mock an empty `deviceLabelTrigger`, cause the `getUserMedia` requires either `audio` or `video` to be `true`. I would appreciate it if you can teach me how to do this.

richnew10: Yes, that's exactly the way I suggest you do it — 

```
defaultDeviceController.setDeviceLabelTrigger(() => Promise.resolve(new MediaStream()));
```

should be sufficient if the user doesn't want a permissions dialog to be triggered. That will cause device labels to be missing if the user hasn't granted permissions.

xuesichao: @richnew10 Cool, now I know what to do! Thank you so much!

xuesichao: @richnew10 I tried to improve the code as you suggested. It works well on suppressing device permission request. But there is new problem. Although we set an empty `deviceLabelTrigger` to JS SDK, but we still call `audioVideo.listDevicesOfKind` for each kind of device in `DeviceProvider`. The devices info with empty label will be cached in `deviceInfoCache`. And when we want to invoke the `DeviceProvider` to trigger device permission request, we have to run `audioVideo.listDevicesOfKind` for the desired deviceKind again. At this time the device info with empty label in the `deviceInfoCache` will be directly returned. The `deviceLabelTrigger` won't be called, so there is no device permission request and the devices info won't be updated. How could we solve this problem? 

Potential Solution:
* somehow reset `deviceInfoCache`
* somehow force `deviceLabelTrigger` being called

Since this is a change to component SDK, I think we should avoid the change to JS SDK.

```ts
private async listDevicesOfKind(deviceKind: string): Promise<MediaDeviceInfo[]> {
    if (this.deviceInfoCache === null || !this.isWatchingForDeviceChanges()) {
    // This function is skipped in the second calling
      await this.updateDeviceInfoCacheFromBrowser();
    }
    return this.listCachedDevicesOfKind(deviceKind);
  }

  private async updateDeviceInfoCacheFromBrowser(): Promise<void> {
    const doesNotHaveAccessToMediaDevices = typeof MediaDeviceInfo === 'undefined';
    if (doesNotHaveAccessToMediaDevices) {
      this.deviceInfoCache = [];
      return;
    }

    let devices = await navigator.mediaDevices.enumerateDevices();

    let hasDeviceLabels = true;
    for (const device of devices) {
      if (!device.label) {
        hasDeviceLabels = false;
        break;
      }
    }
    if (!hasDeviceLabels) {
      try {
        this.logger.info('attempting to trigger media device labels since they are hidden');
        const triggerStream = await this.deviceLabelTrigger();
        devices = await navigator.mediaDevices.enumerateDevices();
        for (const track of triggerStream.getTracks()) {
          track.stop();
        }
      } catch (err) {
        this.logger.info('unable to get media device labels');
      }
    }
    this.deviceInfoCache = devices;
  }
```

richnew10: If the builder never calls `chooseAudioOutputDevice` or selects an input device or adds a device change observer then `onDeviceChangeCallback` will not be set, and so `isWatchingForDeviceChanges` will be false, and so the cache should be re-updated, no? So the issue is that they've selected a media stream but we've still decided to subscribe to device change notifications.

Regardless, I think this is worth fixing in the SDK, because it affects customers who don't use the component library:

```typescript
ddc.setDeviceLabelTrigger(() => Promise.resolve(new MediaStream()));
await ddc.listAudioInputDevices();
ddc.setDeviceLabelTrigger(() => navigator.mediaDevices.getUserMedia({audio: true}));
await ddc.listAudioInputDevices();  // < returns stale data, and no way to change it
```

If there are no device labels and no cache, we perhaps shouldn't set `deviceInfoCache` at the end of `updateDeviceInfoCacheFromBrowser`.

If the builder changes the trigger, we should perhaps throw away the cache.

And it also makes sense to allow the builder to discard the cache manually without recreating the device controller, or to force an update by adding a `force` parameter to the list functions.

ltrung: @xuesichao Agreed with Richard here. Let add an API to refresh device cache in JS SDK.

xuesichao: Ok I get it! Let me do this. Thanks guys!

xuesichao: @richnew10 I updated the PR with my current fix. The `invokeDeviceProvider(deviceLabels)` works well right now. The updated PR should pass the GitHub check successfully after next JS SDK release.

The last part left is I need to handle the logic when users pass a `deviceLabels` of `DeviceLabelTrigger` type to `meetingManager.join(deviceLabels)`.  I think the main difference should be the constraints in `getUserMedia()`.

```js
{ audio: true, video: true }
```

Since we already cover all the cases in`DeviceLabels` type, I'm curious how should we handle this case?

I mean if user pass a customized `deviceLabelTrigger` like `meetingManager.join(deviceLabelsTrigger)`, how should we control the logic of `DeviceProvider`. The invoke logic can be controlled by the users themselves. But how could we know which `DeviceProvider` should be delayed?

xuesichao:
@ltrung @richnew10 Updated the code. Test it locally and it works well. The PR should pass all the checks after today's JS SDK release.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -314,7 +329,19 @@ export class MeetingManager implements AudioVideoObserver {
 
   async listAndSelectDevices(): Promise<void> {
     await this.updateDeviceLists();
```
xuesichao, line 331: Keep the original design. Since we still need to call all `init*Devices()` function in `DeviceProvider`s when user calls `meetingManager.join()`. It's okay that we update all the device lists in any cases.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -314,7 +329,19 @@ export class MeetingManager implements AudioVideoObserver {
...
+
+    let hasAudioInput = false, hasAudioOutput = false, hasVideoInput = false;
+    const devices = await navigator.mediaDevices.enumerateDevices();
+    for (const device of devices) {
+      if (device.label) {
+        if (device.kind === "audioinput") hasAudioInput = true;
+        if (device.kind === "audiooutput") hasAudioOutput = true;
+        if (device.kind === "videoinput") hasVideoInput = true;
+      }
+    }
```
xuesichao, line 341: This may be a breakthrough change. The main logic here is that we only call `choose*Device` for those devices with labels.

For example, set `deviceLabels` to `DeviceLabels.Video`. On the device setting page, the first device is selected for video input, and empty labels are displayed for audio input and audio output. Later on the meeting page, the first device is selected for video input, and "None" is selected for audio input and audio output.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -416,6 +445,26 @@ export class MeetingManager implements AudioVideoObserver {
...
+        }
+
+        if (this.invokeAudioOutputProvider) {
+          this.invokeAudioOutputProvider();
+        }
+
+      case deviceLabels === DeviceLabels.AudioAndVideo || deviceLabels === DeviceLabels.Video:
+        if (this.invokeVideoInputProvider) {
+          this.invokeVideoInputProvider();
+        }
```
xuesichao, line 464: Is it good to write the code like this or better change to `if ... else ...` one?

src/providers/DevicesProvider/VideoInputProvider.tsx
```diff
@@ -68,14 +68,15 @@ const VideoInputProvider: React.FC = ({ children }) => {
         return;
       }
 
-      const devices = await audioVideo.listVideoInputDevices();
+      const devices = await audioVideo.listVideoInputDevices(true);
```
richnew10, line 71: Why do you need to force?

src/providers/DevicesProvider/VideoInputProvider.tsx
```diff
@@ -68,14 +68,15 @@ const VideoInputProvider: React.FC = ({ children }) => {
...
 
       if (isMounted) {
         setVideoInputs(devices);
         audioVideo.addDeviceChangeObserver(observer);
       }
     }
 
+    meetingManager.invokeVideoInputProvider = initVideoInput;
```
richnew10, line 79: This seems like it will leak badly.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -101,6 +101,12 @@ export class MeetingManager implements AudioVideoObserver {
 
   simulcastEnabled: boolean = false;
 
+  invokeAudioInputProvider: (() => Promise<void>) | null = null;
+
+  invokeAudioOutputProvider: (() => Promise<void>) | null = null;
+
+  invokeVideoInputProvider: (() => Promise<void>) | null = null;
+
```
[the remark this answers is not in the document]
richnew10: I'm not sure I follow.

Before use, the builder sets their device label preference.

The existing logic ensures that the correct hook is invoked.

When they change their device label preference, update the hook in `audioVideo`. That will clear the cache. 

Then you need one thing: a way to make sure that `list*Inputs` is invoked again.

The right way to do that, I think, is to make the `useEffect` blocks in `{Video,Audio}InputProvider` depend on a `deviceLabels` prop, so when it changes they update their state.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -269,31 +276,39 @@ export class MeetingManager implements AudioVideoObserver {
...
-          video: hasVideoInput
-        });
-
-        this.devicePermissionStatus = DevicePermissionStatus.GRANTED;
-        this.publishDevicePermissionStatus();
-        return stream;
-      } catch (error) {
-        console.error('Failed to get device permissions');
-        this.devicePermissionStatus = DevicePermissionStatus.DENIED;
+    if (typeof deviceLabels === "function") {
```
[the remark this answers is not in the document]
richnew10: TypeScript should make the following work:

```typescript
switch deviceLabels {
  case DeviceLabels.None: {
    …
    break;
  }
…

  case DeviceLabels.Audio: {
  }

…

  default: {
    // I must be a function.
  }
}
```

See [this Playground](https://www.typescriptlang.org/play?#code/KYOwrgtgBAYg9nKBvAUFKAjAhgJwDRqZYBeKAvigGZggDGALgJZwhQAeAFAG4BcsCUAD5QOHAJRQAvAD4oXOIwAmYsX3lLkhAM4B3RvVoALEVwmp06Wli3B+cAHTYcfDDmBYA1gG5Cl67fgHbGIXN08fCyhFYEosMAAbeh5fCy5xCKgKMiA).

src/providers/MeetingProvider/MeetingManager.ts
richnew10: It is OK to choose devices that don't have labels…

src/providers/MeetingProvider/MeetingManager.ts
richnew10: It's a bit opaque.

richnew10 (changes requested):

src/providers/DevicesProvider/VideoInputProvider.tsx
xuesichao: Currently, this function is called twice in general:
1. When user joins a meeting without devices. At this time the device info cache with empty label establishes.
2. When user wants to invoke the devices. At this time we need to force update the previous device info cache.

To cover the case 2, I made it force update.
But I need to improve the logic of invoking devices, so the logic may be changed here. I will take a look at it.

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: This is to make the code concise. The basic logic here is, if `DeviceLabels` contains "Audio", we invoke audio input/output devices, if `DeviceLabels` contains "Video", we invoke video devices.

But we gonna change the invoke logic to `useEffect` one, so this could be unnecessary.

If we split the cases, the code is quite redundant.
```ts
  invokeDeviceProvider(deviceLabels: DeviceLabels): void {
    this.setupDeviceLabelTrigger(deviceLabels);

    switch (deviceLabels) {
      case DeviceLabels.Audio:
        if (this.invokeAudioInputProvider) {
          this.invokeAudioInputProvider();
        }
        if (this.invokeAudioOutputProvider) {
          this.invokeAudioOutputProvider();
        }
        break;
      case DeviceLabels.Video:
        if (this.invokeVideoInputProvider) {
          this.invokeVideoInputProvider();
        }
        break;
      case DeviceLabels.AudioAndVideo:
        if (this.invokeAudioInputProvider) {
          this.invokeAudioInputProvider();
        }

        if (this.invokeAudioOutputProvider) {
          this.invokeAudioOutputProvider();
        }

        if (this.invokeVideoInputProvider) {
          this.invokeVideoInputProvider();
        }
        break;
    }
  }
```

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
xuesichao: I'm not sure I follow, it seems the `switch case` is not good for our cases. What we need to do is:
- `DeviceLabels` type
  1. Config the `constraints` of `getUserMedia()` based on `DeviceLabels`
  2. Set the callback with constraints
- `DeviceLabelTrigger` type
  1. Set the user passed callback

So the code is more like:
``` ts
let callback: () => Promise<MediaStream>;

switch (deviceLabels) {
  case DeviceLabels.None: {
    callback = () => Promise.resolve(new MediaStream());
    break;
  }
  case DeviceLabels.Audio: {
  // config constraints of `getUserMedia`
  }
  case DeviceLabels.Video: {
  // config constraints of `getUserMedia`
  }
  case DeviceLabels.AudioAndVideo: {
  // config constraints of `getUserMedia`
  }

  // If the `deviceLables` is `DeviceLabels` type, we need to assign 
  // the device label trigger with `constraints` to `callback`, and break here

  default: {
    // I must be a function.
    callback = deviceLabels;
  }

  this.audioVideo?.setDeviceLabelTrigger(callback);
}
```
Or it could be quite redundant, if we set the callback in each case and break.

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
xuesichao: Make sense, this is a better way. Let me try it! Thanks!

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
richnew10: Redundancy is OK! Optimize for clarity and code changing over time.

Imagine if another engineer needs to add a case to `DeviceLabels` in the future. Using a comprehensive `switch` statement will guarantee you a compiler error if you fail to handle the new case, which is great. `if` doesn't get you that all the time. You can do a `"function"` type check then a `switch` afterwards, if you prefer. Use your judgment — as requested, I was just suggesting alternative possibilities.

src/providers/MeetingProvider/MeetingManager.ts
richnew10: Concise is less important than (a) clear, (b) easy to change safely, (c) easy to test. Your redundant case there is at least very easy to read and easy to tell that it's correct!

You could also try something like:

```typescript
// Audio input and output.
switch (deviceLabels) {
  case DeviceLabels.Audio:
  case DeviceLabels.AudioAndVideo:
    if (this.invokeAudioInputProvider) {
      this.invokeAudioInputProvider();
    }
    if (this.invokeAudioOutputProvider) {
      this.invokeAudioOutputProvider();
    }
    break;
  }
}

// Video input.
switch deviceLabels {
  case DeviceLabels.Video:
  case DeviceLabels.AudioAndVideo: {
    if (this.invokeVideoInputProvider) {
      this.invokeVideoInputProvider();
    }
    break;
  }
}
```

However, the existence of those conditional `invoke` things and this kind of complexity is a good indicator that some part of the design isn't right.

xuesichao: Yes, you are right. I will keep this in mind. Thank you!

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: @richnew10 This is to handle the error and enhance user experience. 
In `listAndSelectDevices` function, we call `audioVideo?.choose*Device()` which may call `getUserMedia()`. This could lead to some error which users don't expect, after we add the device permission control feature.

For example, user wants to join in view-only mode. They pass the `deviceLabels.None` via `mm.join(deviceLabels)`. This will set the `deviceLabelTrigger` as `() => Promise.resolve(new MediaStream());`. And if we don't make this change, when user joins the meeting, there will be errors like below:
image

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
xuesichao: @richnew10 I use observer pattern to achieve it, it works. Should we do the same change for `AudioOutputProvider` ?

src/providers/MeetingProvider/MeetingManager.ts
richnew10: My point is: browsers encourage applications to persist device IDs across application sessions.

So you might need to get labels _once_, and after that you'll just call

```typescript
audioVideo.chooseAudioInputDevice('some-persisted-id');
```

without (a) that device having a label, (b) having called `enumerateDevices`, or (c) having triggered a permission request with a hook.

Such an application will use `DeviceLabels.None` to avoid triggering an arbitrary microphone and video input for no reason, and will then directly choose an input. At no point does that application want `enumerateDevices` being called or the device label hook to result in `getUserMedia` being called.

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
richnew10: Very likely, yes; I can't be sure without seeing the code.

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
xuesichao: Thank you for such a detailed explanation, I appreciate it. I will keep it in mind!

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
xuesichao: @richnew10 Updated the code. I made the change to `AudioOutputProvider` as well. Hope we can merge this by tomorrow. Thank you for your time!

src/providers/DevicesProvider/VideoInputProvider.tsx
xuesichao: Updated the invoke approach to be based on observer pattern. Removed those `invoke*` callbacks.

src/providers/DevicesProvider/VideoInputProvider.tsx
xuesichao: Removed `forceUpdate` here. Cause the `deviceInfoCache` is updated after setting up new `deviceLabelTrigger`.

richnew10:
As mentioned in May, this needs a big pile of developer docs explaining how a builder (a) joins a meeting without requesting audio or video permissions, (b) joins a meeting requesting only camera (silent), (c) joins a meeting requesting only microphone (no video), and (d) how they need to control when permissions are requested in order to perform certain actions, like listing devices or choosing a non-synthetic audio device.

You also need to explain how they can supply a synthetic non-silent audio stream in the case of joining muted.

I also want to make sure that this works correctly when the builder's app saves a device ID and label on the first run (e.g., to local storage) and then if the ID is present skips the entire permissions/label flow, only firing that up if the user tries to pick from a list and has not already selected a device. Getting this right is tricky, but it is a huge part of the value that the SDK offers to builders.

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -104,11 +104,20 @@ const AudioInputProvider: React.FC = ({ children }) => {
...
 
+    const callback = (deviceLabels: DeviceLabels): void => {
+      if (deviceLabels === DeviceLabels.Audio || deviceLabels === DeviceLabels.AudioAndVideo) {
+        initAudioInput();
+      }
+    };
+
+    meetingManager.subscribeToDeviceLabelsChange(callback);
+
     initAudioInput();
```
ltrung, line 115: Isn't this incorrect? We always call initAudioInput() same as before?

src/providers/DevicesProvider/AudioOutputProvider.tsx
```diff
@@ -59,11 +59,20 @@ const AudioOutputProvider: React.FC = ({ children }) => {
...
 
+    const callback = (deviceLabels: DeviceLabels): void => {
+      if (deviceLabels === DeviceLabels.Audio || deviceLabels === DeviceLabels.AudioAndVideo) {
+        initAudioOutput();
+      }
+    };
+
+    meetingManager.subscribeToDeviceLabelsChange(callback);
+
     initAudioOutput();
```
ltrung, line 70: Same here.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -134,7 +138,7 @@ export class MeetingManager implements AudioVideoObserver {
     this.audioVideoObservers = {};
   }
 
-  async join({ meetingInfo, attendeeInfo }: MeetingJoinData) {
+  async join({ meetingInfo, attendeeInfo, deviceLabels = DeviceLabels.AudioAndVideo }: MeetingJoinData) {
```
ltrung, line 141: Can we not use join for this? DeviceLabel is not join data so its weird to include them here.

src/providers/DevicesProvider/AudioInputProvider.tsx
xuesichao: We control the device permission request by `deviceLabelTrigger`. We setup the `deviceLabelTrigger` before the first time `initAUdioInput()` is called.

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: Should I move the `deviceLabels` out of the `MeetingInfo` but still keep it as an optional parameter in `join` function? Or should I have a function `setDeviceLabels` to change  the `deviceLabels` in `meetingManager`? I would prefer the first solution.

xuesichao:
> As mentioned in May, this needs a big pile of developer docs explaining how a builder (a) joins a meeting without requesting audio or video permissions, (b) joins a meeting requesting only camera (silent), (c) joins a meeting requesting only microphone (no video), and (d) how they need to control when permissions are requested in order to perform certain actions, like listing devices or choosing a non-synthetic audio device.

@richnew10 We planned to write a how-to doc later this sprint, which will explain these in detail. Do I need to write them in the CHANGE log?

src/providers/MeetingProvider/MeetingManager.ts
richnew10: The problem is that it's not really suitable as a constructor argument (both because you might not know the value at that time, and also it's awkward to thread a new `MeetingManager` into `MeetingProvider`), and mutating state sucks.

It's not join data, but it's a very clear description of what the builder intends the join to be: with audio or without.

richnew10:
> @richnew10 We planned to write a how-to doc later this sprint, which will explain these in detail. Do I need to write them in the CHANGE log?

Not in the CHANGELOG, but builders need to know how to use this feature, and your developer docs will tell you if you got the interface wrong or if it's buggy! That is, when you write the doc, you will explore behaviors that you have probably not tested in the main demo app.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -416,6 +462,12 @@ export class MeetingManager implements AudioVideoObserver {
     this.publishSelectVideoInputDeviceError();
   };
 
+  invokeDeviceProvider = (deviceLabels: DeviceLabels) => {
+    this.setupDeviceLabelTrigger(deviceLabels);
+    this.deviceLabels = deviceLabels;
+    this.publishDeviceLabelsChange();
```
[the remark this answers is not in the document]
xuesichao: Good point! I am just following the implementation of other observers in the `MeetingManager`, but it seems redundant to me. Will remove the `this.deviceLabels`.

src/providers/DevicesProvider/AudioInputProvider.tsx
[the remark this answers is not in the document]
xuesichao: This naming is because I added `this.deviceLabels` to subscribe to. But the underlying logic here is indeed subscribing to `deviceLabelTrigger` change. Since I removed the `this.deviceLabels` right now, will rename it to `*DeviceLabelTriggerChange`.

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
xuesichao: > You also need to explain how they can supply a synthetic non-silent audio stream in the case of joining muted.

@richnew10 If you are saying the fake audio stream for detecting the user activity, I think JS SDK will handle this automatically. React SDK doesn't seem to support this. Please correct me if I'm wrong.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -269,31 +275,57 @@ export class MeetingManager implements AudioVideoObserver {
...
+        try {
+          const devices = await navigator.mediaDevices.enumerateDevices();
+          const hasVideoInput = devices.some(value => {
+            return value.kind === 'videoinput';
+          });
+
+          const stream = await navigator.mediaDevices.getUserMedia({
+            audio: constraints.audio,
+            video: constraints.video && hasVideoInput,
+          });
```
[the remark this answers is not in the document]
xuesichao: > Do you really need this extra work for `hasVideoInput`? The default implementation is simply:
> 
> ```ts
>             return navigator.mediaDevices.getUserMedia({ audio: true, video: true });
> ```

`hasVideoInput` is the original design, I don't know why someone added it, so I kept it. I will test if we can delete it.

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
richnew10: Phrased differently: when they're joining muted, they will select the null audio device, which will generate synthetic non-silent audio. The point of the docs is to make sure the user knows how to do what they think they want to do…

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -314,7 +346,19 @@ export class MeetingManager implements AudioVideoObserver {
...
+    for (const device of devices) {
+      if (device.label) {
+        if (device.kind === "audioinput") hasAudioInput = true;
+        if (device.kind === "audiooutput") hasAudioOutput = true;
+        if (device.kind === "videoinput") hasVideoInput = true;
+      }
+    }
+
     if (
+      hasAudioInput &&
```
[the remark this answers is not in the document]
xuesichao: @richnew10 This is to handle the error and enhance user experience. 
In `listAndSelectDevices` function, we call `audioVideo?.choose*Device()` which may call `getUserMedia()`. This could lead to some errors which users don't expect, after we merged this PR.

For example, user wants to join in view-only mode. They pass the `deviceLabels.None` via `mm.join(deviceLabels)`. This will set the `deviceLabelTrigger` as `() => Promise.resolve(new MediaStream());`. If we don't make this change, when user joins the meeting, the devices with empty labels will be selected and the below errors pops up:
image

This seems a limit of our SDK. Making this change could address this issue.

xuesichao:
> I also want to make sure that this works correctly when the builder's app saves a device ID and label on the first run (e.g., to local storage) and then if the ID is present skips the entire permissions/label flow, only firing that up if the user tries to pick from a list and has not already selected a device. Getting this right is tricky, but it is a huge part of the value that the SDK offers to builders.

I don't think we can support the case like this due to the limit in SDK. We can select a device with empty label, but there will be errors.

This case seems beyond the scope of this PR. But I do agree it could benefit our customers a lot. I would like to give it a test. But if we can't support it now, should we create an issue in the backlog to enhance it later or some other ideas?

richnew10:
> @richnew10 I don't think we can support the case like this due to the limit in SDK. We can select a device with empty label, but there will be errors.

I don't see why there would be errors if the browser allows it. And it does: doing the following I got a device-specific permissions request at the point of calling `getUserMedia`.

Screen Shot 2021-06-08 at 12 06 54 PM

> This case seems beyond the scope of this PR. But I do agree it could benefit our customers a lot. I would like to give it a test. But if we can't support it now, should we create an issue in the backlog to enhance it later or some other ideas?

The point of me bringing it up is that with your change here you now make it possible for builders to avoid the initial permissions prompt before enumerating devices, and thus immediately after release builders can do exactly what I described. We need to know what happens when they do, and if it's buggy we need to fix it or document it before release.

src/providers/DevicesProvider/AudioOutputProvider.tsx
```diff
@@ -59,11 +59,20 @@ const AudioOutputProvider: React.FC = ({ children }) => {
       }
     }
 
+    const callback = (deviceLabels: DeviceLabels): void => {
+      if (deviceLabels === DeviceLabels.Audio || deviceLabels === DeviceLabels.AudioAndVideo) {
+        initAudioOutput();
+      }
+    };
```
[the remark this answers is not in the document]
xuesichao: Yeah, you are right. Since we the device permission request only depends on the `deviceLabelTrigger` now. We could simply use `init*()` as callback here. Addressed.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -269,31 +275,57 @@ export class MeetingManager implements AudioVideoObserver {
...
+    let callback: () => Promise<MediaStream>;
 
-        const stream = await navigator.mediaDevices.getUserMedia({
-          audio: true,
-          video: hasVideoInput
-        });
+    if (typeof deviceLabels === "function") {
+      callback = deviceLabels;
+    } else if (deviceLabels === DeviceLabels.None) {
+      callback = () => Promise.resolve(new MediaStream());
```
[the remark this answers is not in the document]
xuesichao: Sorry, I'm not clear what do you mean here. This code block is used only in `setupDeviceLabelTrigger` in `Meeting Manager` now, why do you say it's repeated? Do you mean that `setupDeviceLabelTrigger` is called repeatedly?

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -269,31 +275,57 @@ export class MeetingManager implements AudioVideoObserver {
...
+        case DeviceLabels.Video:
+          constraints.video = true;
+          break;
+        case DeviceLabels.AudioAndVideo:
+          constraints.audio = true;
+          constraints.video = true;
+          break;
+        default:
+          constraints.audio = true;
+          constraints.video = true;
```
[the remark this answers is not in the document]
xuesichao: My VS Code does not provide special tips here. I agree with you. Removed the default case here.

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
xuesichao: @richnew10 
There are two reason why I made it like this:
1. Enhance user experience
If user joins a meeting without a kind of device, user may expect the 'None' is selected for this kind of device.
Without this change, the device with empty label (it is the first one) may be selected. I thought this could be a bad behavior.

2. Avoid error
It seems the error is **only about audio device**. Whenever user ignore the audio device(`DeviceLabels.Video`, `DeviceLabels.None`), without this change, the above error pops up in Chrome, Firefox and Safari. 
I think if we support user to select the device kind they want to trigger device permission prompt, we shouldn't show any related error.
Besides, the Firefox will still trigger the device permission prompt for audio device only. This seems to be caused by `chooseAudioInputDevice` => `chooseInputIntrinsicDevice()` => 
    ```ts
    const proposedConstraints = this.calculateMediaStreamConstraints(kind, device);
    newDevice.stream = await navigator.mediaDevices.getUserMedia(proposedConstraints);
    ```
    You can check the code [here](https://github.com/aws/amazon-chime-sdk-js/blob/952ad05d94980aeff09693ab2e9efd6df6066661/src/devicecontroller/DefaultDeviceController.ts#L1396). I don't know how to address this issue.

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
richnew10: Just drop this at the top level.

```typescript
function noOpDeviceLabelHook(): Promise<MediaStream> {
  return Promise.resolve(new MediaStream());
}
```

That's guaranteed not to close over any locals in the enclosing scope, and you're guaranteed to only ever have a single block of JITted code that does this job.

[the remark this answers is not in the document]
richnew10: The device label hook function/closure sticks around for the life of the meeting and longer, and so it's good practice to make sure we don't accidentally have it capture an object graph and cause a leak.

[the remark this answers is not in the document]
xuesichao: Oh I get it! I thought you were talking about these four lines of code. Thanks!

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: Keep the current solution.

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
richnew10: There is clearly a bug here, and it might be related to the misordered init we've observed. Let's follow up on this.

richnew10 (dismissed):
I'm happy with this if @ltrung is.

demo/meeting/src/containers/DynamicMeetingControls/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+const DynamicMeetingControls = () => {
+  const { toggleNavbar, closeRoster, showRoster } = useNavigation();
+  const { isUserActive } = useUserActivityState();
+  const permission = useDevicePermissionStatus();
+
+  const handleToggle = () => {
+    if (showRoster) {
+      closeRoster();
+    }
```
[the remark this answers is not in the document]
xuesichao: I can't image a common use case of the reverse behavior now. If there is any you could see, please let me know, I would love to improve it. This component is only in meeting demo now, the main purpose here is to build a view-only experience and allow QA to test this change.
The requirements for the correct use of this component are a bit strict - it works only if the device permission is unset. In any other cases, the it can't change. This is due to a limitation in our SDK, it would be great we could have a Provider to reveal the device permission status, but it's hard to do this now I think.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -269,31 +277,52 @@ export class MeetingManager implements AudioVideoObserver {
...
-  setupDeviceLabelTrigger(): void {
-    const callback = async (): Promise<MediaStream> => {
-      this.devicePermissionStatus = DevicePermissionStatus.IN_PROGRESS;
-      this.publishDevicePermissionStatus();
-      try {
-        const devices = await navigator.mediaDevices.enumerateDevices();
-        const hasVideoInput = devices.some(value => {
-          return value.kind === 'videoinput';
-        });
+  setupDeviceLabelTrigger(deviceLabels: DeviceLabels | DeviceLabelTrigger = DeviceLabels.AudioAndVideo): void {
```
ltrung, line 280: Isn't this a breaking change since you modify a signature of a non-private method?

xuesichao: I set the default value of this function. And the default setting remains as original.

src/providers/MeetingProvider/MeetingManager.ts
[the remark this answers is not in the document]
xuesichao: Recorded the bug. I will keep my eye on it.

demo/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -37,8 +39,10 @@ const MeetingForm: React.FC = () => {
   const [nameErr, setNameErr] = useState(false);
   const [region, setRegion] = useState(appRegion);
   const [isLoading, setIsLoading] = useState(false);
+  const [isSelected, setIsSelected] = useState(false)
```
ltrung, line 42: nit: Can we rename this to something that says what it actually does? IsSpectatorModeSelected or something

xuesichao: Yes, will do!

demo/meeting/src/views/Meeting/index.tsx
```diff
@@ -26,7 +28,10 @@ const MeetingView = () => {
             className="videos"
             noRemoteVideoView={<MeetingDetails />}
           />
-          <MeetingControls />
+          {props.mode === MeetingMode.Spectator ?
+            <DynamicMeetingControls /> :
```
ltrung, line 32: Do we still need meeting control then? Cant we always use DynamicMeetingControls?

xuesichao: We can't always use `DynamicMeetingControls`. The requirements for the correct use of this component are a bit strict - it works only if the device permission is unset after user've already joined the meeting. In any other cases, it can't transform. This is due to a limitation in our SDK, it would be great we could have a Provider to reveal the device permission status, but it's hard to do this now I think. Any suggestion?

xuesichao:
> > ```
> > * Click `Device` button ⮕ Device permission request prompts ⮕ click allow and `MeetingControl` works as normal ⮕ User can select devices in `MeetingControl`
> > ```
> 
> Seem like the format for testing is messed up.

Updated.

demo/meeting/src/types/index.ts
```diff
@@ -34,3 +34,17 @@ export type ContentShareControlContextType = {
   toggleContentShare: () => Promise<void>;
   togglePauseContentShare: () => void;
 };
+
+export enum DeviceLabels {
```
ltrung, line 38: nit: Cant we reuse the enum from the library?

demo/meeting/src/types/index.ts
```diff
@@ -34,3 +34,17 @@ export type ContentShareControlContextType = {
...
+  Video,
+  AudioAndVideo,
+};
+
+export enum MeetingMode {
+  Spectator,
+  Attendee,
+};
+
+export type DeviceLabelTrigger = () => Promise<MediaStream>;
```
ltrung, line 50: Is this used?

xuesichao: It's not used now. But it could be used if we want to test the input `deviceLabels` of `DeviceLabelTrigger` type.

Deleted it in meeting demo and exported the one in the library. Thanks!

demo/meeting/src/types/index.ts
xuesichao: Yes we can. I forgot to export this enum from the library. Deleted now.

michhyun1 (approved):

PR #561, opened 2021-07-13
Fix the issue that MeetingManager selects unexpected devices

xuesichao:
**Issue #:** 
Sometimes MeetingManager selects unexpected device. 

For example, when calling meetingManager.join(DeviceLabels.Audio) to join a meeting, Firefox/Safari will first ask the permission for microphone. After the user grants permission for it, Firefox/Safari will ask the permission for the camera devices.

The second permission request is unexpected. And the cause of it is we use the wrong check to determine whether or not to select the device.

For example, user pass `DeviceLabels.Audio`, then the call stack is like:
1. call `mm.join(DeviceLabels.Audio)`
2. call `mm.setupDeviceLabelTrigger()`, which sets `deviceLabelTrigger` to `getUserMedia({audio: true})`
3. call `mm.updateDeviceLists()`, which triggers the `deviceLabelTrigger`
    * First prompt appear
    * Here Firefox and Safari also get the device label, device Id of video devices as well
4. pass the check and set `hasVideoInput` to true [here](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/d3a2570b7a90f98b0552c52ac23bc8c2a457b7ff/src/providers/MeetingProvider/MeetingManager.ts#L352-L354)
5. call [`mm.publishSelectedVideoInputDevice()`](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/d3a2570b7a90f98b0552c52ac23bc8c2a457b7ff/src/providers/MeetingProvider/MeetingManager.ts#L396-L397), which selects the video device using its device Id
    * Second prompt appear (Firefox and Safari)
    * Cause in step 3 the device Id is exposed

**Description of changes:**
- Add optional parameter `deviceLabels: DeviceLabels | DeviceLabelTrigger` in `meetingManager.listAndSelectDevices()` to let builder indicate the type of devices they want to select.
- The `listAndSelectDevices()` method in `MeetingManager` now selects the devices based on its parameter `deviceLabels` instead of the existence of device label.

**Testing**
Yes

In Firefox, Safari, Chrome, try to pass the following parameters in `meetingManager.join()` to join the meeting, and check if the expected behavior can be observed.
* `DeviceLabels.AudioAndVideo`
* `DeviceLabels.Audio`
* `DeviceLabels.Video`
* `DeviceLabels.None`
* Customized `deviceLabelTrigger`

Yes


src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -349,12 +331,11 @@ export class MeetingManager implements AudioVideoObserver {
   async listAndSelectDevices(): Promise<void> {
     await this.updateDeviceLists();
 
-    const hasAudioInput = this.audioInputDevices?.some(value => value.label) || false;
-    const hasAudioOutput = this.audioOutputDevices?.some(value => value.label) || false;
-    const hasVideoInput = this.videoInputDevices?.some(value => value.label) || false;
+    const hasAudioPermission = !!this.constraints.audio;
+    const hasVideoPermission = !!this.constraints.video;
```
xuesichao, line 335: We can not check the existence of device `label` to determine whether or not to select devices. The design of device permission is different in browsers. Firefox and Safari exposes the device label and device id of camera/microphone even if they only obtain the permission of microphone/ camera.

To fit this difference, we need to store the `constraints` of `getUserMedia`, and do the check based on it.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -463,6 +444,22 @@ export class MeetingManager implements AudioVideoObserver {
     this.publishDeviceLabelTriggerChange();
   };
 
+  setConstraints = (deviceLabels: DeviceLabels) => {
```
xuesichao, line 447: We cannot tell what kind of devices builder want to ask for permission in their customized `deviceLabelTrigger`.

Expose this function so when user passes the `deviceLabels: DeviceLabelTrigger` to `meetingManager.join()`, they can set the value of `constraints` to align it.

src/providers/MeetingProvider/MeetingManager.ts
richnew10: What happens when the user declines permissions for one or both constraint?

src/providers/MeetingProvider/MeetingManager.ts
richnew10: This is a breaking API change — a working app will stop working unless they now call this method, right?

This is all an indication that this design is incorrect.

When the hook runs, it gets a stream. The stream's tracks will include audio and/or video. Can we use that knowledge?

richnew10 (changes requested):

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: > What happens when the user declines permissions for one or both constraint?

I'm not clear about the 'constraint'` you referred here. I assume you are talking about the device permission prompt.

For example, if users call `mm.join(DeviceLabels.Audio)` to join a meeting, SDK should let browser trigger the device permission prompt **ONLY for mic**. But in Firefox/Safari, there will be two permission prompts, the first one ask for mic permission, the second ask for camera permission, the latter will only appear if users allow the former.

So for your question, if users decline the first one, the second one won't appear, and there is no way to decline the both.

The problem here is, the second device permission prompt should not appear in anyway, this is a bug or incorrect design of PR #499, in which we allow users pass a `deviceLabels: DeviceLabels | DeviceLabelTrigger` to control the device permission logic in SDK.

This PR is to fix this bug or incorrect design of PR #499.

richnew10: In Firefox, `getUserMedia({audio: true, video: true})` produces a combined permissions prompt. The user can click 'Block', or the decline can be remembered, and the entire `gUM` call will be rejected.

We should not be triggering two prompts in any circumstance in Firefox or Chrome. I believe Safari does permission prompt coalescing, but haven't tested.

It's also possible to have no audio devices (typically on Linux) or no camera.

Why does a second prompt appear?

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: > This is a breaking API change — a working app will stop working unless they now call this method, right?

I don't think this is a breaking API, there are 4 cases for `deviceLabelTrigger` (same for the customized `deviceLabelTrigger`):

| Case | DeviceLabelTrigger                                                      | No `setConstraints` calling                | With  `setConstraints` calling |
| ---- | ----------------------------------------------------------------------- | ------------------------------------------ | ------------------------------ |
| 1.   | `async () => await navigator.getUserMedia({audio: true});`              | First ask for mic, then camera (incorrect) | Only one for mic(fixed)        |
| 2.   | `async () => await navigator.getUserMedia({video: true});`              | First ask for camera, then mic (incorrect) | Only one for camera(fixed)     |
| 3.   | `async () => await navigator.getUserMedia({audio: true, video: true});` | Only one for mic and camera                | Only one for mic and camera    |
| 4.   | `() => Promise.resolve(new MediaStream());`                             | None                                       | None                           |

With calling `setConstraints` a working App will still work as before, while a buggy App now get fixed.

> When the hook runs, it gets a stream. The stream's tracks will include audio and/or video. Can we use that knowledge?

I don't think we can use that. In case 1 and 2:
1. The first device permission prompt is triggered by `audioVideo.list*Devices()` -> `audioVideo.listDevicesOfKind()` -> `deviceLabelTrigger`.
    * The stream we get. We can't use this to determine the below one, or it could be a big change.
2. The second device permission prompt is triggered by `audioVide.choose*Device(deviceId)`;

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: > Why does a second prompt appear?

For example: 
1. users call `mm.join(DeviceLabels.Audio)`
2. `mm.setupDeviceLabelTrigger()` set `deviceLabelTrigger` to `getUserMedia({audio: true})`
3. `mm.updateDeviceLists()` which triggers the `deviceLabelTrigger`
    * First prompt appear
    * Here Firefox and Safari also get the device label, device Id of video devices as well
4. pass the check and set `hasVideoInput` to true [here](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/d3a2570b7a90f98b0552c52ac23bc8c2a457b7ff/src/providers/MeetingProvider/MeetingManager.ts#L352-L354)
5. select the video device using its device Id
    * Second prompt appear (Firefox and Safari)
    * Cause in step 3 the device Id is exposed

src/providers/MeetingProvider/MeetingManager.ts
richnew10: What's the behavior for a customer's app upgrading from an older version of the component library?

If I understand correctly, what you're observing here is that the browser is remembering the requested permission, and so the issue is that the request for the real media stream (chooseDevice) triggers a second permission prompt if it is not a subset of a previously requested permission. That's actually OK.

The behavior in the table you wrote above should depend on whether you're calling `chooseAudioDevice` or `chooseVideoDevice`.

I think this is going in the wrong direction: if a customer specifies a hook, it means they want control.

If I use a device label trigger of `() => navigator.getUserMedia({audio: true})` and then call `chooseAudioInputDevice`, what is causing a camera request to occur?

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -110,6 +110,10 @@ export class MeetingManager implements AudioVideoObserver {
 
   simulcastEnabled: boolean = false;
 
+  hasAudioPermission: boolean = false;
+
+  hasVideoPermission: boolean = false;
+
```
xuesichao, line 116: I can't find any better idea to avoid this. We need to access these variables in `setupDeviceLabelTrigger` and `listAndSelectDevices` unless we modify these functions. Is it acceptable or any suggestions?

xuesichao: Initialize these two variables to `false` to handle the corner case like `getUserMedia({audio:false, video: false})`, in this case, the `deviceLabelTrigger` stops before executing the code to set their values.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -143,7 +147,7 @@ export class MeetingManager implements AudioVideoObserver {
     this.audioVideoObservers = {};
   }
 
-  async join({ meetingInfo, attendeeInfo, deviceLabels = DeviceLabels.AudioAndVideo, eventReporter}: MeetingJoinData) {
+  async join({ meetingInfo, attendeeInfo, deviceLabels = DeviceLabels.AudioAndVideo, eventReporter }: MeetingJoinData) {
```
richnew10, line 150: Thank you for tidying as you go! 🥇

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -286,9 +290,19 @@ export class MeetingManager implements AudioVideoObserver {
...
+        return stream;
+      };
     } else if (deviceLabels === DeviceLabels.None) {
-      callback = noOpDeviceLabelHook;
+      callback = async (): Promise<MediaStream> => {
+        const stream = await noOpDeviceLabelHook();
+        this.hasAudioPermission = !!stream.getAudioTracks().length;
+        this.hasVideoPermission = !!stream.getVideoTracks().length;
+        return stream;
+      };
```
richnew10, line 305: By definition the new fields will be false here — an empty stream has no tracks.

What's supposed to happen if the builder uses a hook (say, camera-only), and then switches to `DeviceLabels.None`? Do we keep `hasVideoPermission = true`, or do we set it to false? That applies more generally to the case when they use any non-subset value, such as audio and then video.

src/providers/MeetingProvider/MeetingManager.ts
richnew10: Yeah, this is fine. 

Really these are `wasAudioPermissionEverApparentlyGrantedOrTheBuilderPassedUsASyntheticAudioTrack`, or `weSawAnAudioTrackLastTimeWeRanAHook`, so you might want to comment about exactly what these mean and when the values change. 

They don't really mean whether we have permissions — we only know that at the exact moment we request an input stream by calling `getUserMedia`.

It's important to remember that customers can use the SDK without *ever* calling `getUserMedia` (e.g., by supplying their own synthetic video and audio streams), or they can use the SDK and manage input streams themselves by directly calling `getUserMedia` and then choosing those streams as input devices. We should make sure that `MeetingManager` behaves appropriately in those cases.

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: #### Case 1

1. call `mm.join(DeviceLabels.Audio)`, and redirect to device page
    * `deviceLabelTrigger` is triggered, browser ask permission for **microphone**
    *  Allow the permission
    * `hasAudioPermission = true`, `hasVideoPermission = false`
2. redirect to meeting page
3. call `mm.invokeDeviceProvider(DeviceLabels.Video)`
    * the `mm.setupDeviceLabelTrigger(DeviceLabels.Video)` is called
        * then `deviceLabelTrigger` is triggered, browser ask permission for **camera**
            * `hasAudioPermission = false`, `hasVideoPermission = true`
            * `hasAudioPermission` is set to false, cause this time returned `stream` has no `AudioTrack`

#### Case 2

1. call `mm.join(DeviceLabels.Audio)`, and redirect to device page
    * `deviceLabelTrigger` is triggered, browser ask permission for **microphone**
        *  Allow the permission
        * `hasAudioPermission = true`, `hasVideoPermission = false`
2. redirect to meeting page
3. call `mm.invokeDeviceProvider(DeviceLabels.None)`
    * the `mm.setupDeviceLabelTrigger(DeviceLabels.None)` is called
        * (Chrome) the `noOpDeviceLabelHook` `deviceLabelTrigger` is triggered, no permission prompt
            * `hasAudioPermission = false`, `hasVideoPermission = false`
            * update the value, cause returned `stream` has no tracks
        * (Firefox/Safari) `deviceLabelTrigger` is **NOT** triggered
            * `hasAudioPermission = true`, `hasVideoPermission = false`
            * keep the previous value

xuesichao: I would say the essence of these two variables is a helper variable **ONLY** used to check whether or not to select one kind of devices in `listAndSelectDevices()`. But to share these variables accross `setupDeviceLabelTrigger` and `listAndSelectDevices`, we have to add them as a property of `MeetingManager`, thus we need to make their value meaningful. 

It feels like a bad design to me. I think we have following solutions:
* Should we add comments to explain the responsibility of these two variables to address this issue?
* Should we do some extra work to keep their value meaningful? I don't think it's good to keep the browser's property in `MeetingManager` as you mentioned. Especially the behavior of browser varies.
* Avoid adding them as a property, and find another way to do it.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -346,15 +346,33 @@ export class MeetingManager implements AudioVideoObserver {
...
+    // If `deviceLabels` is of `DeviceLabelTrigger` type, no device will be selected.
+    // In this case, you need to handle the device selection yourself.
+    if (typeof deviceLabels === 'function') return;
+
+    let isAudioDevicesRequested: boolean = false;
+    let isVideoDevicesRequested: boolean = false;
+
+    switch (deviceLabels) {
+      case DeviceLabels.None:
+        break;
```
xuesichao, line 361: For the readability, I keep this empty case here to cover all the cases. Not sure if it's reasonable to do this.

richnew10: Closed/exhaustive sets are best — this is good Rust style, so I support it here!

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -346,15 +346,33 @@ export class MeetingManager implements AudioVideoObserver {
...
 
-    const hasAudioInput = this.audioInputDevices?.some(value => value.label) || false;
-    const hasAudioOutput = this.audioOutputDevices?.some(value => value.label) || false;
-    const hasVideoInput = this.videoInputDevices?.some(value => value.label) || false;
+    // If `deviceLabels` is of `DeviceLabelTrigger` type, no device will be selected.
+    // In this case, you need to handle the device selection yourself.
+    if (typeof deviceLabels === 'function') return;
+
+    let isAudioDevicesRequested: boolean = false;
+    let isVideoDevicesRequested: boolean = false;
```
richnew10, line 357: Nit: you can un-plural these (`isAudioDeviceRequested`).

richnew10 (approved):

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: Makes sense. Updated them to singular form. Thank you for your help.

richnew10 (approved):
I love how short this ended up!

PR #575, opened 2021-07-23
Add support to customize the downlink policy of the meeting

xuesichao:
**Issue #:** 

User wants to set the `Logger` and `VideoDownlinkBandwidthPolicy` for the meeting session.

**Description of changes:**
* Add an option parameter in `MeetingConfig` to pass in a `VideoDownlinkBandwidthPolicy` instance to customize the downlink policy of the meeting
* Add an option parameter in `MeetingConfig` to pass in a `Logger` instance to customize the `Logger` of the meeting

**Testing**
Yes

Check the console logger and downlink policy is set correctly.

Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add optional parameter `deviceLabels: DeviceLabels | DeviceLabelTrigger` in `meetingManager.listAndSelectDevices()` to let builder indicate the type of devices they want to select.
-- Add several how-to docs in storybook.
+- Add optional parameter `logger` and `videoDownlinkBandwidthPolicy` in `ManagerConfig` to pass in the `Logger` object and `VideoDownlinkBandwidthPolicy` object that you want to be used in the meeting session.
+- [Doc] Add several how-to docs in storybook.
 
 ### Changed
 
--- a/src/providers/MeetingProvider/MeetingManager.ts
+++ b/src/providers/MeetingProvider/MeetingManager.ts
@@ -14,7 +14,9 @@ import {
   AudioVideoObserver,
   MultiLogger,
   MeetingSessionPOSTLogger,
-  EventReporter
+  EventReporter,
+  VideoDownlinkBandwidthPolicy,
+  Logger,
 } from 'amazon-chime-sdk-js';
 
 import {
@@ -110,15 +112,26 @@ export class MeetingManager implements AudioVideoObserver {
 
   simulcastEnabled: boolean = false;
 
-  constructor(config: ManagerConfig) {
-    this.logLevel = config.logLevel;
+  videoDownlinkBandwidthPolicy: VideoDownlinkBandwidthPolicy | undefined;
 
+  logger: Logger | undefined;
+
+  constructor(config: ManagerConfig) {
     if (config.simulcastEnabled) {
       this.simulcastEnabled = config.simulcastEnabled;
     }
 
-    if (config.postLogConfig) {
-      this.postLoggerConfig = config.postLogConfig;
+    if (config.logger) {
+      this.logger = config.logger;
+    } else {
+      this.logLevel = config.logLevel;
+      if (config.postLogConfig) {
+        this.postLoggerConfig = config.postLogConfig;
+      }
+    }
+
+    if (config.videoDownlinkBandwidthPolicy) {
+      this.videoDownlinkBandwidthPolicy = config.videoDownlinkBandwidthPolicy;
     }
   }
 
@@ -141,6 +154,8 @@ export class MeetingManager implements AudioVideoObserver {
     this.meetingStatus = MeetingStatus.Loading;
     this.publishMeetingStatus();
     this.audioVideoObservers = {};
+    this.videoDownlinkBandwidthPolicy = undefined;
+    this.logger = undefined;
   }
 
   async join({ meetingInfo, attendeeInfo, deviceLabels = DeviceLabels.AudioAndVideo, eventReporter }: MeetingJoinData) {
@@ -149,16 +164,18 @@ export class MeetingManager implements AudioVideoObserver {
       attendeeInfo
     );
 
+    this.meetingRegion = meetingInfo.MediaRegion;
+    this.meetingId = this.configuration.meetingId;
+
     if (this.simulcastEnabled) {
       this.configuration.enableUnifiedPlanForChromiumBasedBrowsers = true;
       this.configuration.enableSimulcastForUnifiedPlanChromiumBasedBrowsers = true;
     };
 
-    this.meetingRegion = meetingInfo.MediaRegion;
-    this.meetingId = this.configuration.meetingId;
     if (eventReporter) {
       this.eventReporter = eventReporter;
     }
+
     await this.initializeMeetingSession(this.configuration, deviceLabels);
   }
 
@@ -198,7 +215,12 @@ export class MeetingManager implements AudioVideoObserver {
     configuration: MeetingSessionConfiguration,
     deviceLabels: DeviceLabels | DeviceLabelTrigger = DeviceLabels.AudioAndVideo,
   ): Promise<any> {
-    const logger = this.createLogger(configuration);
+    const logger = this.logger ? this.logger : this.createLogger(configuration);
+
+    if (this.videoDownlinkBandwidthPolicy) {
+      configuration.videoDownlinkBandwidthPolicy = this.videoDownlinkBandwidthPolicy;
+    }
+
     const deviceController = new DefaultDeviceController(logger);
     this.meetingSession = new DefaultMeetingSession(
       configuration,
--- a/src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
+++ b/src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
@@ -65,6 +65,7 @@ const Root = () => {
 ### Prop values
 
 #### logLevel
+
 The level of logging you want with your meeting session. It is defined by the [LogLevel](https://aws.github.io/amazon-chime-sdk-js/enums/loglevel.html) enum in `amazon-chime-sdk-js`.
 
 ```javascript
@@ -77,7 +78,8 @@ enum LogLevel {
 }
 ```
 
-#### postLogger
+#### postLogConfig
+
 Configuration for setting up a [MeetingSessionPOSTLogger](https://aws.github.io/amazon-chime-sdk-js/classes/meetingsessionpostlogger.html).
 
 ```javascript
@@ -91,11 +93,50 @@ interface PostLogConfig = {
 ```
 
 #### simulcastEnabled
+
 To enable simulcast for the meeting pass the `simulcastEnabled` prop with value set to `true`. By default, it is `false`.
 For more information on Simulcast, check Amazon Chime JS SDK [simulcast guide](https://github.com/aws/amazon-chime-sdk-js/blob/master/guides/05_Simulcast.md).
 
+#### logger
+
+The `Logger` object that you want to be used in the meeting session.
+If you pass in a `Logger` object using this parameter, the `MeetingManager` will use this object instead of creating a logger based on `logLevel` and `postLogConfig` to initialize the meeting session.
+
+#### videoDownlinkBandwidthPolicy
+
+The `VideoDownlinkBandwidthPolicy` object that you want to be used in the meeting session.
+For more information on `VideoDownlinkBandwidthPolicy`, check Amazon Chime JS SDK [priority downlink policy guide](https://aws.github.io/amazon-chime-sdk-js/modules/prioritybased_downlink_policy.html).
+
+For example, to use the `VideoPriorityBasedPolicy` in the meeting:
+
+```jsx
+import {
+  LogLevel,
+  ConsoleLogger,
+  VideoPriorityBasedPolicy,
+} from 'amazon-chime-sdk-js';
+import { MeetingProvider } from 'amazon-chime-sdk-component-library-react';
+
+const Root = () => {
+  const logger = new ConsoleLogger('SDK', LogLevel.INFO);
+  const videoDownlinkBandwidthPolicy = new VideoPriorityBasedPolicy(logger);
+
+  const meetingConfig = {
+    logger,
+    videoDownlinkBandwidthPolicy,
+  };
+
+  return (
+    <MeetingProvider {...meetingConfig}>
+      <MyApp />
+    </MeetingProvider>
+  );
+};
+```
+
 #### meetingManager
-There may be use cases to re-use the same `MeetingManager` instance accross multiple different `MeetingProvider`s.
+
+There may be use cases to re-use the same `MeetingManager` instance across multiple different `MeetingProvider`s.
 Hence, you can create a new `MeetingManager` instance and then pass it as a prop to your `MeetingProvider`s. If not passed,
 a new `MeetingManager` instance will be created internally with each `MeetingProvider` when rendered; and, 
 you will get the `MeetingManager` instance associated with that particular `MeetingProvider` when `useMeetingManager` hook is called.
--- a/src/providers/MeetingProvider/types.ts
+++ b/src/providers/MeetingProvider/types.ts
@@ -1,7 +1,7 @@
 // Copyright 2020-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { EventReporter, LogLevel } from 'amazon-chime-sdk-js';
+import { EventReporter, LogLevel, VideoDownlinkBandwidthPolicy, Logger } from 'amazon-chime-sdk-js';
 import { DeviceLabels, DeviceLabelTrigger } from '../../types';
 
 export enum DevicePermissionStatus {
@@ -44,4 +44,6 @@ export interface ManagerConfig {
   logLevel: LogLevel;
   postLogConfig?: PostLogConfig;
   simulcastEnabled?: boolean;
+  logger?: Logger;
+  videoDownlinkBandwidthPolicy?: VideoDownlinkBandwidthPolicy;
 }
--- a/src/providers/MeetingProvider/docs/MeetingManager.stories.mdx
+++ b/src/providers/MeetingProvider/docs/MeetingManager.stories.mdx
@@ -34,7 +34,7 @@ const MyApp = () => {
 
 ## Constructor
 
-Accepts a config object of `logLevel`, `postLogConfig`, and `simulcastEnabled` that will be used when initializing a meeting session. See [MeetingProvider](/docs/sdk-providers-meetingprovider--page) documentation for more details.
+Accepts a config object of `logLevel`, `postLogConfig`, `simulcastEnabled`, `logger` and `videoDownlinkBandwidthPolicy` that will be used when initializing a meeting session. See [MeetingProvider](/docs/sdk-providers-meetingprovider--page) documentation for more details.
 
 ## Interface
 
@@ -53,6 +53,10 @@ interface MeetingJoinData {
 
   // The response of calling chime:CreateAttendee from your server.
   attendeeInfo: any;
+
+  // The kind of device for which the browser requests permission.
+  // Or override the default device label trigger in the Amazon Chime SDK for JavaScript.
+  deviceLabels?: DeviceLabels | DeviceLabelTrigger;
 
   // Override the default event reporter in the Amazon Chime SDK for JavaScript.
   eventReporter: EventReporter;
--- a/src/providers/MeetingProvider/index.tsx
+++ b/src/providers/MeetingProvider/index.tsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React, { useContext, useState, createContext } from 'react';
-import { LogLevel } from 'amazon-chime-sdk-js';
+import { Logger, LogLevel, VideoDownlinkBandwidthPolicy } from 'amazon-chime-sdk-js';
 
 import MeetingManager from './MeetingManager';
 import { PostLogConfig } from './types';
@@ -22,10 +22,18 @@ interface Props {
   postLogConfig?: PostLogConfig;
   /** Whether or not to enable simulcast for the meeting session */
   simulcastEnabled?: boolean;
+  /** The `Logger` object you want to use in meeting session.
+   * If you pass in a `Logger` object using this parameter, 
+   * the `MeetingManager` will use this object instead of creating a logger 
+   * based on `logLevel` and `postLogConfig` to initialize the meeting session.
+   */
+  logger?: Logger;
+  /** The `VideoDownlinkBandwidthPolicy` object you want to use in meeting session */
+  videoDownlinkBandwidthPolicy?: VideoDownlinkBandwidthPolicy;
   /** Pass a `MeetingManager` instance if you want to share this instance 
    * across multiple different `MeetingProvider`s. This approach has limitations.
    * Check `meetingManager` prop documentation for more information.
-  */
+   */
   meetingManager?: MeetingManager;
 }
 
@@ -35,11 +43,13 @@ export const MeetingProvider: React.FC<Props> = ({
   logLevel = LogLevel.WARN,
   postLogConfig,
   simulcastEnabled = false,
+  logger,
+  videoDownlinkBandwidthPolicy,
   meetingManager: meetingManagerProp,
   children,
 }) => {
   const [meetingManager] = useState(
-    () => meetingManagerProp || new MeetingManager({ logLevel, postLogConfig, simulcastEnabled })
+    () => meetingManagerProp || new MeetingManager({ logLevel, postLogConfig, simulcastEnabled, logger, videoDownlinkBandwidthPolicy })
   );
 
   return (


demo/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -35,19 +33,19 @@ const MeetingForm: React.FC = () => {
...
   const [name, setName] = useState('');
   const [nameErr, setNameErr] = useState(false);
   const [region, setRegion] = useState(appRegion);
   const [isLoading, setIsLoading] = useState(false);
-  const [isSpectatorModeSelected, setIsSpectatorModeSelected] = useState(false)
+  const [isSpectatorModeSelected, setIsSpectatorModeSelected] = useState(false);
   const { errorMessage, updateErrorMessage } = useContext(getErrorContext());
   const history = useHistory();
   const { setMeetingMode } = useAppState();
-  const [enableEventReporting, setEnableEventReporting] = useState<boolean>(true);
```
xuesichao, line 50: To keep the code style consistent.

demo/meeting/src/containers/MeetingForm/index.tsx
```diff
@@ -75,10 +73,12 @@ const MeetingForm: React.FC = () => {
...
-        deviceLabels: isSpectatorModeSelected === true ? DeviceLabels.None : DeviceLabels.AudioAndVideo,
-        ...(enableEventReporting ? {} : {
-          eventReporter: new NoOpEventReporter()
-        })
+        deviceLabels: isSpectatorModeSelected
+          ? DeviceLabels.None
+          : DeviceLabels.AudioAndVideo,
+        ...(enableEventReporting
+          ? {}
+          : { eventReporter: new NoOpEventReporter() })
```
xuesichao, line 81: To improve the readability.

demo/meeting/src/meetingConfig.ts
```diff
@@ -20,12 +20,17 @@ const postLogConfig = {
   batchSize: 85,
   intervalMs: 2000,
   url: `${BASE_URL}logs`,
-  logLevel: SDK_LOG_LEVELS.info
+  logLevel: SDK_LOG_LEVELS.info,
+};
+
+const videoDownlinkBandwidthPolicyWrapper = (logger: Logger): VideoPriorityBasedPolicy => {
```
xuesichao, line 26: To create a downlink policy, we need to pass in a `Logger`. In React SDK we create the `Logger` in `MeetingManager` `initializeMeetingSession` function. To allow user customize the downlink policy we need to wrap it.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -238,6 +221,38 @@ export class MeetingManager implements AudioVideoObserver {
...
+      }
+
+      this.audioVideo.stop();
+      this.audioVideo.removeObserver(this.audioVideoObservers);
+    }
+    this.initializeMeetingManager();
+    this.publishAudioVideo();
+    this.publishActiveSpeaker();
+  }
+
```
xuesichao, line 255: Change the order of the function for better readability.

demo/meeting/src/meetingConfig.ts
devalevenkatesh: Can we not pass the custom `VideoPriorityBasedPolicy` directly?

Also, if this `Logger` is created separately here, we would need a way to use this same while creating the meeting session right? Otherwise, there would be separate loggers one created in the default meeting session creation and this one.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -110,6 +112,8 @@ export class MeetingManager implements AudioVideoObserver {
 
   simulcastEnabled: boolean = false;
 
+  videoDownlinkBandwidthPolicyWrapper: ((logger: Logger) => VideoDownlinkBandwidthPolicy) | null = null;
```
xuesichao, line 115: @devalevenkatesh @simmkyu Could you please give me some feedback on this design?

To initialize a `VideoDowninkBandwidthPolicy`, we must pass in a `logger`.
```js
new VideoPriorityBasedPolicy(logger);
```
The React SDK is different from JS SDK. In React SDK, we create `logger` in the `initializeMeetingSession` in [`MeetingManager` here](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/ae5874591d4895a182a826384a84bfe9ec378074/src/providers/MeetingProvider/MeetingManager.ts#L201).

```js
async initializeMeetingSession(): Promise<any> {
  const logger = this.createLogger(configuration);
  ...
```

There are two solutions to allow user pass a customized downlink policy.

### 1. Wrap the downlink policy as a callback, and trigger the callback after initializing the logger

Wrap the `VideoDowninkBandwidthPolicy` as a callback. Pass the downlink policy wrapper via `MeetingConfig`.

```js
// In meetingConfig.ts

...

const videoDownlinkBandwidthPolicyWrapper = (logger: Logger): VideoPriorityBasedPolicy => {
  return new VideoPriorityBasedPolicy(logger);
};

const config = {
  logLevel,
  postLogConfig,
  videoDownlinkBandwidthPolicyWrapper,
};
```
Trigger the Wrapper to initialize the downlink policy.

```js
async initializeMeetingSession(
  configuration: MeetingSessionConfiguration,
  deviceLabels: DeviceLabels | DeviceLabelTrigger = DeviceLabels.AudioAndVideo,
): Promise<any> {
  const logger = this.createLogger(configuration);

  if (this.videoDownlinkBandwidthPolicyWrapper) {
    configuration.videoDownlinkBandwidthPolicy = this.videoDownlinkBandwidthPolicyWrapper(logger);
  } 

  const deviceController = new DefaultDeviceController(logger);
  this.meetingSession = new DefaultMeetingSession(
    configuration,
    logger,
    deviceController,
    this.eventReporter
  );
```

#### pros
* No need to change the design of logger
* Only add a new property to `MeetingConfig`

#### cons
* The logic is not very straightforward, we need to wrap the downlink policy as a callback

### 2. Initialize the logger and downlink policy, pass them via MeetingConfig

```js
// In meetingConfig.ts

...

const logger = new ConsoleLogger('SDK', logLevel);
const videoDownlinkBandwidthPolicy = new VideoPriorityBasedPolicy(logger);

const config = {
  logLevel,
  postLogConfig,
  logger,
  videoDownlinkBandwidthPolicy,
};
```

```
// In the MeetingManager.ts

...

  constructor(config: ManagerConfig) {
    this.logLevel = config.logLevel;

    if (config.simulcastEnabled) {
      this.simulcastEnabled = config.simulcastEnabled;
    }

    if (config.postLogConfig) {
      this.postLoggerConfig = config.postLogConfig;
    }

    if (config.logger) {
      this.logger = config.logger;
    }
    
    if (config.videoDownlinkBandwidthPolicy) {
      this.videoDownlinkBandwidthPolicy = config.videoDownlinkBandwidthPolicy;
    }
  }

...

  async initializeMeetingSession(
    configuration: MeetingSessionConfiguration,
    deviceLabels: DeviceLabels | DeviceLabelTrigger = DeviceLabels.AudioAndVideo,
  ): Promise<any> {
    const logger = this.createLogger(configuration);

    if (this.videoDownlinkBandwidthPolicy) {
      configuration.videoDownlinkBandwidthPolicy = this.videoDownlinkBandwidthPolicy;
    }

    const deviceController = new DefaultDeviceController(logger);
    this.meetingSession = new DefaultMeetingSession(
      configuration,
      logger,
      deviceController,
      this.eventReporter
    );
```

#### pros
* Things are pretty straightforward, user can initiazlie the logger and downlink policy themselves and pass them via `MeetingConfig`

#### cons
* The newly added `logger` attribute conflicts with the existing (`logLevel` and `postLogConfig`), user should use either former or latter
* Need to add the logic to deal with this conflict

Which do you think is a better idea for this feature?

demo/meeting/src/meetingConfig.ts
xuesichao: I post the solution below. You are right, if we want to pass the custom `VideoPriorityBasedPolicy` directly, we need to initialize the logger in the `MeetingConfig.ts`. Then we need to handle the conflict of two loggers.

src/providers/MeetingProvider/MeetingManager.ts
devalevenkatesh: Yes, I checked your posted solution and looks like first way is the one we can go with as of now. Getting the logger out would be good but confusing with the existing method as you pointed out and overall my opinion is we may anyhow require a refactor on how the meeting configuration is done currently to make it more flexible and easily extensible and if there are better patterns/ways to do this which may result in a v3 😅 .

xuesichao: Based on the discussion, I will use the second solution.

src/providers/MeetingProvider/types.ts
```diff
@@ -44,4 +44,6 @@ export interface ManagerConfig {
   logLevel: LogLevel;
   postLogConfig?: PostLogConfig;
   simulcastEnabled?: boolean;
+  consoleLogger?: ConsoleLogger;
```
ltrung, line 47: This can be any logger just not console logger. Can we just use the interface `Logger` here?

xuesichao: It was `Logger` before. But Venky said I should handle the case where user pass a `postLogConfig` and `consoleLogger` at the same time. Check the code [here](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/8a19883caa8fcbf638e3e0fa7c10cbf684c099cf/src/providers/MeetingProvider/MeetingManager.ts#L242-L254)

xuesichao: @devalevenkatesh , could you please check here? Am I following your suggestion correctly?

devalevenkatesh: Hmm, so this should be `logger?: Logger`. I meant in the `createLogger` function I suggested to override the existing `ConsoleLogger` creation when `logger` is passed in the `meetingConfig`.

xuesichao: Oh, how about changing this
```js
// Current Changes
createLogger(configuration: MeetingSessionConfiguration) {
  let consoleLogger: ConsoleLogger;

  if (this.consoleLogger) {
    consoleLogger = this.consoleLogger;
  } else {
    consoleLogger = new ConsoleLogger('SDK', this.logLevel);
  }

  let logger: ConsoleLogger | MultiLogger = consoleLogger;
  this.meetingSession = new DefaultMeetingSession(
    configuration,
    logger,
    deviceController,
    this.eventReporter
  );
```
to
```js
// After
createLogger(configuration: MeetingSessionConfiguration) {
  // Change to `Logger`, as well as the one in `MeetingConfig`
  let consoleLogger: Logger;

  if (this.consoleLogger) {
    consoleLogger = this.consoleLogger;
  } else {
    consoleLogger = new ConsoleLogger('SDK', this.logLevel);
  }

  let logger: ConsoleLogger | MultiLogger = consoleLogger;
  this.meetingSession = new DefaultMeetingSession(
    configuration,
    logger,
    deviceController,
    this.eventReporter
  );
```

devalevenkatesh: So when a user passes a `logger` and `postLogConfig`, you would handle both the logger creation in the `createLogger` using `MultiLogger`. It would be something like below:
```js
createLogger(configuration: MeetingSessionConfiguration) {
    let logger: Logger;

    if (this.postLoggerConfig) {
      logger = new MultiLogger(
        this.logger ? this.logger : new ConsoleLogger('SDK', this.logLevel),
        new MeetingSessionPOSTLogger(
          this.postLoggerConfig.name,
          configuration,
          this.postLoggerConfig.batchSize,
          this.postLoggerConfig.intervalMs,
          this.postLoggerConfig.url,
          this.postLoggerConfig.logLevel
        )
      );
    }
    return logger;
  }
```

xuesichao: I remember there is a type conflict, let me double check and I will update if possible

CHANGELOG.md
```diff
@@ -17,7 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add optional parameter `deviceLabels: DeviceLabels | DeviceLabelTrigger` in `meetingManager.listAndSelectDevices()` to let builder indicate the type of devices they want to select.
-- Add several how-to docs in storybook.
+- Add optional parameter `logger` and `videoDownlinkBandwidthPolicy` in `ManagerConfig` to pass in the `ConsoleLogger` instance and `VideoDownlinkBandwidthPolicy` instance that you want to be used in the meeting session.
```
devalevenkatesh: Based on previous learning, it is recommended to say "object" over "instance". Had received one comment on an earlier PR of mine.
https://github.com/aws/amazon-chime-sdk-component-library-react/pull/554#discussion_r667229752

xuesichao: Yes sir, good point, will address all.

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
```diff
@@ -94,6 +94,13 @@ interface PostLogConfig = {
 To enable simulcast for the meeting pass the `simulcastEnabled` prop with value set to `true`. By default, it is `false`.
 For more information on Simulcast, check Amazon Chime JS SDK [simulcast guide](https://github.com/aws/amazon-chime-sdk-js/blob/master/guides/05_Simulcast.md).
 
+#### logger
+The `Logger` object that you want to be used in the meeting session.
+If you pass in a `Logger` object using this parameter, the `MeetingManager` will use this object instead of creating a logger based on `logLevel` to initialize the meeting session.
+
+#### videoDownlinkBandwidthPolicy
+The `VideoDownlinkBandwidthPolicy` object that you want to be used in the meeting session.
```
devalevenkatesh, line 102: Can add later, but since the React SDK does not have any guides on video downlink policy it would be good to link the JS SDK guides from here so the builders can get more information about it.

devalevenkatesh (dismissed):

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -218,12 +240,13 @@ export class MeetingManager implements AudioVideoObserver {
   }
 
   createLogger(configuration: MeetingSessionConfiguration) {
-    const consoleLogger = new ConsoleLogger('SDK', this.logLevel);
-    let logger: ConsoleLogger | MultiLogger = consoleLogger;
+    let logger: Logger = this.logger
```
ltrung, line 243: I think if they pass their own logger, we should not call createLogger at all

xuesichao: Yeah, I think it makes sens, cause user is able to pass a`MultiLogger`. No need to call `createLogger`.

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
ltrung: Maybe we should include a code sample on how they can pass priority downlink policy for example

devalevenkatesh: Yeah agree, a how-to suits here.

xuesichao: Good idea, will do!

xuesichao: Updated

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
```diff
@@ -91,11 +93,52 @@ interface PostLogConfig = {
...
 
+#### logger
+
+The `Logger` object that you want to be used in the meeting session.
+If you pass in a `Logger` object using this parameter, the `MeetingManager` will use this object instead of creating a logger based on `logLevel` and `postLogConfig` to initialize the meeting session.
+
+#### videoDownlinkBandwidthPolicy
+
+The `VideoDownlinkBandwidthPolicy` object that you want to be used in the meeting session.
+For more information on VideoDownlinkBandwidthPolicy, check Amazon Chime JS SDK [priority downlink policy guide](https://aws.github.io/amazon-chime-sdk-js/modules/prioritybased_downlink_policy.html).
```
devalevenkatesh, line 108: Could you please backtick VideoDownlinkBandwidthPolicy?

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
```diff
@@ -91,11 +93,52 @@ interface PostLogConfig = {
...
+
+For example, to use the `VideoPriorityBasedPolicy` in the meeting:
+
+```jsx
+import { LogLevel } from 'amazon-chime-sdk-js';
+import {
+  MeetingProvider,
+  ConsoleLogger,
+  VideoPriorityBasedPolicy,
```
devalevenkatesh, line 117: The ConsoleLogger and VideoPriorityBasedPolicy should be imported from 'amazon-chime-sdk-js' right?

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
xuesichao: Yes will do!

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
xuesichao: oh yes, that's right...

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
```diff
@@ -91,11 +93,52 @@ interface PostLogConfig = {
...
+} from 'amazon-chime-sdk-js';
+import { MeetingProvider } from 'amazon-chime-sdk-component-library-react';
+
+const Root = () => {
+  const logLevel = LogLevel.info;
+  const logger = new ConsoleLogger('SDK', logLevel);
+  const videoDownlinkBandwidthPolicy = new VideoPriorityBasedPolicy(logger);
+
+  const meetingConfig = {
+    logLevel,
```
ltrung, line 126: We should not pass logLevel in this example since it is not used.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -198,7 +216,11 @@ export class MeetingManager implements AudioVideoObserver {
     configuration: MeetingSessionConfiguration,
     deviceLabels: DeviceLabels | DeviceLabelTrigger = DeviceLabels.AudioAndVideo,
   ): Promise<any> {
-    const logger = this.createLogger(configuration);
+    if (this.videoDownlinkBandwidthPolicy) {
+      configuration.videoDownlinkBandwidthPolicy = this.videoDownlinkBandwidthPolicy;
+    }
+
+    const logger = this.logger ? this.logger : this.createLogger(configuration);
```
ltrung, line 223: Can this one be the first line as before? Not sure why we need to switch this after videoDownLinkBandwidthPolicy

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
devalevenkatesh: Actually the problem is logLevel is not optional 😅

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -120,6 +126,14 @@ export class MeetingManager implements AudioVideoObserver {
     if (config.postLogConfig) {
```
ltrung, line 126: Do we even want to set this if config.logger is set?

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
xuesichao: Yes, @devalevenkatesh is right

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: Yes, will update

xuesichao: The reason is for better readability.
```
    if (this.videoDownlinkBandwidthPolicy) {
      configuration.videoDownlinkBandwidthPolicy = this.videoDownlinkBandwidthPolicy;
    }
```
This code is not related to the logger. Inserting it behind logger breaks the logic.

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: If we want to handle this, we need extra condition here to check. Seems not very necessary to me cause it will make the code more complicated.
I'm not sure actually. In my opinion I think, If we provide this parameter, we should make sure it works. The purpose of a parameter is to pass in the external variable and set the value of internal variable. Whether to use it or not should be determined in the function, here is the `initializeMeetingSession()`.
Let me know your idea.

src/providers/MeetingProvider/MeetingManager.ts
ltrung: what logic?

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
ltrung: If i read the code right it is auto default to LogLevel.WARN in meeting provider anyway right? The required parameter is with the MeetingManager constructor.

src/providers/MeetingProvider/MeetingManager.ts
ltrung: i don't think it is more complicated. We do not use these parameters so it would be strange to set it and can cause bugs if other code read this properties and use it.

xuesichao: Yeah, make sense, I will update it!

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
xuesichao: Oh yes, you are right. Thanks for pointing out, I will update

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
```diff
@@ -91,11 +93,50 @@ interface PostLogConfig = {
...
+```jsx
+import {
+  LogLevel,
+  ConsoleLogger,
+  VideoPriorityBasedPolicy,
+} from 'amazon-chime-sdk-js';
+import { MeetingProvider } from 'amazon-chime-sdk-component-library-react';
+
+const Root = () => {
+  const logger = new ConsoleLogger('SDK', LogLevel.info);
```
ltrung, line 121: Just to confirm but have you tried this sample code out and make sure it is working?

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -110,16 +112,28 @@ export class MeetingManager implements AudioVideoObserver {
...
+
   constructor(config: ManagerConfig) {
     this.logLevel = config.logLevel;
 
     if (config.simulcastEnabled) {
       this.simulcastEnabled = config.simulcastEnabled;
     }
 
-    if (config.postLogConfig) {
+    if (!config.logger && config.postLogConfig) {
```
ltrung, line 126: Can we just do this instead?
```
if (config.logger) {
 this.logger = config.logger;
}  else {
 this.logLevel = config.logLevel;
 if (config.postLogConfig) {
   this.postLoggerConfig = config.postLogConfig;
 }
}
```

xuesichao: Yes, addressed

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
xuesichao: Yes, I packed the local sdk and installed it in the meeting demo in aws sapmle. It works. There is a typo here, the `LogLevel.info` should be `LogLevel.INFO`, addressed.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -110,15 +112,28 @@ export class MeetingManager implements AudioVideoObserver {
 
   simulcastEnabled: boolean = false;
 
+  videoDownlinkBandwidthPolicy: VideoDownlinkBandwidthPolicy | null = null;
```
devalevenkatesh, line 115: @ltrung should we start setting this to `undefined` instead of `| null`. In the recent JS SDK version we are pushing to use `undefined` instead of `null`. Want to know your opinion. Sorry for the late comment @xuesichao .

https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined

devalevenkatesh: The guidelines apply for the TypeScript contributors but checking on what we should use.

simmkyu: Yes, we should prefer `undefined` over `null`. Adding `| null` can make a variable to have three types of values unnecessarily: e.g., `VideoDownlinkBandwidthPolicy`, `undefined`, and `null`.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -110,15 +112,26 @@ export class MeetingManager implements AudioVideoObserver {
 
   simulcastEnabled: boolean = false;
 
-  constructor(config: ManagerConfig) {
-    this.logLevel = config.logLevel;
+  videoDownlinkBandwidthPolicy: VideoDownlinkBandwidthPolicy | undefined;
+
+  logger: Logger | undefined;
```
xuesichao, line 117: Change `null` to `undefined`

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -141,6 +154,8 @@ export class MeetingManager implements AudioVideoObserver {
     this.meetingStatus = MeetingStatus.Loading;
     this.publishMeetingStatus();
     this.audioVideoObservers = {};
+    this.videoDownlinkBandwidthPolicy = undefined;
+    this.logger = undefined;
```
xuesichao, line 158: Reset to undefined.

src/providers/MeetingProvider/MeetingManager.ts
ltrung: Can't we just keep it as logger: Logger? No need to do undefined here.

ltrung (approved):

src/providers/MeetingProvider/MeetingManager.ts
devalevenkatesh: Actually due to the `"strictNullChecks": true` setting in the React SDK's tsconfig, if we do this then we cannot reset it to `undefined` in initialize function. In the JS SDK we do not have this config.

ltrung: Yeah Sichao also told me that. Thus, I have approved the PR :D

devalevenkatesh (approved):

PR #586, opened 2021-08-23
Improve the way MeetingManager initializes MeetingSession

xuesichao:
**Issue #:** 

* [Bug] When `leave` method in `MeetingManager` is called, the `logger` and `videoDowlinkBandwidthPolicy` will be set to `undefined` unexpectedly
* The way `MeetingManager` stores input parameter and initializes `MeetingSession` needs to be improved

**Description of changes:**
* In `MeetingManager` add a private property `config` to store the `config` object passed in
* Remove `logLevel`, `postLoggerConfig`, `simulcastEnabled`, `videoDownlinkBandwidthPolicy` and `logger` properties
* Use the newly added property `config` instead of the removed properties to initialize the `MeetingSession`

**Testing**
1. Have you successfully run `npm run build:release` locally?
Yes

2. How did you test these changes?
* Run the meeting demo and check that the bug is fixed, the `MeetingManager` initialize `MeetingSession` correctly, and the demo works.

3. If you made changes to the component library, have you provided corresponding documentation changes?
Yes

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #587, opened 2021-08-24
Fix meeting manager init

devalevenkatesh:
**Issue #:** 

**Description of changes:**
- Add `ManagerConfig` object to `MeetingManager`.
- Move object creation/initialization from `MeetingManager` constructor to `initializeMeetingSession` method.
- Reverts earlier commit (#586) where I did not checked for backward compatibility.
- This PR addresses the required changes with backward compatability.

**Testing**
Yes.

By running, the meeting demo, tested general flow.

### DIFF
--- a/src/providers/MeetingProvider/MeetingManager.ts
+++ b/src/providers/MeetingProvider/MeetingManager.ts
@@ -6,6 +6,7 @@ import {
   ConsoleLogger,
   DefaultDeviceController,
   DefaultMeetingSession,
+  LogLevel,
   MeetingSessionConfiguration,
   DefaultActiveSpeakerPolicy,
   MeetingSessionStatus,
@@ -14,6 +15,8 @@ import {
   MultiLogger,
   MeetingSessionPOSTLogger,
   EventReporter,
+  VideoDownlinkBandwidthPolicy,
+  Logger,
   EventName,
   EventAttributes,
 } from 'amazon-chime-sdk-js';
@@ -29,6 +32,7 @@ import {
   MeetingJoinData,
   AttendeeResponse,
   FullDeviceInfoType,
+  PostLogConfig,
   ManagerConfig
 } from './types';
 
@@ -102,11 +106,38 @@ export class MeetingManager implements AudioVideoObserver {
     fullDeviceInfo: FullDeviceInfoType
   ) => void)[] = [];
 
+  logLevel: LogLevel = LogLevel.WARN;
+
+  postLoggerConfig: PostLogConfig | null = null;
+
   eventReporter: EventReporter;
+
+  simulcastEnabled: boolean = false;
+
+  videoDownlinkBandwidthPolicy: VideoDownlinkBandwidthPolicy | undefined;
+
+  logger: Logger | undefined;
 
   private meetingEventObserverSet = new Set<(name: EventName, attributes: EventAttributes) => void>();
 
-  constructor(private config: ManagerConfig) { }
+  constructor(config: ManagerConfig) {
+    if (config.simulcastEnabled) {
+      this.simulcastEnabled = config.simulcastEnabled;
+    }
+
+    if (config.logger) {
+      this.logger = config.logger;
+    } else {
+      this.logLevel = config.logLevel;
+      if (config.postLogConfig) {
+        this.postLoggerConfig = config.postLogConfig;
+      }
+    }
+
+    if (config.videoDownlinkBandwidthPolicy) {
+      this.videoDownlinkBandwidthPolicy = config.videoDownlinkBandwidthPolicy;
+    }
+  }
 
   initializeMeetingManager(): void {
     this.meetingSession = null;
@@ -127,6 +158,8 @@ export class MeetingManager implements AudioVideoObserver {
     this.meetingStatus = MeetingStatus.Loading;
     this.publishMeetingStatus();
     this.audioVideoObservers = {};
+    this.videoDownlinkBandwidthPolicy = undefined;
+    this.logger = undefined;
   }
 
   async join({ meetingInfo, attendeeInfo, deviceLabels = DeviceLabels.AudioAndVideo, eventReporter }: MeetingJoinData) {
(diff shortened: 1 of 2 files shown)


src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -185,16 +199,34 @@ export class MeetingManager implements AudioVideoObserver {
...
+      }
+    }
     if (videoDownlinkBandwidthPolicy) {
-      configuration.videoDownlinkBandwidthPolicy = videoDownlinkBandwidthPolicy;
+      this.videoDownlinkBandwidthPolicy = videoDownlinkBandwidthPolicy;
+    }
+    const logger = this.logger ? this.logger : this.createLogger(configuration);
+
+    if (this.videoDownlinkBandwidthPolicy) {
+      configuration.videoDownlinkBandwidthPolicy = this.videoDownlinkBandwidthPolicy;
```
ltrung, line 229: Could we just set this.videoDownlinkBandwidthPolicy and not use them? In v3, we probably just remove them anyway so not using them would make remove them later easier.

devalevenkatesh: Ok makes sense. Have update the code.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -224,20 +254,19 @@ export class MeetingManager implements AudioVideoObserver {
   }
 
   createLogger(configuration: MeetingSessionConfiguration) {
-    const { logLevel, postLogConfig } = this.config;
-    const consoleLogger = new ConsoleLogger('SDK', logLevel);
+    const consoleLogger = new ConsoleLogger('SDK', this.logLevel);
```
ltrung, line 257: can we also use this.meetingManagerConfig here instead of this.logLevel and this.postLoggerConfig?  Also maybe we want to add a comment to say these variables are kept for compatibility reason and will be removed in next major version update.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -185,16 +199,32 @@ export class MeetingManager implements AudioVideoObserver {
...
+    }
+    if (configLogger) {
+      this.logger = configLogger;
+    } else {
+      this.logLevel = logLevel;
+      if (postLogConfig) {
+        this.postLoggerConfig = postLogConfig;
+      }
+    }
+    const logger = this.logger ? this.logger : this.createLogger(configuration);
```
ltrung, line 223: use configLogger instead of this.logger.

devalevenkatesh: Updated, thanks!

src/providers/MeetingProvider/MeetingManager.ts
devalevenkatesh: Added comment to variables as well as in the method, this is a good call, thanks!

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -206,14 +248,14 @@ export class MeetingManager implements AudioVideoObserver {
     );
 
     this.audioVideo = this.meetingSession.audioVideo;
-
+
```
xuesichao, line 251: Empty indent here, better to delete it.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -206,14 +248,14 @@ export class MeetingManager implements AudioVideoObserver {
...
-
+    
     // When an attendee leaves, we remove AudioVideoObservers and nullify AudioVideoFacade and MeetingSession object.
     // This results into missing few meeting events triggered with audioVideoDidStop such as meetingEnded, meetingFailed and meetingStartFailed.
     // We may also loose audioInputUnselected and videoInputUnselected events as we choose null devices when an attendee leaves.
     // When a new AudioVideoFacade object is created remove and re-add the eventDidReceive observer which wont leak.
     this.audioVideo.removeObserver({ eventDidReceive: this.eventDidReceive });
     this.audioVideo.addObserver({ eventDidReceive: this.eventDidReceive });
-
+
```
xuesichao, line 258: Same here

xuesichao (commented):
Thank you guys, this PR is a very good example to learn how to deprecate variables.

src/providers/MeetingProvider/MeetingManager.ts
devalevenkatesh: Fixed, thanks for the catch!

src/providers/MeetingProvider/MeetingManager.ts
devalevenkatesh: Fixed, thanks for the catch!

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -185,16 +209,34 @@ export class MeetingManager implements AudioVideoObserver {
...
-    };
+    }
+    if (configLogger) {
+      this.logger = configLogger;
+    } else {
+      this.logLevel = logLevel;
+      if (postLogConfig) {
+        this.postLoggerConfig = postLogConfig;
+      }
+    }
```
xuesichao, line 234: Before we assign the value of `config.xxx` to `this.xxx` in the constructor, which means `this.xxx` get assigned when the App is opened (before calling `join` method). But now they are assigned after `join` is called.

Will this affect backward compatibility? Do we need to keep the original constructor code?

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -206,15 +248,13 @@ export class MeetingManager implements AudioVideoObserver {
     );
 
     this.audioVideo = this.meetingSession.audioVideo;
-
+    this.publishAudioVideo();
```
xuesichao, line 251: Why we move `publishAudioVideo()` here. Feel like it's better to publish it after remove/add observer.

devalevenkatesh: Remove and add is different from publishing an update here. The publishAudioVideo() will just send updates that we now have a `audioVideo` object. It actually should be done as soon as we get a valid object hence moved up.

src/providers/MeetingProvider/MeetingManager.ts
devalevenkatesh: Yeah that makes sense actually, builders may be using the variables before they call join(), I do not know what the use-case would be but I will keep the constructor code. Nice call!

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: Make sense, thanks for your explanation!

ltrung (dismissed):

xuesichao (approved):
Looks good to me. Nice work!

ltrung (approved):

xuesichao (approved):

PR #593, opened 2021-08-27
[Do not merge] Add a prepare script

devalevenkatesh:
**Issue #:** 
- If we want to install this GitHub package in React meeting demo directly, we need to have the `lib` folder created hence we need `npm run build` to run. Else, the demo while gives lot of type definitions and import errors.

**Description of changes:**
- Add a `prepare` script in library `package.json` to enable npm install from github URL directly.
- This helps in installing the library from GitHub directly.
- `prepare` runs before `publish` and `pack` and after `install`.
    https://docs.npmjs.com/cli/v7/using-npm/scripts
- This is actually not a good way since `npm pack` and `npm install` both will run `prepare` script.
- Once, this is merged the dependency in the demo would like:
"amazon-chime-sdk-component-library-react": "github:aws/amazon-chime-sdk-component-library-react"
- I tried `postinstall` script but that does not work as `prepare` does.

**Testing**
Yes.
Tested by installing this branch into the React meeting demo in aws-samples below:
https://github.com/aws-samples/amazon-chime-sdk/tree/main/apps/meeting
Steps:
2.1. Goto the react meeting demo in above linked repo.
2.2. In the /meeting folder do `npm install https://github.com/aws/amazon-chime-sdk-component-library-react.git/#<branch-name>`
2.3. Check that the installed `amazon-chime-sdk-component-library-react` in the node_modules in demo contains a lib folder with all exported library files.
2.4. Removed this install by doing `npm uninstall amazon-chime-sdk-component-library-react` and then re-installed from npm by doing `npm install amazon-chime-sdk-component-library-react`, re-verified whether the folder structure and lib folder is the same as with github install.


.github/workflows/publish.yml
```diff
@@ -17,8 +17,6 @@ jobs:
           registry-url: https://registry.npmjs.org/
       - name: NPM Install
         run: npm install
-      - name: NPM run build
```
devalevenkatesh, line 20: npm install runs `prepare` npm run script which is running `npm run build`.

package-lock.json
```diff
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "amazon-chime-sdk-component-library-react",
```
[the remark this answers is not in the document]
devalevenkatesh: Upgrading to node v16 solve the issue.

package.json
```diff
@@ -17,14 +17,15 @@
   },
   "types": "lib/index.d.ts",
   "scripts": {
+    "prepare": "npm run build",
     "build": "rollup -c",
     "check": "node scripts/check-codestyle.js",
     "release": "node scripts/release.js",
     "prebuild:publish": "npm install && node scripts/prebuild.js --publish",
-    "build:publish": "npm run check && npm run build && npm run test ",
+    "build:publish": "npm run check && npm run test ",
```
devalevenkatesh, line 25: `build:publish` runs `prebuild:publish` which has `npm install` which runs `prepare` internally and hence `npm run build` will get executed as is.

.github/workflows/publish.yml
michhyun1: It looks like npm `publish` also calls `prepare` script. Is there any way to not have repeat the build twice here?

What if I just want to do a npm install? I have to wait for the prepare script to finish also?

package.json
michhyun1: Do we even use this?

.github/workflows/publish.yml
devalevenkatesh: Yes multiple npm scripts actually call `prepare` script, and hence I mentioned this would not be a good way. But I could only found this way to make it work upon installation from GitHub directly currently.

devalevenkatesh:
Closing in favor of : https://github.com/aws-samples/amazon-chime-sdk/pull/52

PR #594, opened 2021-08-28
Update AudioInputVFControl component to fix Voice Focus related issue

jing-chen1:
**Issue #:** N/A

**Description of changes:**
* Set `meetingManager.selectedAudioInputDevice` as the default value of `device` in `AudioInputVFControl`. So the default device won't be `none` when user joins the meeting.
* Add `devices.length` as dependency of `useEffect()`, so the dropdown list of microphone would be updated when the devices change.

**Testing**

2. How did you test these changes? 1) User joins the meeting by entering pin and name 2) User enter the meeting, the `Default` option is selected in Mic drop down in the meeting controls instead of `None` 3) User connects audio device in the middle of the meeting 4) Audio devices are updated properly in the meeting


ltrung:
Need better PR description. QA issue does not make sense in this context and can be removed.

   - User joins the meeting with 'None' option selected as default in Mic Drop-down -> what the expected behavior ?
   - Audio device to be update under microphone option when user connects external audio in middle of the meeting -> what is the previous issue?

devalevenkatesh (approved):

Issue #605, opened 2021-09-02
The `audioVideoDidStop()` callback is not called when meeting has ended/failed


amardeep24:
### What happened and what did you expect to happen?

The `MeetingManager.audioVideo.addObserver()` method accepts an `AudioVideoObserver` implementation, which has a callback `audioVideoDidStop()`. This callback should have been called when the meeting has been ended or the meeting failed due to network disruption.

### Have you reviewed our existing documentation?

- [x] Amazon Chime SDK for JavaScript GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] Amazon Chime SDK React Components Library GitHub [issues](https://github.com/aws/amazon-chime-sdk-component-library-react/issues)
- [X] Storybook [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/introduction--page)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-component-library-react#readme)
- [X] How-to [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/how-tos--page)
- [X] React SDK meeting [demo](https://github.com/aws/amazon-chime-sdk-component-library-react/tree/master/demo/meeting#readme---react-meeting-demo)

### Reproduction steps

Step1: Join the meeting `MeetingManager.join(joinInfo)`

Step 2: Start audio video `MeetingManager.start()

Step 3: Disconnect network for > 10 mins, `audioVideoDidStop()` not called when meeting ended.

### Amazon Chime SDK React Components Library version

2.8.0

### What browsers are you seeing the problem on?

Google Chrome

### Browser version

Version 93.0.4577.63 (Official Build) (x86_64)

### Device Information

macOS Big Sur 11.4

### Meeting and Attendee ID Information.

```
{
  "meeting": {
    "meetingId": "dcd32ca2-34bf-4049-a7b5-925be0eb0706",
    "externalMeetingId": "ad927913-fedb-4ddd-8506-840f7cf28ddb",
    "mediaPlacement": {
      "audioHostUrl": "0307003aa242bb954d28cbd581b9dba3.k.m2.ue1.app.chime.aws:3478",
      "audioFallbackUrl": "wss://haxrp.m2.ue1.app.chime.aws:443/calls/dcd32ca2-34bf-4049-a7b5-925be0eb0706",
[...]

PR #608, opened 2021-09-08
Remove audio video observers in the audioVideoDidStop()

jing-chen1:
**Issue #:** [#605](https://github.com/aws/amazon-chime-sdk-component-library-react/issues/605)

**Description of changes:**
The `audioVideoDidStop()` has not been called when attendee leaves/ends the meeting. The root issue is that we remove all observers before the `audioVideoDidStop()` being called in the `audioVideo.stop()`. So remove the audio video observers in the `audioVideoDidStop()` function instead of `leave()` function.

**Testing**

    - Join the meeting
    - Leave meeting
    - `audioVideoDidStop` is called since there is a log `[MeetingManager audioVideoDidStop]` in the develop tool.
    - Join the meeting
    - End the meeting for all
    - `audioVideoDidStop` is called again.
    - Join the meeting again
    - Disconnect the network
    - `audioVideoDidStop` is called after 2 mins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+### Added
+
+### Changed
+
+- Remove the audio video observers in the `audioVideoDidStop()` function instead of `leave()` function in the `MeetingManager`.
+
+### Removed
+
 ## [2.9.1] - 2021-09-02
 
 ### Fixed
--- a/src/providers/MeetingProvider/MeetingManager.ts
+++ b/src/providers/MeetingProvider/MeetingManager.ts
@@ -229,7 +229,6 @@ export class MeetingManager implements AudioVideoObserver {
       }
 
       this.audioVideo.stop();
-      this.audioVideo.removeObserver(this.audioVideoObservers);
     }
     this.initializeMeetingManager();
     this.publishAudioVideo();
@@ -329,6 +328,11 @@ export class MeetingManager implements AudioVideoObserver {
     } else {
       console.log(`[MeetingManager audioVideoDidStop] session stopped with code ${sessionStatusCode}`);
     }
+    
+    if (this.audioVideo) {
+      this.audioVideo.removeObserver(this.audioVideoObservers);
+    }
+    
     this.leave();
   };
 


src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -273,12 +272,13 @@ export class MeetingManager implements AudioVideoObserver {
...
-    // When an attendee leaves, we remove AudioVideoObservers and nullify AudioVideoFacade and MeetingSession object.
-    // This results into missing few meeting events triggered with audioVideoDidStop such as meetingEnded, meetingFailed and meetingStartFailed.
+    // When an attendee leaves, we nullify AudioVideoFacade and MeetingSession object.
+    // We don't remove AudioVideoObservers in the leave function because AudioVideoObservers will be removed before being called.
     // We may also loose audioInputUnselected and videoInputUnselected events as we choose null devices when an attendee leaves.
-    // When a new AudioVideoFacade object is created remove and re-add the eventDidReceive observer which wont leak.
+    // When a new AudioVideoFacade object is created, remove and re-add the eventDidReceive observer which wont leak.
     this.audioVideo.removeObserver({ eventDidReceive: this.eventDidReceive });
     this.audioVideo.addObserver({ eventDidReceive: this.eventDidReceive });
+    this.audioVideo.removeObserver(this.audioVideoObservers);
```
michhyun1, line 281: When do these observers get removed when we normally call meetingManager.leave()? Won't they still be active if you join a meeting and then leave a meeting? We'd never removeObservers because we never call initializeMeetingSession when we leave a meeting, right? Only when we join a meeting.

jing-chen1: Make sense, remove audio video observers in the `audioVideoDidStop()` function instead

devalevenkatesh (commented):
Looks like the changelog check is broken. We need a CHANGELOG entry for this. I also suspect say for some reason `audioVideoDidStop` does not get called then we would still need a way to clear the observer which would be when we set it up [here](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/f74f0e14948c486953daa2da77478a490d370a16/src/providers/MeetingProvider/MeetingManager.ts#L345). But, I would want to get in the failure state first and then add that later, hence, for now this looks good to me.

CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ All notable changes to this project will be documented in this file.
...
 
+## [Unreleased]
+
+### Fixed
+
+### Added
+
+### Changed
+
+- Remove the audio video observers in the `audioVideoDidStop()` function instead of `leave()` function.
```
devalevenkatesh, line 18: Can you extend this sentence to end with "in MeetingManager"? We need builders to understand the scope of these functions hence asking.

jing-chen1: Yep, sure

devalevenkatesh (approved):

devalevenkatesh (changes requested):
I think you rebased with one extra commit which exports useVoiceFocus hook.

devalevenkatesh (approved):

PR #618, opened 2021-09-16
Add LoggerProvider and useLogger components

devalevenkatesh:
**Issue #:** 

**Description of changes:**
- Add `LoggerProvider` and `useLogger` components to enable logging from
component library.
- Update existing console log to use logger from `useLogger` hook.
- Add relevant documentation.

**Testing**

2. How did you test these changes? Tested by using the new `LoggerProvider` in the react meeting demo app.
2.1 Wrapped the existing tree in [here](https://github.com/aws-samples/amazon-chime-sdk/blob/main/apps/meeting/src/app.tsx#L26-L55) like below and checked logs appeared as expected.
```js
import {
  ...,
LoggerProvider
} from 'amazon-chime-sdk-component-library-react';

const App: FC = () => (
  <Router>
    <LoggerProvider>
      <AppStateProvider>
        ...
            <VoiceFocusProvider>
              <MeetingProvider {...meetingConfig}>
                <NavigationProvider>
                  <Switch>
                    <Route exact path={routes.HOME} component={Home} />
                    <Route path={routes.DEVICE}>
                      <NoMeetingRedirect>
                        <DeviceSetup />
                      </NoMeetingRedirect>
                    </Route>
                    <Route path={routes.MEETING}>
                      <NoMeetingRedirect>
                        <MeetingModeSelector />
                      </NoMeetingRedirect>
                    </Route>
                  </Switch>
                </NavigationProvider>
              </MeetingProvider>
            </VoiceFocusProvider>
          ...
      </AppStateProvider>
    </LoggerProvider>
  </Router>
);
```

3. If you made changes to the component library, have you provided corresponding documentation changes? Yes.


src/providers/LoggerProvider/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+interface Props {
+  /**
+   * Represents [LogLevel](https://aws.github.io/amazon-chime-sdk-js/enums/loglevel.html) to determine how verbose the logging should be.
+   */
+  logLevel?: LogLevel;
+  
+  /**
+   * This name will be used as a prefix when we log statements.
+   */
+  name?: string;
```
ltrung, line 19: I thought we would do multiLogger. I think that is important since that is how customer store logs and send to us right?

src/providers/LoggerProvider/docs/LoggerProvider.stories.mdx
```diff
@@ -0,0 +1,105 @@
+import { Props } from '@storybook/addon-docs/blocks';
+import { LoggerProvider } from '../';
+
+<Meta title="SDK Providers/LoggerProvider" />
+
+# LoggerProvider
+
+The `LoggerProvider` provides a `Logger` to log component state changes and lifecycle events such as mounting and unmounting.
```
ltrung, line 8: Should we just have one guide for both LoggerProvider and useLogger since they are supposed to be used together right?

devalevenkatesh: Yeah, we have generally followed a pattern and have documentation for both. But it looks like most of the documentation repeats and is not required. I have deleted the `useLogger` story and have updated the `LoggerProvider` story.

src/providers/LoggerProvider/index.tsx
devalevenkatesh: So we have two routes:
1. Get POSTLogger config but for that builders must pass `MeetingSessionConfiguration` as well and then only we can create a `MultiLogger`.
2. Provide a `Logger` prop and let builders pass any class implementing `Logger` interface, but if not provided we fall back to `ConsoleLogger` which itself can be customize with a `name` and `LogLevel`.

I have went with option 2 as anyhow I think we would have to provide the `Logger` prop to let builders pass their own prop. I have provided a usage example as well on how to provide a custom logger.

ltrung: So in the case of multiLogger where they post logs to CloudWatch, how would they have both React and meeting post to the same log group then? Will they have to create one Logger and pass to post the `MeetingProvider` and `LoggerProvider`?
You will need to do that with the react meeting demo right?

devalevenkatesh: Yes, they have to create and pass to both the MeetingProvider and LoggerProvider. We want to make `LoggerProvider `completely independent from any meeting specific configuration hence we discussed to only support `ConsoleLogger` initially as constructing the POSTLogger needs the meeting session configuration.

devalevenkatesh: The POST logging IMO can be completely dropped until we support creating the `MeetingSessionConfiguration` before calling the `meetingManager.join()` then only we could log to same group. Also, the POST logger taking the meetingSessionConfiguration will still have the same attendeeId and meetingId as it would be created from the same CreateAttendee and CreateMeeting responses but it is not a good way I think.

devalevenkatesh:
Closing this PR since currently we could only add console logging even with using Logger provider if we are use JS SDK's POSTLogging which is actually beneficial than adding the logger with just simple ConsoleLogger.

PR #620, opened 2021-09-21
Fix the issue that Amazon Voice Focus does not get applied on new devices in the middle of meeting

jing-chen1:

**Description of changes:**
1. This issue only happens in Chrome, since only Chrome have the `default` device. If user selects a specific device or uses other browser such as Safari/FireFox, the current device won't be changed to the new device automatically which is expected.
2. The issue only happens when the new device is AirPods, since only AirPods have two steps to be the current device. When you open the container, AirPods will be shown in the device dropdown list; When you put AirPods on your ears, the current device will be set to AirPods automatically. However, the VF dropdown list only gets updated at the first step, since the length of available devices changes, and it won't be updated at the second step, since no dependencies change in this [useEffect](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/components/sdk/MeetingControls/AudioInputVFControl.tsx#L119). The `selectedDevice` is still `default`.    

Fix: `devices[0]` stands for the default device in Chrome, so we can add `devices[0].label` as a dependency, since although the `selectedDevice`(`devices[0].deviceId`) keeps the same(`default`), but `devices[0].label` will change when a new device is set to the default device.

**Testing**

2. How did you test these changes? 1.Join the meeting in Chrome 2.Keep the default device as audio input 3.Enable Amazon Voice Focus 4.Apply a new audio device (AirPods) in the middle of meeting 5.Current audio device switches to AirPods voice focus transform device.

### DIFF
--- a/src/components/sdk/MeetingControls/AudioInputVFControl.tsx
+++ b/src/components/sdk/MeetingControls/AudioInputVFControl.tsx
@@ -119,6 +119,7 @@ const AudioInputVFControl: React.FC<Props> = ({
     addVoiceFocus,
     device,
     devices.length,
+    devices[0].label,
     isLoading,
     isVoiceFocusEnabled,
     isVoiceFocusOn,
--- a/src/providers/DevicesProvider/AudioInputProvider.tsx
+++ b/src/providers/DevicesProvider/AudioInputProvider.tsx
@@ -68,6 +68,9 @@ const AudioInputProvider: React.FC = ({ children }) => {
             'Previously selected audio input lost. Selecting a default device.'
           );
           meetingManager.selectAudioInputDevice(newAudioInputs[0].deviceId);
+
+          // Safari and Firefox don't have this "default" as device Id
+          // Only Chrome have this "default" device
         } else if (selectedInputRef.current === 'default') {
           console.log(
             `Audio devices updated and "default" device is selected. Reselecting input.`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- Fix the issue that Amazon Voice Focus does not get applied on new devices in the middle of meeting
 
 ### Added
 - Add `activeSpeakerPolicy` and `videoUplinkBandwidthPolicy` in `MeetingManagerConfig` to allow builders to pass in 


src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -119,6 +119,7 @@ const AudioInputVFControl: React.FC<Props> = ({
     addVoiceFocus,
     device,
     devices.length,
+    devices[0].label,
```
michhyun1, line 122: hmm how does this change actually make it so that VF is enabled? Won't this change just render the dropdown again with updates?

ltrung: isn't label in this case still `default`?

michhyun1: No, device id is `default` - the label would change from `Macbook Microphone` to `Airpods`

jing-chen1: The `devices[0].deviceId` is still `default`, but `devices[0].label` is changed from `Internal Microphone` to `AirPods`

jing-chen1: > hmm how does this change actually make it so that VF is enabled? Won't this change just render the dropdown again with updates?

Yep, exactly. The current device is actually set to the new device, but the change does not reflect on the dropdown list. We only need to update the dropdown list

ltrung: We should handle this using the audioInputsChanged event from JS SDK. This looks a bit hacky.

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -68,6 +68,9 @@ const AudioInputProvider: React.FC = ({ children }) => {
             'Previously selected audio input lost. Selecting a default device.'
           );
           meetingManager.selectAudioInputDevice(newAudioInputs[0].deviceId);
+
+          // Safari and Firefox don't have this "default" as device Id
+          // Only Chrome have this "defaul" device
```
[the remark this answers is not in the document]
jing-chen1: Thanks, updated

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
michhyun1: Are you sure VF would be enabled still? In AudioInputProvider,  we call 
`await audioVideo?.chooseAudioInputDevice(selectedInputRef.current);` - This is called directly (not via meetingManager) - how would the selectedAudioInputTransformDevice get updated?

jing-chen1: >We should handle this using the audioInputsChanged event from JS SDK. This looks a bit hacky.  

Actually `audioInputsChanged` has been trigged twice as there are 2 steps for AirPods to be set as default device and in the second time, only `audioInputs` changes. This dropdown list is only updated when length of `audioInputs` or `audioInputs[0].deviceId` changes and will stay the same when `audioInputs[0].label` changes. 

Since we can not put `audioInputs` as a dependency, so use `audioInputs[0].label` instead.

jing-chen1: > Are you sure VF would be enabled still? In AudioInputProvider, we call
> `await audioVideo?.chooseAudioInputDevice(selectedInputRef.current);` - This is called directly (not via meetingManager) - how would the selectedAudioInputTransformDevice get updated?

Without VF, when this `audioVideo?.chooseAudioInputDevice` is called, the `selectedAudioInputTransformDevice` gets updated

ltrung: Shouldn't we just remove devices.length and devices[0].label and just use devices here?

jing-chen1: I'd love to...but if we use `devices` here would cause `Maximum update depth exceeded` issue

jing-chen1: > Are you sure VF would be enabled still? In AudioInputProvider, we call
> `await audioVideo?.chooseAudioInputDevice(selectedInputRef.current);` - This is called directly (not via meetingManager) - how would the selectedAudioInputTransformDevice get updated?

Just confirmed, our `selectedAudioInputTransformDevice` won't change, it would keep as string `default`. But the `default` device changes in `audioInputs`, no matter whether VF is enabled or not.

michhyun1: So the second time the audioInputsChanged is called, the device list is technically the same but the 'default' device has a different label, which is why we have to change based on that label. Makes sense

michhyun1 (approved):

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
jing-chen1: Yep, exactly

PR #621, opened 2021-09-22
Add Amazon Voice Focus to default device

jing-chen1:

**Description of changes:**
Add Amazon Voice Focus to the default device when a new blue tooth device is plugged in during the meeting. 

**Testing**

2. How did you test these changes? Open two Chrome tabs, select the default device as audio input and disable speaker in one tab, disable microphone and set built in device as speaker in another tab. Turn on Amazon Voice Focus in the first tab, and apply a new device (AirPods), the current device changes as AirPods. Then make some background noice, we don't hear any noice from the second tab, it shows Amazon Voice Focus is enabled in the first tab.

### DIFF
--- a/src/providers/DevicesProvider/AudioInputProvider.tsx
+++ b/src/providers/DevicesProvider/AudioInputProvider.tsx
@@ -9,13 +9,20 @@ import React, {
   useMemo,
   useRef,
 } from 'react';
-import { DeviceChangeObserver } from 'amazon-chime-sdk-js';
+
+import { 
+  DeviceChangeObserver, 
+  isAudioTransformDevice, 
+  Device, 
+  VoiceFocusTransformDevice,
+} from 'amazon-chime-sdk-js';
 
 import { useAudioVideo } from '../AudioVideoProvider';
 import { useMeetingManager } from '../MeetingProvider';
 import { getFormattedDropdownDeviceOptions } from '../../utils/device-utils';
 import { DeviceTypeContext, DeviceConfig } from '../../types';
 import { AUDIO_INPUT } from '../../constants/additional-audio-video-devices';
+import { useVoiceFocus } from '../VoiceFocusProvider';
 
 const Context = createContext<DeviceTypeContext | null>(null);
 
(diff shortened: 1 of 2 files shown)


michhyun1 (approved):

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -76,7 +84,11 @@ const AudioInputProvider: React.FC = ({ children }) => {
             `Audio devices updated and "default" device is selected. Reselecting input.`
           );
           try {
-            await audioVideo?.chooseAudioInputDevice(selectedInputRef.current);
+            let currentDevice: Device | VoiceFocusTransformDevice   = 'default'
```
ltrung, line 87: Have we tested out in JS SDK meeting demo to see whether it has the same issue? I don't see any VF logic handling there so I assume so.

jing-chen1: Just checked in JS SDK meeting demo, we don't have this kind issue. And yes, VF does not handle this issue.

michhyun1: Wait, so does JS SDK have this issue or not?

michhyun1 (approved):

src/providers/DevicesProvider/AudioInputProvider.tsx
jing-chen1: > Wait, so does JS SDK have this issue or not?

Yep, JS SDK meeting demo does not have this issue. If the device is a default VF transform device, the `currentAudioInputDevice` in this [line](https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/browser/app/meetingV2/meetingV2.ts#L1438) won't be a string `default`, instead, it will be a VF transformed device. However, in React library, we use `meetingManager.selectedAudioInputDevice: string | null` to define the current audio input, when the current device is a default VF transformed device, it is still string `default`. I think the root issue is that, in JS SDK, we receive a real device as selected device, while in React library, we only accept `String`.

PR #623, opened 2021-09-22
Fix bug regarding setting meetingStatus to Loading when leaving meeting

michhyun1:
**Issue #:** 

**Description of changes:**
We don't want to set meetingStatus to Loading - this is the same state we use when we are joining a meeting. It can be confusing for builders if they are listening to this Loading state. 

**Testing**
yes

Start a meeting, leave a meeting.

N/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,16 +11,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fix the issue that Amazon Voice Focus does not get applied on new devices mid-meeting
+- Fix the issue where we call `meetingManager.leave` an additional time when we call `meetingManager.leave`
 
 ### Added
 - Add `activeSpeakerPolicy` and `videoUplinkBandwidthPolicy` in `MeetingManagerConfig` to allow builders to pass in 
   custom policies.
 - For more flexibility, allow passing `MeetingManagerConfig` to `meetingManager.join` method. Passing the config here would override config passed through `MeetingProvider` props.
+- Add `MeetingStatus.Left` and set it when explicitly leaving the meeting
+- Publish `MeetingStatus.Failed` when `audioVideoDidStop` gets triggered with one of the Failure types of `MeetingSessionStatus` 
+
 
 ### Changed
 - Remove the audio video observers in the `audioVideoDidStop()` function instead of `leave()` function in the `MeetingManager`.
 
 ### Removed
+- Remove setting the `MeetingStatus` to `MeetingStatus.Loading` when we call `meetingManager.leave`
 
 ## [2.9.1] - 2021-09-02
 
--- a/src/providers/MeetingProvider/MeetingManager.ts
+++ b/src/providers/MeetingProvider/MeetingManager.ts
@@ -183,8 +183,6 @@ export class MeetingManager implements AudioVideoObserver {
     this.videoInputDevices = [];
     this.activeSpeakers = [];
     this.activeSpeakerListener = null;
-    this.meetingStatus = MeetingStatus.Loading;
-    this.publishMeetingStatus();
     this.audioVideoObservers = {};
   }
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fix the issue that Amazon Voice Focus does not get applied on new devices mid-meeting
+- Fix the issue where we set the MeetingStatus to Loading when we call meetingManager.leave()
```
devalevenkatesh, line 14: nit: Backtick MeetingStatus, Loading and meetingManager.leave().

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -166,7 +166,7 @@ export class MeetingManager implements AudioVideoObserver {
     }
   }
 
-  initializeMeetingManager(): void {
```
devalevenkatesh, line 169: We do want to change the name of this method as this is a public method hence lets not change it for now, but only change the meeting status.

src/types/index.ts
```diff
@@ -44,6 +44,7 @@ export enum MeetingStatus {
   Succeeded,
   Failed,
   Ended,
+  Left,
```
devalevenkatesh, line 47: We also need to update the `useMeetingStatus` storybook hook documentation. It has the enum there which needs to be updated.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -329,19 +327,24 @@ export class MeetingManager implements AudioVideoObserver {
...
+      this.meetingStatus = MeetingStatus.Left;
+      this.publishMeetingStatus();
     } else if (sessionStatusCode === MeetingSessionStatusCode.AudioJoinedFromAnotherDevice) {
       console.log('[MeetingManager audioVideoDidStop] Meeting joined from another device');
       this.meetingStatus = MeetingStatus.JoinedFromAnotherDevice;
       this.publishMeetingStatus();
+      this.leave();
     } else {
       console.log(`[MeetingManager audioVideoDidStop] session stopped with code ${sessionStatusCode}`);
+      this.leave();
```
devalevenkatesh, line 342: We should publish some status code here like `MeetingSession.Failed` if the it is a failure. We can check if failure using `sessionStatus.isFailure()` or  `sessionStatus.isAudioConnectionFailure()`. Else, add a MeetingStatus as Unknow if not related to this. This is quick fix so up to you if you want to add this now. I think at least the Failed should be published here.

src/providers/MeetingProvider/MeetingManager.ts
michhyun1: oh yeah youre right

src/providers/MeetingProvider/MeetingManager.ts
ltrung: Yeah but if Michael is pressed for time then we could come back to this later :). It is a bit out of scope for this PR :)

src/providers/MeetingProvider/MeetingManager.ts
ltrung: Hmm I don';t think this is we discussed. I think we propose to create a new (please private) function reset/tearDownMeetingManager which does the samething as `intializeMeetingManager` except publish status.
`initializeMeetingManager` can just call `resetMeetingManager` and publish status.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -329,19 +327,24 @@ export class MeetingManager implements AudioVideoObserver {
       console.log('[MeetingManager audioVideoDidStop] Meeting ended for all');
       this.meetingStatus = MeetingStatus.Ended;
       this.publishMeetingStatus();
+      this.leave();
```
ltrung, line 330: Can you put a comment here explain why we do not call leave here?
Also maybe a switch statement is better here :D

michhyun1: we do call leave() for this case - and yes, Im swithcing to use a swtich case

ltrung (approved):

ltrung:
Seem like build:release fail :D

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -325,23 +323,54 @@ export class MeetingManager implements AudioVideoObserver {
...
+        // No need to call leave() here, since we already called meetingManager.leave() to get here
+        break;
+      case MeetingSessionStatusCode.AudioJoinedFromAnotherDevice:
+        console.log('[MeetingManager audioVideoDidStop] Meeting joined from another device');
+        this.meetingStatus = MeetingStatus.JoinedFromAnotherDevice;
+        this.publishMeetingStatus();
+        this.leave();
+        break;
+      // The following status codes are Failures according to MeetingSessionStatus
+      case MeetingSessionStatusCode.AudioAuthenticationRejected:
```
devalevenkatesh, line 346: I think you can simply check this using `sessionStatus.isFailure()` right?

devalevenkatesh (commented):
I think no need of package-lock.json with this PR so can we remove it?

src/providers/MeetingProvider/MeetingManager.ts
michhyun1: Yup I guess thats better

michhyun1:
> I think no need of package-lock.json with this PR so can we remove it?

I had node 14 xD it was causing it to break

devalevenkatesh (approved):

Issue #624, opened 2021-09-23
Voice Focus selectedDevice is not updating when switching to new audio device


jing-chen1:
### What happened and what did you expect to happen?

If user chooses the default device as the current audio input device, and applies a new device during the meeting, `selectedDevice` will not be updated since `selectedDevice` stands for the deviceId, and the value of deviceId remains the same, still being `default`. 

The root issue for this issue is that we only accept string as device in React library. We will fix this issue in the next major version.

Mitigation: If you want to get the update when switching to a new audio device, it would be better to use the `devices`. When a new 'default' device is added, the `devices` list will have changed. We suggest you to watch for changes to `devices[0]` - this will indicate when the default device has changed. 

**Note**
This issue only happens in Chrome, since Safari and FireFox don't have any "default" device.

### Have you reviewed our existing documentation?

- [X] Amazon Chime SDK for JavaScript GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] Amazon Chime SDK React Components Library GitHub [issues](https://github.com/aws/amazon-chime-sdk-component-library-react/issues)
- [X] Storybook [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/introduction--page)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-component-library-react#readme)
- [X] How-to [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/how-tos--page)
- [X] React SDK meeting [demo](https://github.com/aws/amazon-chime-sdk-component-library-react/tree/master/demo/meeting#readme---react-meeting-demo)

### Reproduction steps

1. Join the meeting
2. Choose default as audio input device
3. Turn on Amazon Voice Focus
4. Add a new device during the meeting
5. `selectedDevice` is not updating

### Amazon Chime SDK React Components Library version

2.9.1

### What browsers are you seeing the problem on?

Chrome

### Browser version

Version 92.0.4515.159

### Device Information

MacBook Pro 2017

### Browser console logs

Audio devices updated and "default" device is selected. Reselecting input.

jing-chen1:
We have fix this issue in [Add Amazon Voice Focus to default device in VF provider](https://github.com/aws/amazon-chime-sdk-component-library-react/pull/639).

PR #625, opened 2021-09-23
Add more documentation for AudioInputProvider

jing-chen1:
**Issue #:** [#624](https://github.com/aws/amazon-chime-sdk-component-library-react/issues/624)

**Description of changes:**
Add some documentation to mitigate this issue.

**Testing**

2. How did you test these changes? N/A

### DIFF
--- a/src/providers/DevicesProvider/docs/AudioInputProvider.stories.mdx
+++ b/src/providers/DevicesProvider/docs/AudioInputProvider.stories.mdx
@@ -4,6 +4,11 @@
 
 The `AudioInputProvider` provides a list of the user's available `audio-input` devices, the currently selected device, and the error if device selection fails.
 
+NOTE: If user chooses the default device as the current audio input device in Chrome, and plugs in a new device during the meeting that becomes the new default device, 
+`selectedDevice` will not be updated since `selectedDevice` stands for the deviceId, and the value of deviceId remains the same, still being `default`. If you want to get 
+the update when switching to a new audio device, it would be better to use `devices`. When a new 'default' device is added, the `devices` list will have changed. 
+We suggest you to watch for changes to `devices[0]` - this will indicate when the default device has changed. 
+
 ## State
 
 ```javascript
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `activeSpeakerPolicy` and `videoUplinkBandwidthPolicy` in `MeetingManagerConfig` to allow builders to pass in 
   custom policies.
 - For more flexibility, allow passing `MeetingManagerConfig` to `meetingManager.join` method. Passing the config here would override config passed through `MeetingProvider` props.
+- Add more details in the `AudioInputProvider` on storybook.
 
 ### Changed
 - Remove the audio video observers in the `audioVideoDidStop()` function instead of `leave()` function in the `MeetingManager`.


src/providers/DevicesProvider/docs/AudioInputProvider.stories.mdx
```diff
@@ -4,6 +4,11 @@
 
 The `AudioInputProvider` provides a list of the user's available `audio-input` devices, the currently selected device, and the error if device selection fails.
 
+NOTE: If user chooses the default device as the current audio input device in Chrome, and applies a new device during the meeting, `selectedDevice` will not be updated
```
ltrung, line 7: > , and applies a new device during the meeting,
how about 'and plug in a new device during the meeting that becomes the new default device?

src/providers/DevicesProvider/docs/AudioInputProvider.stories.mdx
```diff
@@ -4,6 +4,11 @@
 
 The `AudioInputProvider` provides a list of the user's available `audio-input` devices, the currently selected device, and the error if device selection fails.
 
+NOTE: If user chooses the default device as the current audio input device in Chrome, and applies a new device during the meeting, `selectedDevice` will not be updated 
+since `selectedDevice` stands for the deviceId, and the value of deviceId remains the same, still being `default`. If you want to get the update when switching 
+to a new audio device, it would be better to use the `devices`. When a new 'default' device is added, the `devices` list will have changed. We suggest you to watch
```
ltrung, line 9: to use `devices`.

src/providers/DevicesProvider/docs/AudioInputProvider.stories.mdx
jing-chen1: yep, that's better. will update it

src/providers/DevicesProvider/docs/AudioInputProvider.stories.mdx
```diff
@@ -4,6 +4,11 @@
 
 The `AudioInputProvider` provides a list of the user's available `audio-input` devices, the currently selected device, and the error if device selection fails.
 
+NOTE: If user chooses the default device as the current audio input device in Chrome, and plugs in a new device during the meeting that becomes the new default device,
```
devalevenkatesh, line 7: Builders might know what is 'default' device. So should we add some details about it?

devalevenkatesh, line 7: We do not mention devices until we provide the state information in the doc, so should we move this after we explain or show the devices?

jing-chen1: Added some expected behavior to the "default" device

jing-chen1: Moved it below the `state` section

src/providers/DevicesProvider/docs/AudioInputProvider.stories.mdx
```diff
@@ -22,6 +22,12 @@ The `AudioInputProvider` provides a list of the user's available `audio-input` d
 
 ```
 
+NOTE: In Chrome, there is a "default" device. If a user chooses the "default" device as the current audio input device, and plugs in a new device during the
```
harshitOberoi, line 25: How about putting the suggestion first and give reasons for using that. Something like below:

NOTE: In Chrome , there is a "default" device which can used as the current audio input device but it would be better to use `devices`. When a new "default" device is added, the `devices` list will have changed. We suggest you to watch for changes to `devices[0] this will indicate when the "default" device has changed.

harshitOberoi: and then give the disadvantage of using default.

src/providers/DevicesProvider/docs/AudioInputProvider.stories.mdx
```diff
@@ -22,6 +22,12 @@ The `AudioInputProvider` provides a list of the user's available `audio-input` d
 
 ```
 
+NOTE: In Chrome, there is a "default" device. If a user chooses the "default" device as the current audio input device, and plugs in a new device during the 
+meeting, the new device will become the current audio input device also the new "default" device. But `selectedDevice` will not be updated since `selectedDevice` 
+stands for the deviceId, and the value of deviceId remains the same, still being `default`. If you want to get the update when switching to a new audio device,
```
harshitOberoi, line 27: "and the value of deviceId remains the same, still being `default`"  I think we should remove ",still being `default`" as it is a little confusing as if new device is updated as "default" why "selectedDevice" still being default is not updated.

src/providers/DevicesProvider/docs/AudioInputProvider.stories.mdx
jing-chen1: Sure, I can first give the mitigation method and then give the reason why `selectedDevice` is not updated when switching to a new device. But it is not the disadvantage of using default device.

src/providers/DevicesProvider/docs/AudioInputProvider.stories.mdx
jing-chen1: Maybe I need to add more details about it so builders can understand why `selectedDevice` is still being default.

src/providers/DevicesProvider/docs/AudioInputProvider.stories.mdx
```diff
@@ -22,6 +22,14 @@ The `AudioInputProvider` provides a list of the user's available `audio-input` d
 
 ```
 
+NOTE: In Chrome, there is a "default" device which can used as the current audio input device. When a new device is added during the 
+meeting, the new device will become the current audio input device also the new "default" device. If you want to get this device change, 
+it would be better to use `devices`. When a new "default" device is added, the `devices` list will have changed. We suggest you to 
+watch for changes to `devices[0]` - this will indicate when the "default" device has changed. `selectedDevice` will not be updated when 
+switching to a new audio device since it stands for the deviceId, and the value of deviceId remains the same, still being `default`. 
+The deviceId that was previously in use was `default`. When you switch to a new device, that device becomes "default" device. Since the old 
+deviceId was `default` and the new deviceId is STILL `default`, So `selectedDevice` does not change. But actually they correspond to different physical devices.
```
harshitOberoi, line 31: Change "STILL" to lowercase

harshitOberoi (approved):

PR #583, opened 2021-08-10
add voice focus feature

jing-chen1:
**Issue #:** N/A

**Description of changes:**
1. Extend the ability to accept device and AudioTransformDevice as audio input by adding a new props `selectedAudioInputTransformDevice` 
2. Create a new provider which is separate from meeting provider. This provider can initial the voice focus and transform a device to voice focus transform device
3. Create a new UI component `AudioInputVFControl` providers voice focus option for users.

Todo:
1. Test on meeting demo
2. Add doc

**Testing**
1. Have you successfully run `npm run build:release` locally?

2. How did you test these changes?

3. If you made changes to the component library, have you provided corresponding documentation changes?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #630, opened 2021-09-28
Add logs for Voice Focus components

jing-chen1:
**Issue #:** N/A

**Description of changes:**
Add more logs for Voice Focus components. So builders can know whether Amazon Voice Focus is supported or not, also the VF transform device if VF is enabled.

**Testing**

2. How did you test these changes? Test on meeting demo to see how the logs behave. They behave as expected.

### DIFF
--- a/src/components/sdk/MeetingControls/AudioInputVFControl.tsx
+++ b/src/components/sdk/MeetingControls/AudioInputVFControl.tsx
@@ -42,7 +42,7 @@ const AudioInputVFControl: React.FC<Props> = ({
   const meetingManager = useMeetingManager();
   const [isLoading, setIsLoading] = useState(false);
   // When the user click on Amazon Voice Focus option, the state will change.
-  const [isVoiceFocusCheckboxOn, setIsVoiceFocusCheckboxOn] = useState<boolean | undefined>(undefined);
+  const [isVoiceFocusChecked, setIsVoiceFocusChecked] = useState(false);
   // Only when the current input audio device is an Amazon Voice Focus device, the state will be true. Otherwise, it will be false.
   const [isVoiceFocusEnabled, setIsVoiceFocusEnabled] = useState(false);
   const [dropdownWithVFOptions, setDropdownWithVFOptions] = useState<ReactNode[] | null>(null);
@@ -82,7 +82,7 @@ const AudioInputVFControl: React.FC<Props> = ({
         children={<span>{device.label}</span>}
         checked={isOptionActive(selectedDevice, device.deviceId)}
         onClick={async (): Promise<void> => {
-          if (isVoiceFocusCheckboxOn && !isLoading) {
+          if (isVoiceFocusChecked && !isLoading) {
             setIsLoading(true);
             const receivedDevice = audioInputSelectionToDevice(device.deviceId);
             const currentDevice = await addVoiceFocus(receivedDevice);
@@ -108,7 +108,7 @@ const AudioInputVFControl: React.FC<Props> = ({
           disabled={isLoading}
           onClick={() => {
             setIsLoading(true);
-            setIsVoiceFocusCheckboxOn(current => !current);
+            setIsVoiceFocusChecked(current => !current);
           }} />
       )
       dropdownOptions?.push(<PopOverSeparator key = 'separator' />);
@@ -126,26 +126,22 @@ const AudioInputVFControl: React.FC<Props> = ({
     devices[0].label,
     isLoading,
     isVoiceFocusEnabled,
-    isVoiceFocusCheckboxOn,
+    isVoiceFocusChecked,
     isVoiceFocusSupported,
     selectedDevice,
   ]);
 
   useEffect(() => {
     async function onVFCheckboxChange() {
-      if (isVoiceFocusCheckboxOn === undefined) {
-        return;
-      }
-
       let current = device;
-      if (isVoiceFocusCheckboxOn) {
+      if (isVoiceFocusChecked) {
         console.info('User turned on Amazon Voice Focus.')
         if (typeof (device) === 'string') {
           const currentDevice = audioInputSelectionToDevice(device);
           current = await addVoiceFocus(currentDevice);
         }
       } else {
-        console.info('User turned off Amazon Voice Focus.')
+        console.info('Amazon Voice Focus is off by default or user turned off Amazon Voice Focus.')
         if (device instanceof VoiceFocusTransformDevice) {
           current = device.getInnerDevice();
         }
@@ -155,7 +151,7 @@ const AudioInputVFControl: React.FC<Props> = ({
     }
 
     onVFCheckboxChange();
-  }, [isVoiceFocusCheckboxOn]);
+  }, [isVoiceFocusChecked]);
 
   return (
     <ControlBarButton
--- a/src/providers/VoiceFocusProvider/docs/VoiceFocusProvider.stories.mdx
+++ b/src/providers/VoiceFocusProvider/docs/VoiceFocusProvider.stories.mdx
@@ -1,8 +1,12 @@
+import { Props } from '@storybook/addon-docs/blocks';
+import { VoiceFocusProvider } from '../';
+
 <Meta title="SDK Providers/VoiceFocusProvider" />
 
 # VoiceFocusProvider
 
-The `VoiceFocusProvider` provides a function transforming a normal device to an Amazon Voice Focus device, and whether Amazon Voice Focus is supported.
+The `VoiceFocusProvider` provides a function transforming a normal audio device to an Amazon Voice Focus device, and whether Amazon Voice Focus is supported. 
+Amazon Voice Focus related logs can be found in the browser developer tools when Amazon Voice Focus is enabled.
 
 This provider is independent from `MeetingProvider`. You can put `VoiceFocusProvider` wherever you want in the component tree to control the workflow of Voice Focus.
 
@@ -25,6 +29,10 @@ Amazon Voice Focus is supported on your device and browser. For more information
 
 You can access the state by using the [useVoiceFocus](/docs/sdk-hooks-usevoicefocus--page) hook.
 
+## Props
+
+<Props of={VoiceFocusProvider} />
+
 ## Importing
 
 ```javascript
@@ -35,6 +43,7 @@ import { VoiceFocusProvider } from 'amazon-chime-sdk-component-library-react';
 
 ```jsx
 import React from 'react';
+import { VoiceFocusTransformDevice } from 'amazon-chime-sdk-js';
 import {
   MeetingProvider,
   VoiceFocusProvider,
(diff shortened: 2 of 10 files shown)


src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -42,7 +42,7 @@ const AudioInputVFControl: React.FC<Props> = ({
   const meetingManager = useMeetingManager();
   const [isLoading, setIsLoading] = useState(false);
   // When the user click on Amazon Voice Focus option, the state will change.
-  const [isVoiceFocusOn, setIsVoiceFocusOn] = useState(false);
+  const [isVoiceFocusOn, setIsVoiceFocusOn] = useState<boolean | undefined>(undefined);
```
devalevenkatesh, line 45: Earlier this was `false` now `undefined`, checking if you tested cause earlier it meant that voice focus will be off by default but now its undefined. Suggest to keep this as `false`.

If you do this for logging, below code should also work and give you initial as well as updated states right?
```
useEffect(() => {
    console.info(`Amazon Voice Focus state isVoiceFocusOn: ${isVoiceFocusOn}`);
  }, [isVoiceFocusOn]);
```

Hooks are executed in the order, so I think initial state should also get printed here.

src/providers/VoiceFocusProvider/docs/VoiceFocusProvider.stories.mdx
```diff
@@ -19,6 +19,9 @@ This provider is independent from `MeetingProvider`. You can put `VoiceFocusProv
 
 ```
 
+You should see "Amazon Voice Focus is supported." in the dev tool of the broswer if your device is sufficient for Amazon Voice Focus feature,
```
devalevenkatesh, line 22: Suggest adding below link for more information on support:
https://aws.github.io/amazon-chime-sdk-js/modules/amazonvoice_focus.html#can-i-use-amazon-voice-focus-in-my-application

Should we re-frame the sentence to, "You should see 'Amazon Voice Focus is supported.' or 'Amazon Voice Focus is not supported.' console logs in your browser developer tools based on Amazon Voice Focus support on your machine and browser. For more information, check Amazon Chime SDK for JavaScript's Amazon Voice Focus [guide](Link to guide)."

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -159,12 +159,18 @@ const VoiceFocusProvider: React.FC<Props> = ({
...
-      console.debug('Amazon Voice Focus is supported.');
+      console.info('Amazon Voice Focus is supported.');
     } else {
       console.warn('Amazon Voice Focus is not supported.');
     }
   }, [isVoiceFocusSupported]);
 
+  useEffect(() => {
+    if (voiceFocusDevice) {
+      console.info('Amazon Voice Focus transform device: ', voiceFocusDevice);
```
devalevenkatesh, line 170: Checking if this shows [Object object] in logs or how? Also, once you change the device does the earlier console log with earlier device printed still useful if someone provides us with the logs. If reference is updated then the earlier console log will also print the same device. Request you to check once.

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
michhyun1: It makes sense to keep this as false default - we want to reduce the usage of `null` and replace those with `undefined` (if it makes sense) - this doesnt apply to booleans.

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -61,6 +61,13 @@ const AudioInputVFControl: React.FC<Props> = ({
     };
   }, []);
 
+  useEffect(() => {
+    if (isVoiceFocusOn === undefined) {
```
michhyun1, line 65: === false

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
michhyun1: Also - is it possible to rename this to `isVoiceFocusCheckboxOn`?

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -61,6 +61,13 @@ const AudioInputVFControl: React.FC<Props> = ({
...
+      return;
+    }
+    console.info('Updating Amazon Voice Focus display state: ', isVoiceFocusOn)
```
michhyun1, line 68: Updating Amazon Voice Focus display state -> `Updating Amazon Voice Focus checkbox state`

src/providers/VoiceFocusProvider/index.tsx
michhyun1: Yeah it will - you can print: JSON.stringify(voiceFocusDevice, null , 2);

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
jing-chen1: The reason to set the default value as `undefined` is that I don't want to show the initial state in the browser, since it would be `false` every time at the beginning. So I would like to avoid this unnecessary log.  But seems like it is okay to print `false` as initial state every time when user joins the meeting?

`isVoiceFocusCheckboxOn` -> it is clearer, will update the name

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
jing-chen1: sure, will update it

src/providers/VoiceFocusProvider/docs/VoiceFocusProvider.stories.mdx
jing-chen1: Good point, will add the link for reference

src/providers/VoiceFocusProvider/index.tsx
jing-chen1: I have tested on meeting demo, it shows the device correctly. 
Screen Shot 2021-10-04 at 3 14 44 PM

Every time when VF transform device changes, it will print the current device. The earlier console log print the previous device, not current device. (Not sure if I understand your question correctly)

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -42,7 +42,7 @@ const AudioInputVFControl: React.FC<Props> = ({
   const meetingManager = useMeetingManager();
   const [isLoading, setIsLoading] = useState(false);
   // When the user click on Amazon Voice Focus option, the state will change.
-  const [isVoiceFocusOn, setIsVoiceFocusOn] = useState(false);
+  const [isVoiceFocusCheckboxOn, setIsVoiceFocusCheckboxOn] = useState(false);
```
devalevenkatesh, line 45: We do not use `Checkbox`, but, its a `Check` icon, so can we remove `Checkbox` and keep `isVoiceFocusCheckOn` ?

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -61,6 +61,10 @@ const AudioInputVFControl: React.FC<Props> = ({
     };
   }, []);
 
+  useEffect(() => {
+    console.info('Updating Amazon Voice Focus checkbox state: ', isVoiceFocusCheckboxOn)
```
devalevenkatesh, line 65: `checkbox` -> `check` ?
This will print when user clicks on Enable voice focus so, logging the state at sometime may not be useful to debug for builder, it would be easy to understand for us, but from builders perspective to make this easy, should we log as?
```
console.info(`User ${isVoiceFocusCheckOn ? 'turned on' : 'turned off'} Amazon voice focus.`);
```

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
jing-chen1: I saw @simmkyu comment on the background blur on the same place:
>Two similar variable names-isVideoTransformOn and isVideoTransformEnabled-are confusing. You can make them more specific, e.g., isVideoTransformCheckBoxOn.

I also feel it is a little confusing between `isVoiceFocusOn` and `isVoiceFocusEnabled`, so rename it.

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
jing-chen1: That's true, It is better for builders to understand. But since the default value of `isVoiceFocusCheckOn` is false, so it will show `User turned off Amazon Voice Focus.` when user joins the meeting. Is it okay to show that? (That's why I set `undefined` as default value, just want to skip the log of initial status)

devalevenkatesh: Yeah thats true as well, Ok so lets simply then say:
```
console.info("Amazon Voice Focus is", isVoiceFocusCheckOn);
```

jing-chen1: Changed the dependency as `isVoiceFocusEnabled`, that makes more sense. And add logs like "User turned on/off Amazon Voice Focus." in the last `useEffect` which has `isVoiceFocusCheckboxOn` as a dependency. 

Thought it over, changed the default state of `isVoiceFocusCheckboxOn` as `undefined` to avoid the initial state issue. We have used this initial state for `isVoiceFocusSupported`, [here](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/providers/VoiceFocusProvider/index.tsx#L44). There is also a similar discussion before: https://github.com/aws/amazon-chime-sdk-component-library-react/pull/583#discussion_r686356635

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
jing-chen1: Yep, this variable is internal. Do you mean to rename it as `isVoiceFocusChecked`?

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -69,6 +69,11 @@ const VoiceFocusProvider: React.FC<Props> = ({
   const addVoiceFocus = async (device: Device): Promise<Device | VoiceFocusTransformDevice> => {
     if (voiceFocusDevice) {
       const vf = await voiceFocusDevice.chooseNewInnerDevice(device);
+      if (vf instanceof VoiceFocusTransformDevice) {
```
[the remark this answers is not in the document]
jing-chen1: Yes, we can directly print log after the `if` check.

Added logs after `voiceFocusDevice.chooseNewInnerDevice(device);` and 
`const vf = await transformer?.createTransformDevice(device);`

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -67,8 +67,11 @@ const VoiceFocusProvider: React.FC<Props> = ({
   });
 
   const addVoiceFocus = async (device: Device): Promise<Device | VoiceFocusTransformDevice> => {
+    console.info('Chose a new audio input device', device);
```
devalevenkatesh, line 70: nit: Chose -> Choose ?

I used Chose, to specify that the choosing is already finished :)

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -67,8 +67,11 @@ const VoiceFocusProvider: React.FC<Props> = ({
...
+
     if (voiceFocusDevice) {
       const vf = await voiceFocusDevice.chooseNewInnerDevice(device);
+      console.info('Re-used the same internal state to create an Amazon Voice Focus transform device.');
```
devalevenkatesh, line 74: For understanding: Do we re-create the audio transform device again or re-use and replace the inner device?

src/providers/VoiceFocusProvider/index.tsx
jing-chen1: Yes, I think that the device has already been chose when `addVoiceFocus` is triggered, so use the past tense

venkatesh-devale: Then, should we say?
console.log('Add voice focus to following audio input device', device);

devalevenkatesh (approved):

Issue #629, opened 2021-09-27
Additional Devices should be optional


subhendukundu:
### What happened and what did you expect to happen?

This is more of a feature request.
```
const videoInputConfig: DeviceConfig = {
  additionalDevices: true,
};
```
[VideoInputControl.tsx#L21](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/components/sdk/MeetingControls/VideoInputControl.tsx#L21) 

We should make this configurable. We have requirement where we don't need the extra devices. Of course I can use `useVideoInputs` my own and make it as it should work. Then we have to write few additional functionality which already there.

### Have you reviewed our existing documentation?

- [X] Amazon Chime SDK for JavaScript GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] Amazon Chime SDK React Components Library GitHub [issues](https://github.com/aws/amazon-chime-sdk-component-library-react/issues)
- [X] Storybook [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/introduction--page)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-component-library-react#readme)
- [X] How-to [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/how-tos--page)
- [X] React SDK meeting [demo](https://github.com/aws/amazon-chime-sdk-component-library-react/tree/master/demo/meeting#readme---react-meeting-demo)

### Reproduction steps

Use VideoInputControl directly with Controlbar.

### Amazon Chime SDK React Components Library version

2.9.1

### What browsers are you seeing the problem on?

All

### Browser version

Not important

### Device Information

Not important

### Browser console logs

Not important

michhyun1:
@subhendukundu thanks for calling this out - we can certainly make this configureable on our end.

LeviSklaroff:
This is now a feature in the newest version in the react library. Going to close this issue, if this feature does not work for your use case please feel free to reopen this issue.

PR #633, opened 2021-09-30
Add additionalDevices as a prop instead of hardcoded

michhyun1:
**Issue #:** 
https://github.com/aws/amazon-chime-sdk-component-library-react/issues/629

**Description of changes:**
Make the additionalDevices a prop for the media control components. 

**Testing**
yes
Ran it in the meeting demo
yes - I just added a prop with a comment above - they will be picked up by storbybook documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `activeSpeakerPolicy` and `videoUplinkBandwidthPolicy` to `MeetingProvider` props.
 
 ### Changed
+- Change `additionalDevices` to a prop in `AudioInputControl`, `VideoInputControl`, `AudioInputVFControl`, `MicSelection`, and `CameraSelection` components to allow option to turn off that configuration.
 
 ### Removed
 
--- a/src/components/sdk/DeviceSelection/CameraSelection/index.tsx
+++ b/src/components/sdk/DeviceSelection/CameraSelection/index.tsx
@@ -5,6 +5,7 @@ import React from 'react';
 
 import { useVideoInputs } from '../../../../providers/DevicesProvider';
 import { useMeetingManager } from '../../../../providers/MeetingProvider';
+import { DeviceConfig } from '../../../../types';
 import DeviceInput from '../DeviceInput';
 
 interface Props {
@@ -12,14 +13,20 @@ interface Props {
   notFoundMsg?: string;
   /** The label that will be shown for camera selection, it defaults to "Camera source". */
   label?: string;
+  /** A boolean that determines whether or not to include additional sample video devices, such as "None", "Blue", "SMTP Color Bars". Defaults to true. This will be deprecated in the next major version. */
+  appendSampleDevices?: boolean;
 }
 
 export const CameraSelection: React.FC<Props> = ({
   notFoundMsg = 'No camera devices found',
   label = 'Camera source',
+  appendSampleDevices = true,
 }) => {
   const meetingManager = useMeetingManager();
-  const { devices, selectedDevice } = useVideoInputs();
+  const videoInputConfig: DeviceConfig = {
+    additionalDevices: appendSampleDevices,
+  };
+  const { devices, selectedDevice } = useVideoInputs(videoInputConfig);
 
   async function selectVideoInput(deviceId: string) {
     meetingManager.selectVideoInputDevice(deviceId);
--- a/src/components/sdk/DeviceSelection/MicSelection/index.tsx
+++ b/src/components/sdk/DeviceSelection/MicSelection/index.tsx
@@ -5,6 +5,7 @@ import React from 'react';
 
 import { useSelectAudioInputDevice } from '../../../../hooks/sdk/useSelectAudioInputDevice';
 import { useAudioInputs } from '../../../../providers/DevicesProvider';
+import { DeviceConfig } from '../../../../types';
 import DeviceInput from '../DeviceInput';
 
 interface Props {
@@ -12,14 +13,20 @@ interface Props {
   notFoundMsg?: string;
   /** The label that will be shown for microphone selection, it defaults to `Microphone source`. */
   label?: string;
+  /** A boolean that determines whether or not to include additional sample audio input devices, such as "None", "440 Hz". Defaults to true. This will be deprecated in the next major version. */
+  appendSampleDevices?: boolean;
 }
 
 export const MicSelection: React.FC<Props> = ({
   notFoundMsg = 'No microphone devices found',
   label = 'Microphone source',
+  appendSampleDevices = true,
 }) => {
+  const audioInputConfig: DeviceConfig = {
+    additionalDevices: appendSampleDevices,
+  };
   const selectAudioInput = useSelectAudioInputDevice();
-  const { devices, selectedDevice } = useAudioInputs();
+  const { devices, selectedDevice } = useAudioInputs(audioInputConfig);
 
   return (
     <DeviceInput
--- a/src/components/sdk/MeetingControls/AudioInputControl.tsx
+++ b/src/components/sdk/MeetingControls/AudioInputControl.tsx
@@ -17,10 +17,12 @@ interface Props {
   muteLabel?: string;
   /** The label that will be shown when microphone is unmuted, it defaults to `Unmute`. */
   unmuteLabel?: string;
-  /** Title attribute for the icon when muted, it defaults to `Muted microphone` */
+  /** Title attribute for the icon when muted, it defaults to `Muted microphone`. */
   mutedIconTitle?: string;
-  /** Title attribute for the icon when unmuted, it defaults to `Microphone` */
+  /** Title attribute for the icon when unmuted, it defaults to `Microphone`. */
   unmutedIconTitle?: string;
+  /** A boolean that determines whether or not to include additional sample audio input devices, such as "None", "440 Hz". Defaults to true. This will be deprecated in the next major version. */
+  appendSampleDevices?: boolean;
 }
 
 const AudioInputControl: React.FC<Props> = ({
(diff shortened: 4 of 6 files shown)


src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -29,6 +29,8 @@ interface Props {
   voiceFocusOnLabel?: string;
   /** The label that will be shown when the current input audio is not an Amazon Voice Focus device, it defaults to `Enable Amazon Voice Focus`. */
   voiceFocusOffLabel?: string;
+    /** A boolean that determines whether or not to include additional audio devices to the control bar */
+    additionalDevices?: boolean,
```
ltrung, line 33: nit: space :D

src/components/sdk/MeetingControls/VideoInputControl.tsx
```diff
@@ -15,13 +15,17 @@ import useSelectVideoInputDevice from '../../../hooks/sdk/useSelectVideoInputDev
 interface Props {
   /** The label that will be shown for video input control, it defaults to `Video`. */
   label?: string;
+  /** A boolean that determines whether or not to include additional video devices to the control bar */
+  additionalDevices?: boolean,
```
ltrung, line 19: should this be = true like above. Just try to be consistent :D

jing-chen1 (dismissed):

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Remove the audio video observers in the `audioVideoDidStop()` function instead of `leave()` function in the `MeetingManager`.
+- Change `additionalDevices` to a prop in AudioInput/VideoInput/AudioInputVF Control components to allow option to turn off that configuration
```
devalevenkatesh, line 30: Would like to see back-ticked components as well and end the log with a "." :)

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -29,6 +29,8 @@ interface Props {
   voiceFocusOnLabel?: string;
   /** The label that will be shown when the current input audio is not an Amazon Voice Focus device, it defaults to `Enable Amazon Voice Focus`. */
   voiceFocusOffLabel?: string;
+    /** A boolean that determines whether or not to include additional audio devices to the control bar */
```
devalevenkatesh, line 32: Should we add some information on what additional devices are and change the prop name since now this will be public? 

It will help builders else not sure how they can understand what additional devices are. I could not think of a good name which will explain directly but we can have a better name to reflect the devices use it seems.

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
xuesichao: nit: semicolon instead of comma

src/components/sdk/MeetingControls/AudioInputControl.tsx
```diff
@@ -21,18 +21,21 @@ interface Props {
   mutedIconTitle?: string;
   /** Title attribute for the icon when unmuted, it defaults to `Microphone` */
   unmutedIconTitle?: string;
+  /** A boolean that determines whether or not to include additional sample audio devices to the control bar */
+  appendAdditionalDevices?: boolean;
```
devalevenkatesh, line 25: Should we add sample devices to the documentation of each controls, so that builders know what they are in context of our repo? Now that you brought up sample devices, should we just say `addSampleDevices`?

Tagging @ltrung and @simmkyu for suggestion on the prop name.

michhyun1: yeah I think `sample` or `fake` or something like that would be a good distinction

src/components/sdk/MeetingControls/AudioInputControl.tsx
```diff
@@ -21,18 +21,21 @@ interface Props {
...
   mutedIconTitle,
-  unmutedIconTitle
+  unmutedIconTitle,
+  appendAdditionalDevices = true,
 }) => {
   const meetingManager = useMeetingManager();
   const { muted, toggleMute } = useToggleLocalMute();
   const audioInputConfig: DeviceConfig = {
-    additionalDevices: true,
+    additionalDevices: appendAdditionalDevices,
```
devalevenkatesh, line 38: Doing so will remove the ability to select `null` device from `AudioInputControl`, `AudioInputVFControl` as well as `VideoInputControl` due to below reason:
1. These controls depend on `AudioInputProvider`, `VideoInputProvider` respectively.
2. Here we add the additional devices which also contain 'None'. [AudioInputProvider](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/providers/DevicesProvider/AudioInputProvider.tsx#L153-L160) and [VideoInputProvider](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/providers/DevicesProvider/VideoInputProvider.tsx#L111-L120). These both pick the additional devices from [here](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/constants/additional-audio-video-devices.ts).The 'None' is used as `null` device across library I think, so we cannot skip it.

We have to refactor this logic to at least include 'None' mapping.

ltrung: Yeah but this is for UI control right. I don't think most apps need to select None as an option (especially for UI).

ltrung: Another option is instead of boolean allow them to pass in the list of additional devices they want to.

richnew10: The problem with passing a list of devices is that the selected device ends up being interpreted by our code:

https://github.com/aws/amazon-chime-sdk-component-library-react/blob/d7d7a0fbe7e1789066bd448b41c8e6b0f9233180/src/utils/device-utils.ts#L31

and so the list of additional devices has to be part of a restricted set.

michhyun1: Yeah, perhaps we can separate 'None' from the "additional devices"?

src/components/sdk/MeetingControls/VideoInputControl.tsx
```diff
@@ -15,13 +15,17 @@ import useSelectVideoInputDevice from '../../../hooks/sdk/useSelectVideoInputDev
...
-  additionalDevices: true,
-};
-
-const VideoInputControl: React.FC<Props> = ({ label = 'Video' }) => {
+const VideoInputControl: React.FC<Props> = ({ 
+  label = 'Video',
+  appendAdditionalDevices = true,
+}) => {
+  const videoInputConfig: DeviceConfig = {
+    additionalDevices: appendAdditionalDevices,
```
Isabel-hub, line 27: What will be the impact if this is made configurable ?

devalevenkatesh (commented):
Let's do this for MicSelection as well as CameraSelection as this applies to them as well.

src/components/sdk/MeetingControls/VideoInputControl.tsx
michhyun1: Ideally, the user can decide whether or not they want these additional devices in the input controls.

src/components/sdk/DeviceSelection/CameraSelection/index.tsx
```diff
@@ -5,21 +5,28 @@ import React from 'react';
...
 import { useMeetingManager } from '../../../../providers/MeetingProvider';
+import { DeviceConfig } from '../../../../types';
 import DeviceInput from '../DeviceInput';
 
 interface Props {
   /** The message that will be shown when no camera devices are found. */
   notFoundMsg?: string;
   /** The label that will be shown for camera selection, it defaults to "Camera source". */
   label?: string;
+  /** A boolean that determines whether or not to include additional sample video devices */
```
devalevenkatesh, line 16: nit: End with a "dot".

src/components/sdk/DeviceSelection/MicSelection/index.tsx
```diff
@@ -5,21 +5,28 @@ import React from 'react';
...
 import { useAudioInputs } from '../../../../providers/DevicesProvider';
+import { DeviceConfig } from '../../../../types';
 import DeviceInput from '../DeviceInput';
 
 interface Props {
   /** The message that will be shown when no microphone devices are found. */
   notFoundMsg?: string;
   /** The label that will be shown for microphone selection, it defaults to `Microphone source`. */
   label?: string;
+  /** A boolean that determines whether or not to include additional sample audio input devices */
```
devalevenkatesh, line 16: nit: End with a "dot". This comments applies for other components as well.

devalevenkatesh: Suggest adding two things:
1. What is sample devices?
2. A note that we will be deprecating this prop and remove the sample devices in next major version release.

devalevenkatesh (changes requested):
Need to add similar changes to `VideoInputBackgroundControl` as well.

src/components/sdk/DeviceSelection/CameraSelection/index.tsx
```diff
@@ -5,21 +5,28 @@ import React from 'react';
...
 import { useMeetingManager } from '../../../../providers/MeetingProvider';
+import { DeviceConfig } from '../../../../types';
 import DeviceInput from '../DeviceInput';
 
 interface Props {
   /** The message that will be shown when no camera devices are found. */
   notFoundMsg?: string;
   /** The label that will be shown for camera selection, it defaults to "Camera source". */
   label?: string;
+  /** A boolean that determines whether or not to include additional sample video devices, such as "None", "Blue", "SMTP Color Bars". This will be deprecated in the next major version. */
```
devalevenkatesh, line 16: Could you please mention that the default is true? Also, request to update the documentation here and elsewhere that with prop there will be no option to add any sample devices in next major release.

michhyun1: Done. dont the comments from these props go directly into storybook documentation? Is that good enough?

devalevenkatesh: Yeah, I meant expanding the comment here itself no other documentation.

devalevenkatesh (approved):

PR #635, opened 2021-09-30
Correctly remove event observer

devalevenkatesh:
**Issue #:** 
NA

**Description of changes:**
- Fix `eventDidReceive` observer removal.

**Testing**

2. How did you test these changes? Tested by raising this PR: https://github.com/aws-samples/amazon-chime-sdk/pull/67 against the demo app.

### DIFF
--- a/src/providers/MeetingProvider/MeetingManager.ts
+++ b/src/providers/MeetingProvider/MeetingManager.ts
@@ -137,6 +137,8 @@ export class MeetingManager implements AudioVideoObserver {
 
   private meetingEventObserverSet = new Set<(name: EventName, attributes: EventAttributes) => void>();
 
+  private eventDidReceiveRef: AudioVideoObserver;
+
   constructor(private meetingManagerConfig: MeetingManagerConfig) {
     const {
       simulcastEnabled,
@@ -164,6 +166,10 @@ export class MeetingManager implements AudioVideoObserver {
     if (videoDownlinkBandwidthPolicy) {
       this.videoDownlinkBandwidthPolicy = videoDownlinkBandwidthPolicy;
     }
+
+    this.eventDidReceiveRef = { eventDidReceive: (name: EventName, attributes: EventAttributes) => {
+      this.publishEventDidReceiveUpdate(name, attributes);
+    }};
   }
 
   initializeMeetingManager(): void {
@@ -236,10 +242,6 @@ export class MeetingManager implements AudioVideoObserver {
     this.publishActiveSpeaker();
   }
 
-  eventDidReceive = (name: EventName, attributes: EventAttributes) => {
-    this.publishEventDidReceiveUpdate(name, attributes);
-  };
-
   async initializeMeetingSession(
     configuration: MeetingSessionConfiguration,
     deviceLabels: DeviceLabels | DeviceLabelTrigger = DeviceLabels.AudioAndVideo,
@@ -278,12 +280,14 @@ export class MeetingManager implements AudioVideoObserver {
 
     this.audioVideo = this.meetingSession.audioVideo;
     this.publishAudioVideo();
+    
     // When an attendee leaves, we remove AudioVideoObservers and nullify AudioVideoFacade and MeetingSession object.
     // This results into missing few meeting events triggered with audioVideoDidStop such as meetingEnded, meetingFailed and meetingStartFailed.
     // We may also loose audioInputUnselected and videoInputUnselected events as we choose null devices when an attendee leaves.
     // When a new AudioVideoFacade object is created remove and re-add the eventDidReceive observer which wont leak.
-    this.audioVideo.removeObserver({ eventDidReceive: this.eventDidReceive });
-    this.audioVideo.addObserver({ eventDidReceive: this.eventDidReceive });
+    this.audioVideo.removeObserver(this.eventDidReceiveRef);
+    this.audioVideo.addObserver(this.eventDidReceiveRef);
+
     this.setupAudioVideoObservers();
     this.setupDeviceLabelTrigger(deviceLabels);
     await this.listAndSelectDevices(deviceLabels);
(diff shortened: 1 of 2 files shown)


src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -282,8 +284,9 @@ export class MeetingManager implements AudioVideoObserver {
     // This results into missing few meeting events triggered with audioVideoDidStop such as meetingEnded, meetingFailed and meetingStartFailed.
     // We may also loose audioInputUnselected and videoInputUnselected events as we choose null devices when an attendee leaves.
     // When a new AudioVideoFacade object is created remove and re-add the eventDidReceive observer which wont leak.
-    this.audioVideo.removeObserver({ eventDidReceive: this.eventDidReceive });
-    this.audioVideo.addObserver({ eventDidReceive: this.eventDidReceive });
+    this.eventDidReceiveObserver && this.audioVideo.removeObserver(this.eventDidReceiveObserver);
+    this.eventDidReceiveObserver = { eventDidReceive: this.eventDidReceive };
```
ltrung, line 288: Why cant we just initialize it in line 140 and just do remove it here?

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -282,8 +284,9 @@ export class MeetingManager implements AudioVideoObserver {
     // This results into missing few meeting events triggered with audioVideoDidStop such as meetingEnded, meetingFailed and meetingStartFailed.
     // We may also loose audioInputUnselected and videoInputUnselected events as we choose null devices when an attendee leaves.
     // When a new AudioVideoFacade object is created remove and re-add the eventDidReceive observer which wont leak.
-    this.audioVideo.removeObserver({ eventDidReceive: this.eventDidReceive });
-    this.audioVideo.addObserver({ eventDidReceive: this.eventDidReceive });
+    this.eventDidReceiveObserver && this.audioVideo.removeObserver(this.eventDidReceiveObserver);
```
ltrung, line 287: >  // When an attendee leaves, we remove AudioVideoObservers and nullify AudioVideoFacade and MeetingSession object.
    // This results into missing few meeting events triggered with audioVideoDidStop such as meetingEnded, meetingFailed and meetingStartFailed.
    // When a new AudioVideoFacade object is created remove and re-add the eventDidReceive observer which wont leak.

Is this still true?

ltrung, line 287: This code is confusing. Why do we need to add several audioVideo observer here?
Can we just have one?

devalevenkatesh: The audioVideoDidStart and audioVideoDidStop were kept separate due to the meeting events issue.

xuesichao: I think we don't need this condition check here, but maybe it would be helpful for readability.
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/delete, delete an item that is not in the Set won't trigger any error, it will just return `false`

src/providers/MeetingProvider/MeetingManager.ts
devalevenkatesh: Yup that makes sense, update this.

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: I think we don't need this condition check here.
According to [MDN/set/delete](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/delete#return_value), if we delete an item that is not in the Set, it won't throw any error, just return `false`.

devalevenkatesh: I removed the part to down setupAudioVideoObservers section to keep addition, removal in one place. But this still needs to be separate from the existing observers and the comment still applies tested it.

devalevenkatesh: Yes, I tested it and we still follow the logic.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -367,6 +365,13 @@ export class MeetingManager implements AudioVideoObserver {
       return;
     }
 
+    // When an attendee leaves, we remove AudioVideoObservers and nullify AudioVideoFacade and MeetingSession object.
+    // This results into missing few meeting events triggered with audioVideoDidStop such as meetingEnded, meetingFailed and meetingStartFailed.
+    // We may also loose audioInputUnselected and videoInputUnselected events as we choose null devices when an attendee leaves.
+    // When a new AudioVideoFacade object is created remove and re-add the eventDidReceive observer which wont leak.
+    this.eventDidReceiveObserver && this.audioVideo.removeObserver(this.eventDidReceiveObserver);
```
xuesichao, line 372: In case if you didn't see it, we could remove this condition check here.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -164,6 +166,8 @@ export class MeetingManager implements AudioVideoObserver {
     if (videoDownlinkBandwidthPolicy) {
       this.videoDownlinkBandwidthPolicy = videoDownlinkBandwidthPolicy;
     }
+
+    this.eventDidReceiveObserver = { eventDidReceive: this.eventDidReceive };
```
xuesichao, line 170: Is the code from L243 - L245, the `eventDidReceive` function used somewhere else? Could we do not declare this function and just add it into the `eventDidReceiveObserver` at L140?

Feel like we don't need L160-170 and L243 - L245.

src/providers/MeetingProvider/MeetingManager.ts
devalevenkatesh: Yup I have removed it, thanks!

src/providers/MeetingProvider/MeetingManager.ts
devalevenkatesh: Yup updated the code to do it, I refrained as the `eventDidReceive` was an public arrow function so dint thought to remove it, but have removed as it should not be used publicly anyhow. Thanks!

ltrung (approved):

PR #639, opened 2021-10-05
Add Amazon Voice Focus to default device in VF provider

jing-chen1:

**Description of changes:**
Add Amazon Voice Focus to the default device when a new blue tooth device is plugged in during the meeting.
This issue only happens in Chrome and when user selects "default" device as the audio input device. 

**Testing**

Should test with the PR of meeting demo: https://github.com/aws-samples/amazon-chime-sdk/pull/70

 * Open two Chrome tabs (no Safari, no FF)
 * Disable speaker in one tab
 * Disable microphone and set built in device as speaker in another tab
 * Turn on Amazon Voice Focus in the first tab
 * Apply a new device (AirPods)
 * The current device changes to AirPods
 * Then make some background noice, you should not hear any noice from the second tab since Amazon Voice Focus is enabled in the first tab.

3. If you made changes to the component library, have you provided corresponding documentation changes? Yes, in the change log

### DIFF
--- a/src/providers/DevicesProvider/AudioInputProvider.tsx
+++ b/src/providers/DevicesProvider/AudioInputProvider.tsx
@@ -1,7 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { DeviceChangeObserver } from 'amazon-chime-sdk-js';
 import React, {
   createContext,
   useContext,
@@ -10,6 +9,7 @@ import React, {
   useRef,
   useState,
 } from 'react';
+import type { Device, AudioTransformDevice, DeviceChangeObserver } from 'amazon-chime-sdk-js';
 
 import { AUDIO_INPUT } from '../../constants/additional-audio-video-devices';
 import { DeviceConfig, DeviceTypeContext } from '../../types';
@@ -17,9 +17,16 @@ import { getFormattedDropdownDeviceOptions } from '../../utils/device-utils';
 import { useAudioVideo } from '../AudioVideoProvider';
 import { useMeetingManager } from '../MeetingProvider';
 
+interface Props {
+  onDeviceReplacement?: (
+    nextDevice: string,
+    currentDevice: Device | AudioTransformDevice
+  ) => Promise<Device | AudioTransformDevice>;
+}
+
 const Context = createContext<DeviceTypeContext | null>(null);
 
-const AudioInputProvider: React.FC = ({ children }) => {
+const AudioInputProvider: React.FC<Props> = ({ children, onDeviceReplacement }) => {
   const meetingManager = useMeetingManager();
   const audioVideo = useAudioVideo();
   const [audioInputs, setAudioInputs] = useState<MediaDeviceInfo[]>([]);
@@ -30,6 +37,16 @@ const AudioInputProvider: React.FC = ({ children }) => {
   selectedInputRef.current = selectedAudioInputDevice;
   const [selectAudioInputDeviceError, setSelectAudioInputDeviceError] =
     useState(meetingManager.selectAudioInputDeviceError);
+
+  const replaceDevice = async (device: string): Promise<Device | AudioTransformDevice> => {
+    if (onDeviceReplacement) {
+      return onDeviceReplacement(
+        device,
+        meetingManager.selectedAudioInputTransformDevice
+      );
+    }
+    return device;
+  };
 
   useEffect(() => {
     meetingManager.subscribeToSelectAudioInputDeviceError(
@@ -66,6 +83,7 @@ const AudioInputProvider: React.FC = ({ children }) => {
           (device) => device.deviceId === selectedInputRef.current
         );
 
+        let nextInput: string = 'default';
         if (
           selectedInputRef.current &&
           !hasSelectedDevice &&
@@ -74,7 +92,7 @@ const AudioInputProvider: React.FC = ({ children }) => {
           console.log(
             'Previously selected audio input lost. Selecting a default device.'
           );
-          meetingManager.selectAudioInputDevice(newAudioInputs[0].deviceId);
+          nextInput = newAudioInputs[0].deviceId;
 
           // Safari and Firefox don't have this "default" as device Id
           // Only Chrome have this "default" device
@@ -82,11 +100,13 @@ const AudioInputProvider: React.FC = ({ children }) => {
           console.log(
             `Audio devices updated and "default" device is selected. Reselecting input.`
           );
-          try {
-            await audioVideo?.chooseAudioInputDevice(selectedInputRef.current);
-          } catch (e) {
-            console.error(`Error in selecting audio input device - ${e}`);
-          }
+        }
+
+        const nextDevice = await replaceDevice(nextInput);
+        try {
+          await meetingManager.selectAudioInputDevice(nextDevice);
+        } catch (e) {
+          console.error(`Error in selecting audio input device - ${e}`);
         }
 
         setAudioInputs(newAudioInputs);
@@ -119,7 +139,7 @@ const AudioInputProvider: React.FC = ({ children }) => {
       audioVideo?.removeDeviceChangeObserver(observer);
       meetingManager.unsubscribeFromDeviceLabelTriggerChange(callback);
     };
-  }, [audioVideo]);
+  }, [audioVideo, onDeviceReplacement]);
 
   const contextValue: DeviceTypeContext = useMemo(
     () => ({
--- a/src/providers/DevicesProvider/index.tsx
+++ b/src/providers/DevicesProvider/index.tsx
@@ -2,13 +2,21 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React from 'react';
+import type { Device, AudioTransformDevice } from 'amazon-chime-sdk-js';
 
 import { AudioInputProvider, useAudioInputs } from './AudioInputProvider';
 import { AudioOutputProvider, useAudioOutputs } from './AudioOutputProvider';
 import { useVideoInputs, VideoInputProvider } from './VideoInputProvider';
 
-const DevicesProvider: React.FC = ({ children }) => (
-  <AudioInputProvider>
+interface Props {
+  onDeviceReplacement?: (
+    nextDevice: string,
+    currentDevice: Device | AudioTransformDevice
+  ) => Promise<Device | AudioTransformDevice>;
+}
+
+const DevicesProvider: React.FC<Props> = ({ children, onDeviceReplacement }) => (
+  <AudioInputProvider onDeviceReplacement={onDeviceReplacement}>
     <AudioOutputProvider>
       <VideoInputProvider>{children}</VideoInputProvider>
     </AudioOutputProvider>
--- a/src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
+++ b/src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
@@ -98,12 +98,21 @@ To enable simulcast for the meeting pass the `simulcastEnabled` prop with value 
 For more information on Simulcast, check Amazon Chime JS SDK [simulcast guide](https://github.com/aws/amazon-chime-sdk-js/blob/master/guides/05_Simulcast.md).
 
 #### enableWebAudio
+
 If you want to enable Amazon Voice Focus feature, you should enable Web Audio for the meeting and pass the `enableWebAudio` prop with value set to `true`. By default, it is `false`.
 
 #### logger
 
 The `Logger` object that you want to be used in the meeting session.
 If you pass in a `Logger` object using this parameter, the `MeetingManager` will use this object instead of creating a logger based on `logLevel` and `postLogConfig` to initialize the meeting session.
+
+#### onDeviceReplacement
+
+When a selected device is disconnected, or when the Operating System (OS) default device is selected and is changed by the OS (e.g., if a Bluetooth headset is disconnected), 
+the `AudioInputProvider` encapsulated by `DevicesProvider` is forced to automatically select a new audio input. This behaves as expected when the selected device is an ordinary device provided by the browser. 
+However, when your application is using a transform device like Amazon Voice Focus, the `AudioInputProvider` needs a way to transform the newly selected device to avoid reverting to a non-transform device.
+This is accomplished by exposing a `onDeviceReplacement` prop on `MeetingProvider` and `DevicesProvider`, allowing your application to customize the behavior of this device reselection step. 
+Provide a function that accepts a `Device` as input and returns a `Device` or `AudioTransformDevice` of your choice.
 
 #### activeSpeakerPolicy
 
--- a/src/providers/MeetingProvider/index.tsx
+++ b/src/providers/MeetingProvider/index.tsx
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import {
+  AudioTransformDevice,
+  Device,
   Logger,
   LogLevel,
   VideoDownlinkBandwidthPolicy,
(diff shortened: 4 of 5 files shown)


src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -61,6 +61,19 @@ const AudioInputVFControl: React.FC<Props> = ({
...
+    // we will wrap Amazon Voice Focus to the default device.
+    const wrapVoiceFocus = async () => {
+      if (isVoiceFocusOn) {
+        const currentDevice = await addVoiceFocus(selectedDevice);
+        await meetingManager.selectAudioInputDevice(currentDevice);
+      }
+    }
+    
+    wrapVoiceFocus();
+  }, [devices[0].label]);
```
richnew10, line 75: Why not watch `devices` altogether?

Or why not specifically watch the group ID of the default device?

richnew10: If the device changes but Voice Focus isn't on, should you do something?

richnew10: So how come we won't be racing against `AudioInputProvider`?

(Why are we adding `440` etc. to the device list?!)

What problem exactly are you trying to solve here?

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -61,6 +61,19 @@ const AudioInputVFControl: React.FC<Props> = ({
     };
   }, []);
 
+  useEffect(() => {
+    // When user turns on Amazon Voice Focus and the default device changes in Chrome,
+    // we will wrap Amazon Voice Focus to the default device.
+    const wrapVoiceFocus = async () => {
```
ltrung, line 67: hmm We should not fix it here. It is only applied to the control (UI). If builders do not use the UI then they have to handle this themselves?

richnew10: It probably needs to be handled in the same place that chooses the device. A better approach would be welcome.

richnew10: Note that the JS SDK has the application do it, too:

https://github.com/aws/amazon-chime-sdk-js/blob/eeb8670b17914a6312810cb24f7b5d727a5fe4f8/demos/browser/app/meetingV2/meetingV2.ts#L1430

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
jing-chen1: >Why are we adding 440 etc. to the device list?!

I don't know...

>So how come we won't be racing against AudioInputProvider?

`AudioInputProvider` will first call `audioVideo?.chooseAudioInputDevice("default")` and then update the `devices`. After `devices` gets updated, this useEffect() will be triggered.

>What problem exactly are you trying to solve here?

When user selects "default" device as audio input device in Chrome (no Safari, no FF), and enables Amazon Voice Focus. Then a new device (AirPods) is plugged in, the "default" device will automatically changes to AirPods by `audioVideo?.chooseAudioInputDevice(selectedInputRef.current);`. However, it is just a normal audio device without VF. So we need to add VF to this "default" device. But we only `addVoiceFocus` when user click a different device or turn on VF. So need to add a new `useEffect` to keep an eye on the change of default device, then we can add VF to it in time.

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
ltrung: Yes but in React we have logic to handle this in the library itself (`<Audio/Video>InputProvider`) so its a bit weird to ask whoever use that hook to now handle that logic (which they don't have to do before).

jing-chen1: Yes, JS SDK doesn't have this issue, since it can select a transform device when the current default device changes while in React, it can just accept a string, which means it can just take a normal device, so we need to add VF to it later.
`currentAudioInputDevice: AudioInputDevice | undefined;` [line 458](https://github.com/aws/amazon-chime-sdk-js/blob/eeb8670b17914a6312810cb24f7b5d727a5fe4f8/demos/browser/app/meetingV2/meetingV2.ts#L458)

jing-chen1: > hmm We should not fix it here. It is only applied to the control (UI). If builders do not use the UI then they have to handle this themselves?

Yes, it only fix the UI, but there is no problem in our providers. Our meeting provider and audio provider don't recognize transform device, and VF provider is separate from them which is by design. 

We tried to add more logic where handle the "default" device in the `AudioInputProvider` in previous PR, but caused a Sev2, since it would need to wrap `AudioInputProvider` inside `VoiceFocusProvider` in builders' app. If builders don't use the UI, feel like they need to handle this issue themselves...

richnew10: A refactor like that was planned as a follow-up if needed; we prioritized shipping.

ltrung: Yep, discussed with @jing-chen1 and we would write a SIM for this to follow up later so we can ignore this comment for now :)

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
ltrung: > Why are we adding 440 etc. to the device list?


I think this was added since the beginning and probably just followed the JS SDK meeting demo. We have been called out on that https://github.com/aws/amazon-chime-sdk-component-library-react/issues/629

jing-chen1: >the React app will leak a burst of noise for a few hundred milliseconds, right?

Yes, you are right...
Will try the approaches you mentioned

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -67,16 +71,20 @@ const AudioInputProvider: React.FC = ({ children }) => {
           console.log(
             'Previously selected audio input lost. Selecting a default device.'
           );
-          meetingManager.selectAudioInputDevice(newAudioInputs[0].deviceId);
+          await meetingManager.selectAudioInputDevice(newAudioInputs[0].deviceId);
```
richnew10, line 74: Why does this not need similar reselection handling?

src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx
```diff
@@ -98,13 +98,20 @@ To enable simulcast for the meeting pass the `simulcastEnabled` prop with value
...
 
 The `Logger` object that you want to be used in the meeting session.
 If you pass in a `Logger` object using this parameter, the `MeetingManager` will use this object instead of creating a logger based on `logLevel` and `postLogConfig` to initialize the meeting session.
 
+#### reselection
+
+This function passes through `MeetingProvider` and `DevicesProvider` and will be triggered in `AudioInputProvider`. 
+It determines how to deal with the current audio input device when audio put devices change and the current audio input device is the default device.
+You can implement your own audio input transform device with this function.
+
```
richnew10, line 114: Nit: "audio put" => "audio input".

But rephrase this to explain what's going on first:

> When a selected device is disconnected, or when the Operating System (OS) default device is selected and is changed by the OS (_e.g._, if a Bluetooth headset is disconnected), the `AudioInputProvider` encapsulated by `DevicesProvider` is forced to automatically select a new audio input.
>
> This behaves as expected when the selected device is an ordinary device provided by the browser. However, when your application is using a transform device like Amazon Voice Focus, the `AudioInputProvider` needs a way to transform the newly selected device to avoid reverting to a non-transform device.
>
> This is accomplished by exposing a `reselection` prop on `MeetingProvider` and `DevicesProvider`, allowing your application to customize the behavior of this device reselection step. Provide a function that accepts a `Device` as input and returns a `Device` or `AudioTransformDevice` of your choice.

src/providers/DevicesProvider/index.tsx
```diff
@@ -2,13 +2,18 @@
...
+import { Device, AudioTransformDevice } from 'amazon-chime-sdk-js';
 
 import { AudioInputProvider, useAudioInputs } from './AudioInputProvider';
 import { AudioOutputProvider, useAudioOutputs } from './AudioOutputProvider';
 import { VideoInputProvider, useVideoInputs } from './VideoInputProvider';
 
-const DevicesProvider: React.FC = ({ children }) => (
-  <AudioInputProvider>
+interface Props {
+  reselection?: (device: Device) => Promise<Device | AudioTransformDevice>;
```
richnew10, line 12: Should we call this `onDeviceReplacement` or something?

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -67,16 +71,20 @@ const AudioInputProvider: React.FC = ({ children }) => {
...
           console.log(
             `Audio devices updated and "default" device is selected. Reselecting input.`
           );
+          let currentDevice: Device | AudioTransformDevice = 'default';
+          if (reselection) {
+            currentDevice = await reselection(currentDevice);
+          }
           try {
-            await audioVideo?.chooseAudioInputDevice(selectedInputRef.current);
+            await meetingManager.selectAudioInputDevice(currentDevice);
```
richnew10, line 87: Was this a bug?

richnew10 (changes requested):

src/providers/DevicesProvider/AudioInputProvider.tsx
jing-chen1: Good catch, will add the same logic here

src/providers/DevicesProvider/AudioInputProvider.tsx
jing-chen1: I think the previous code works without VF logic. But after we integrate the VF feature, we need to call meeting manager instead of `audioVideo`

src/providers/DevicesProvider/index.tsx
jing-chen1: I am not good at naming... 
will change to `onDeviceReplacement`

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -67,16 +79,18 @@ const AudioInputProvider: React.FC = ({ children }) => {
           console.log(
             'Previously selected audio input lost. Selecting a default device.'
           );
-          meetingManager.selectAudioInputDevice(newAudioInputs[0].deviceId);
+          currentDevice = await replaceDevice(newAudioInputs[0].deviceId);
+          await meetingManager.selectAudioInputDevice(currentDevice);
```
michhyun1, line 83: So we're basically saying - pass in your own implementation for selecting a default device? Do we have demo changes to showcase the correct way to do this?

michhyun1: Or can we provide like a default implementation that they can override?

jing-chen1: Here is the demo change: https://github.com/aws-samples/amazon-chime-sdk/pull/70/files

Builders can implement their own `onDeviceReplacement` method to create a transform device when audio devices change or just ignore this, then we will pass a normal device instead. 

Just realized that, not only when default device changes, we need to call this `onDeviceReplacement` function, also when the selected device is disconnected.

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -74,16 +92,18 @@ const AudioInputProvider: React.FC = ({ children }) => {
           console.log(
             'Previously selected audio input lost. Selecting a default device.'
```
richnew10, line 93: At this point you should probably clean this `if` up and unify the error handling:

```
let nextInput: string;
if (…) {
  console.log('… Selecting a default device.');
  nextInput = newAudioInputs[0].deviceId;
} else {
  console.log('… Reselecting input.');
  nextInput = 'default';
}
const nextDevice = await replaceDevice(nextInput);
try {
  await meetingManager.selectAudioInputDevice(nextDevice);
  …
```

src/providers/DevicesProvider/index.tsx
```diff
@@ -2,13 +2,21 @@
...
-  <AudioInputProvider>
+interface Props {
+  onDeviceReplacement?: (
+    nextDevice: string,
+    currentDevice: Device | AudioTransformDevice
+  ) => Promise<Device | AudioTransformDevice>;
+}
+
+const DevicesProvider: React.FC<Props> = ({ children, onDeviceReplacement }) => (
+  <AudioInputProvider onDeviceReplacement={onDeviceReplacement}>
```
richnew10, line 19: IIRC React lets you write `{onDeviceReplacement}` instead.

src/providers/MeetingProvider/index.tsx
```diff
@@ -73,7 +83,7 @@ export const MeetingProvider: React.FC<Props> = ({
     <MeetingContext.Provider value={meetingManager}>
       <MeetingEventProvider>
         <AudioVideoProvider>
-          <DevicesProvider>
+          <DevicesProvider onDeviceReplacement={onDeviceReplacement}>
```
richnew10, line 86: Same.

richnew10 (changes requested):
Please rebase.

src/providers/DevicesProvider/AudioInputProvider.tsx
jing-chen1: Good point, will update the code

src/providers/DevicesProvider/index.tsx
jing-chen1: Thanks for the advice, seems like I need to write like this
```
<AudioInputProvider {...{onDeviceReplacement}}>
```
So just keep it as is. Please let me know if you are fine with it

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -66,6 +83,7 @@ const AudioInputProvider: React.FC = ({ children }) => {
           (device) => device.deviceId === selectedInputRef.current
         );
 
+        let nextInput: string = '';
```
richnew10, line 86: You should not need the assignment here, but perhaps TypeScript's analysis is weak. If so (try it), start with `default`.

src/providers/DevicesProvider/index.tsx
```diff
@@ -2,13 +2,21 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React from 'react';
+import { Device, AudioTransformDevice } from 'amazon-chime-sdk-js';
```
richnew10, line 5: Can this be `import type`?

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -10,16 +9,24 @@ import React, {
   useRef,
   useState,
 } from 'react';
+import { Device, AudioTransformDevice, DeviceChangeObserver } from 'amazon-chime-sdk-js';
```
richnew10, line 12: Can this be `import type`?

src/providers/DevicesProvider/index.tsx
richnew10: I was thinking of Svelte. As-is is fine. Thanks for checking!

richnew10 (changes requested):

src/providers/DevicesProvider/AudioInputProvider.tsx
jing-chen1: Tried it without the assignment, got an error like `Variable 'nextInput' is used before being assigned. `
Set the value to `default`.

src/providers/DevicesProvider/index.tsx
jing-chen1: Sure, updated it

src/providers/DevicesProvider/AudioInputProvider.tsx
jing-chen1: Updated it

richnew10 (approved):
If it works, this looks good to me!

Issue #642, opened 2021-10-08
ControlBarButton button focus issue


kishan-tocca:
### What happened and what did you expect to happen?

I used this element for controllbar view , while clicking any button the focus was on that button and its getting fired by keyboard keys (Enter and Spacebar)
I expect to event don't get fired on keyboard keys

### Have you reviewed our existing documentation?

- [X] Amazon Chime SDK for JavaScript GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] Amazon Chime SDK React Components Library GitHub [issues](https://github.com/aws/amazon-chime-sdk-component-library-react/issues)
- [X] Storybook [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/introduction--page)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-component-library-react#readme)
- [X] How-to [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/how-tos--page)
- [X] React SDK meeting [demo](https://github.com/aws/amazon-chime-sdk-component-library-react/tree/master/demo/meeting#readme---react-meeting-demo)

### Reproduction steps

just once click on any controllbar button and after that click on keyboard Enter or Spacebar keys

### Amazon Chime SDK React Components Library version

2.1.0

### What browsers are you seeing the problem on?

Google chrome , firefox

### Browser version

Version 94.0.4606.61 (Official Build) (64-bit)

### Device Information

Ubuntu 20.04.3 LTS

### Browser console logs

![screenshot-localhost_3000-2021 10 08-15_36_25](https://user-images.githubusercontent.com/66630960/136538341-b8061df2-8093-4b8a-97f0-b7f766546d57.png)

ldai1:
Hi,

Thanks for using Amazon Chime SDK Component Library. Just to confirm that you want to prevent the button from receiving focus on click, right? This might has some implications for accessibility and we need to double check if this is the consistent focus state for all our buttons and inputs before we can make any changes. Will come back to you once we are done with our research.

kishan-tocca:
> Hi,
> 
> Thanks for using Amazon Chime SDK Component Library. Just to confirm that you want to prevent the button from receiving focus on click, right? This behavior might has some implications for accessibility and we need to double check if this is the consistent focus state for all our buttons and inputs before we can make any changes. Will come back to you once we are done with our research. Thanks for bringing this up.

[...]

PR #648, opened 2021-10-16
Remove button focus after the control bar button is clicked

ldai1:
**Issue #:** 
https://github.com/aws/amazon-chime-sdk-component-library-react/issues/642

**Description of changes:**
Remove button focus after the control bar button is clicked

**Testing**

2. How did you test these changes? Generate storybook for control bar button and use it for testing


ldai1:
We will keep the component as it is for accessibility reasons. One of the reasons for this is so that blind / low vision users know where they are in the DOM. If we remove the focus and following click events, then we’re forcing that user to start at the top of the DOM again and go and re-find their place in the UI. It will also be an usability problem if users want to fire the button with keyboard events.

For users who want to add customized functions when dealing with keyboard events, we would recommend to pass a callback to the button and include whatever behavior you want in that callback, such as disable focus for key-down events.

PR #661, opened 2021-11-01
Update PreviewVideo to support VideoTransformedDevices, and Add VideoFilterSelection component

michhyun1:
**Issue #:** 

**Description of changes:**
Update PreviewVideo to show the `selectedVideoInputTransformDevice` rather than the `selectedVideoInputDevice`. The difference is that the `selectedVideoInputTransformDevice` can be a regular `Device` but also a `VideoTransformDevice` - that means the PreviewVideo can now show video streams that have been transformed.

Next, add a `VideoFilterSelection` component that will simply create a background blur `VideoTransformDevice` and then call `meetingManager.selectVideoInputDevice` to select the background blur video transform device, thus publishing a new `selectedVideoInputTransformDevice`. The PreviewVideo component will listen to any changes on `selectedVideoInputTransformDevice` and update the video element with that video stream. 

Also, fixes a small bug where we were calling `meetingManager.selectVideoInputDevice` on accident in the `VideoInputBackgroundBlurControl` component's useEffect hook. The solution was to take the `meetingManager.selectVideoInputDevice` call out of the useEffect hook so that it wasn't called on mount.

Related demo changes are here: https://github.com/aws-samples/amazon-chime-sdk/pull/86

image

**Testing**
yes

Via the demo app. Changes for the demo app will be posted soon.

Yes, documentation has been updated for `PreviewVideo` and a new doc for `VideoFilterSelection`.

### DIFF
--- a/src/components/sdk/DeviceSelection/CameraSelection/VideoFilterSelection.tsx
+++ b/src/components/sdk/DeviceSelection/CameraSelection/VideoFilterSelection.tsx
@@ -1,0 +1,77 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+  Device,
+  isVideoTransformDevice,
+  VideoTransformDevice,
+} from 'amazon-chime-sdk-js';
+import React, { useEffect, useState } from 'react';
+
+import { useBackgroundBlur } from '../../../../providers/BackgroundBlurProvider';
+import { useMeetingManager } from '../../../../providers/MeetingProvider';
+import { Checkbox } from '../../../ui/Checkbox';
+import { FormField } from '../../../ui/FormField';
+
+interface Props {
+  /** Label shown for video filter selection, by default it is "Blur my background" */
+  label?: string;
+}
+
+export const VideoFilterSelection: React.FC<Props> = ({
+  label = 'Blur my background',
+}) => {
+  const { isBackgroundBlurSupported, createBackgroundBlurDevice } =
+    useBackgroundBlur();
+  const [isLoading, setIsLoading] = useState(false);
+  const meetingManager = useMeetingManager();
+  // TODO: Move this to the Video Input Provider and expose only one selected Video Input device state
+  const [device, setDevice] = useState<Device | VideoTransformDevice | null>(
+    meetingManager.selectedVideoInputTransformDevice
+  );
+
+  // TODO: Move this to the Video Input Provider and expose only one selected Video Input device state
+  useEffect(() => {
+    meetingManager.subscribeToSelectedVideoInputTransformDevice(setDevice);
+    return () => {
+      meetingManager.unsubscribeFromSelectedVideoInputTranformDevice(setDevice);
+    };
+  }, []);
+
+  const toggleBackgroundBlur = async () => {
+    if (isLoading) {
+      return;
+    }
+    if (!isVideoTransformDevice(device)) {
+      setIsLoading(true);
+      if (isBackgroundBlurSupported) {
+        const transformedVideoDevice = await createBackgroundBlurDevice(device);
+        await meetingManager.selectVideoInputDevice(transformedVideoDevice);
+      } else {
+        meetingManager.logger?.warn(
+          'Background blur processor is not supported yet.'
+        );
+      }
+    } else {
+      if ('chooseNewInnerDevice' in device) {
+        setIsLoading(true);
+        // @ts-ignore
+        const deviceId = await device.intrinsicDevice();
+        await meetingManager.selectVideoInputDevice(deviceId);
+      }
+    }
+    setIsLoading(false);
+  };
+
+  return (
+    <FormField
+      field={Checkbox}
+      onChange={toggleBackgroundBlur}
+      value={'Background Blur'}
+      checked={isVideoTransformDevice(device)}
+      label={label}
+    />
+  );
+};
+
+export default VideoFilterSelection;
--- a/src/components/sdk/DeviceSelection/docs/VideoFilterSelection.stories.mdx
+++ b/src/components/sdk/DeviceSelection/docs/VideoFilterSelection.stories.mdx
@@ -1,0 +1,47 @@
+import { Props } from '@storybook/addon-docs/blocks';
+import VideoFilterSelection from '../CameraSelection/VideoFilterSelection';
+
+<Meta title="SDK Components/DeviceSelection/Camera/VideoFilterSelection" />
+
+# VideoFilterSelection
+
+The `VideoFilterSelection` component renders a `Checkbox FormField` to select a background blur device. Currently, the only option is background blur.
+
+The component depends on the `DevicesProvider`. If you are using `MeetingProvider`, `DevicesProvider` is rendered by default.
+
+## Importing
+
+```javascript
+import { VideoFilterSelection } from 'amazon-chime-sdk-component-library-react';
+```
+
+## Usage
+
+```jsx
+import React from 'react';
+import {
+  BackgroundBlurProvider,
+  MeetingProvider,
+  VideoFilterSelection
+} from 'amazon-chime-sdk-component-library-react';
+
+const App = () => {
+  return (
+    <BackgroundBlurProvider >
+      <MeetingProvider>
+        <VideoFilterSelection />
+      </MeetingProvider>
+    </BackgroundBlurProvider >
+  );
+};
+```
+
+## Props
+
+<Props of={VideoFilterSelection} />
+
+### Dependencies
+
+- `BackgroundBlurProvider`
+- `MeetingProvider`
+- `DevicesProvider`, If you are using `MeetingProvider`, `DevicesProvider` is rendered by default
--- a/src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
+++ b/src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
@@ -39,7 +39,6 @@ const VideoInputBackgroundBlurControl: React.FC<Props> = ({
   const { devices, selectedDevice } = useVideoInputs();
   const { isVideoEnabled, toggleVideo } = useLocalVideo();
   const { isBackgroundBlurSupported, createBackgroundBlurDevice } = useBackgroundBlur();
-  const [isVideoTransformChecked, setIsVideoTransformChecked] = useState(false);
   const [isLoading, setIsLoading] = useState(false);
   const [dropdownWithVideoTransformOptions, setDropdownWithVideoTransformOptions] = useState<ReactNode[] | null>(null);
   const [activeVideoDevice, setDevice] = useState<Device | VideoTransformDevice | null>(meetingManager.selectedVideoInputDevice);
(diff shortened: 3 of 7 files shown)


src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
```diff
@@ -54,6 +54,31 @@ const VideoInputBackgroundBlurControl: React.FC<Props> = ({
     };
   }, []);
 
+  async function toggleBackgroundBlur() {
```
ltrung, line 57: Why do we need to make this change? This function depends on `isVideoTransformChecked` so I thought the previous change makes more sense.

src/components/sdk/DeviceSelection/CameraSelection/VideoFilterSelection.tsx
```diff
@@ -0,0 +1,75 @@
...
+      value: 'unselected',
+    },
+    {
+      label: 'Background Blur',
+      value: 'Background Blur'
+    }
+  ];
+
+
+  async function selectFilter(e: ChangeEvent<HTMLSelectElement>) {
```
ltrung, line 38: Since there is one value, I don't think we should use a selection here. Please consult UX but it should be just a checkbox somewhere.

ltrung (commented):
Have you tested with changing video quality when background blur is on? I recently fixed it in JS SDK https://github.com/aws/amazon-chime-sdk-js/pull/1700

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
michhyun1: There was an extra call to `selectVideoInputDevice` when we call this as part of useEffect. When you use a useEffect hook, the useEffect will be triggered when the component mounts, which is not the behaviour we wanted. We just want this to be a function triggered by on click event.

src/components/sdk/PreviewVideo/index.tsx
```diff
@@ -21,39 +22,42 @@ const StyledPreview = styled(VideoTile)`
...
-
-    return () => {
-      mounted = false;
-
-      if (videoElement) {
-        audioVideo.stopVideoPreviewForVideoInput(videoElement);
-        if (isVideoEnabled) setIsVideoEnabled(false);
-      }
-    };
-  }, [audioVideo, selectedDevice]);
```
devalevenkatesh, line 56: Should this still have `audioVideo` dependency?

Also, should we keep the start and the cleanup steps such as stopping preview for `PreviewVideo` component APIs in one hook and the transform device subscription in the first one?

src/components/sdk/DeviceSelection/CameraSelection/VideoFilterSelection.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { isVideoTransformDevice } from 'amazon-chime-sdk-js';
+
+interface Props {
+  /** Label shown for video filter selection, by default it is "Video Filter" */
+  label?: string;
+  /** Label shown in the dropdown when no video filter has been selected yet, by default it is "Select video filter" */
+  labelForUnselected?: string;
+}
+
+export const VideoFilterSelection: React.FC<Props> = ({
```
devalevenkatesh, line 19: Should we not include this component as part of the library but a component in the demo? Selection can be a array or a checkbox based on UX since we have a single processor so I think meeting demo component as an example would be more appropriate and based on customer requests we can pull this as a component in library. Just a thought.

src/components/sdk/DeviceSelection/CameraSelection/VideoFilterSelection.tsx
```diff
@@ -0,0 +1,75 @@
...
+  label = 'Video Filter',
+  labelForUnselected = 'Select video filter',
+}) => {
+  const {isBackgroundBlurSupported, createBackgroundBlurDevice} = useBackgroundBlur();
+  const [filterName, setFilterName] = useState('unselected');
+  const meetingManager = useMeetingManager();
+  const filterOptions = [
```
devalevenkatesh, line 26: Another reason this may be a good demo component is localization of these filter options label, we may want to ask builders to provide props so they can customize it. Still the value will be english if we are to handle it in library else we have to provide guidance on what values to provide for builders to improve on localization here.

src/components/sdk/DeviceSelection/docs/VideoFilterSelection.stories.mdx
```diff
@@ -0,0 +1,43 @@
...
+import React from 'react';
+import {
+  MeetingProvider,
+  VideoFilterSelection
+} from 'amazon-chime-sdk-component-library-react';
+
+const App = () => {
+  return (
+    <MeetingProvider>
+      <VideoFilterSelection />
```
devalevenkatesh, line 30: This needs to be wrapped by a `BackgroundBlurProvider`, else the `useBackgroundBlur` hook will throw an error.

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
ltrung: Then we should pass the new value of `isVideoTransformChecked` inside the function here from the clickEvent.

src/components/sdk/DeviceSelection/CameraSelection/VideoFilterSelection.tsx
michhyun1: okay - changed to a checkbox per Jaimes feedback

src/components/sdk/DeviceSelection/CameraSelection/VideoFilterSelection.tsx
michhyun1: There are other selection components in the library - but this component is not "customizable" all that much. I think thats why we might want it as a demo component, but anyone who uses the BackgroundBlurProvider might want this component, and we wouldnt be releasing this component if it is only in the demo component. So I think we can include it in the react lib.

src/components/sdk/DeviceSelection/docs/VideoFilterSelection.stories.mdx
```diff
@@ -0,0 +1,43 @@
+import { Props } from '@storybook/addon-docs/blocks';
+import VideoFilterSelection from '../CameraSelection/VideoFilterSelection';
+
+<Meta title="SDK Components/DeviceSelection/Camera/VideoFilterSelection" />
+
+# VideoFilterSelection
+
+The `VideoFilterSelection` component renders a `Select FormField` to select a video filter. Currently, the only option is background blur.
```
nainkunal933, line 8: Shouldn't the document say `Check FormField` now?

src/components/sdk/PreviewVideo/index.tsx
```diff
@@ -21,39 +22,42 @@ const StyledPreview = styled(VideoTile)`
...
+    }
     async function startPreview() {
       if (!audioVideo) {
         return;
       }
-
-      await audioVideo.chooseVideoInputDevice(selectedDevice);
-      if (videoEl.current && mounted) {
+      await meetingManager.selectVideoInputDevice(device);
+      if (videoEl.current) {
```
nainkunal933, line 54: What is the reason for swapping `chooseVideoInputDevice` for `selectVideoInputDevice` in the implementation here?

src/components/sdk/PreviewVideo/index.tsx
michhyun1: Yup, my bad. I can separate the subscription one, but the start and stop preview cant depend on the same useEffect dependencies. 

We should only call stopPreview when the audioVideo changes. The startPreview should be called when the device changes, but not when the audioVideo changes. I'll show you in the next revision.

src/components/sdk/DeviceSelection/docs/VideoFilterSelection.stories.mdx
michhyun1: yup - i changed that.

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
michhyun1: Ok - removed the ui state and depend on if the current selected device is a transformed device or not.

src/components/sdk/PreviewVideo/index.tsx
michhyun1: we want to rely on the meetingManager to wrap the sdk instead of calling the js sdk directly.

src/components/sdk/DeviceSelection/docs/VideoFilterSelection.stories.mdx
nainkunal933: Thanks :)

src/components/sdk/DeviceSelection/CameraSelection/VideoFilterSelection.tsx
michhyun1: This is just a solution for background blur only. Its not a generic component - we may add that in the future.

devalevenkatesh: Sorry, but localization issue still stands, If builders want to provide label in their own language should no it be a prop like the `labelForUnselected` ?

ltrung: This code is no longer there right? We changed it to checkbox now.

src/components/sdk/PreviewVideo/index.tsx
devalevenkatesh: Yeah that is correct, but we would still need a valid `audioVideo` to call the `stopVideoPreviewForVideoInput` API hence thought so, to avoid calling startPreview on audioVideo changes we can check against isVideoEnabled state I think right?

src/components/sdk/DeviceSelection/CameraSelection/VideoFilterSelection.tsx
```diff
@@ -0,0 +1,68 @@
...
+    } else {
+      if ('chooseNewInnerDevice' in device) {
+        setIsLoading(true);
+        // @ts-ignore
+        const deviceId = await device.intrinsicDevice();
+        await meetingManager.selectVideoInputDevice(deviceId);
+      }
+    }
+    setIsLoading(false);
+  }
```
devalevenkatesh, line 55: nit: End this arrow function with a semicolon. Normal functions we have not used any semicolons but for arrow functions we have use, please correct me if this is corrected by linter.

src/components/sdk/DeviceSelection/docs/VideoFilterSelection.stories.mdx
```diff
@@ -0,0 +1,47 @@
+import { Props } from '@storybook/addon-docs/blocks';
+import VideoFilterSelection from '../CameraSelection/VideoFilterSelection';
+
+<Meta title="SDK Components/DeviceSelection/Camera/VideoFilterSelection" />
+
+# VideoFilterSelection
+
+The `VideoFilterSelection` component renders a `Checkbox FormField` to select a background blur device. Currently, the only option is background blur.
```
devalevenkatesh, line 8: Do we want to callout how this works with `PreviewVideo` as well as `LocalVideo` ? Also, suggest adding `PreviewVideo` component in the usage example. It might help builder in knowing when to use this component.

devalevenkatesh: This component, I think is very close to `VideoInputBackgroundControl` as well so we should provide some guidance on the differences. Also, one more question, since actually both `VideoInputBackgroundControl` and now the `VideoFilterSelection` are just selecting the transform devices just that the UI is different, should we or can we implement or re-use the creation and changing of video transform device or normal video input device and not repeat in two places?

michhyun1: hmm I think that would be possible, maybe I can put this component in the VideoInputBackgroundControl later.

src/components/sdk/DeviceSelection/CameraSelection/VideoFilterSelection.tsx
michhyun1: thanks

src/components/sdk/PreviewVideo/index.tsx
michhyun1: yeah, but I think its better to seperate logic by useEffects, rather than trying to stuff everything into one useEffect and then use state to determine what code path to run. I've tested that we still are able to call stopVideoPreview when the audioVideo changes.

devalevenkatesh: Yup agree just wanted to have separate relevant ones together.

devalevenkatesh (approved):
Looks fine, but few points if can do as a follow-up:
1. Check if we can re-use the logic and use in both the `VideoInputBackgroundControl` and `VideoFIlterSelection`.
2. Update the `VideoFIlterSelection` to call out how to use it with `PreviewVideo` or `LocalVideo` is possible.

PR #680, opened 2021-11-23
Improved roster test App to facilitate testing

xuesichao:
**Issue #:** 

**Description of changes:**
* Add `MeetingStatusDisplay` component to help check the meeting join and meeting leave.
* Update Webpack to only add proxy to Chime API request.
  * Remove `bypass`
  * Add `historyApiFallback`
* Change the `data-testid` of `<code>` from `attendeeId` based to `attendeeName` based.

**Testing**
N/A

Run the demo and confirm it works.

N/A

### DIFF
--- a/integration/app/test-demo/webpack.config.js
+++ b/integration/app/test-demo/webpack.config.js
@@ -36,15 +36,11 @@ module.exports = {
   ],
   devServer: {
     proxy: {
-      '/': {
-        target: 'http://127.0.0.1:8080',
-        bypass: function (req, _res, _proxyOptions) {
-          if (req.headers.accept.indexOf('html') !== -1) {
-            console.log('Skipping proxy for browser request.');
-            return `/${app}.html`;
-          }
-        }
-      }
+      context: ['/join', '/attendee'],
+      target: 'http://127.0.0.1:8080',
+    },
+    historyApiFallback: {
+      index: `/${app}.html`,
     },
     hot: false,
     port: 9000,
--- a/integration/app/test-demo/src/components/MeetingStatusDisplay.tsx
+++ b/integration/app/test-demo/src/components/MeetingStatusDisplay.tsx
@@ -1,0 +1,12 @@
+import React from 'react';
+import {
+  MeetingStatus,
+  useMeetingStatus,
+} from "amazon-chime-sdk-component-library-react";
+
+const MeetingStatusDisplay: React.FC = () => {
+  const meetingStatus = useMeetingStatus();
+  return <p id="meeting-status">Meeting Status: {MeetingStatus[meetingStatus]}</p>;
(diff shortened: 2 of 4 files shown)


integration/app/test-demo/webpack.config.js
```diff
@@ -36,15 +36,11 @@ module.exports = {
...
-          if (req.headers.accept.indexOf('html') !== -1) {
-            console.log('Skipping proxy for browser request.');
-            return `/${app}.html`;
-          }
-        }
-      }
+      context: ['/join', '/attendee'],
+      target: 'http://127.0.0.1:8080',
+    },
+    historyApiFallback: {
```
ltrung, line 42: Why do we need to enable this?

xuesichao: Only the HTTP requests that start with `/join` and `/attendee` will be redirected to the `server.js` endpoint 'http://127.0.0.1:8080' (via `proxy`). 

For the rest of requests we need a "entry point" to render the HTML page. The `historyApiFallback` will serve`meeting.html` for these requests. e.g. 'http://localhost:9000', 'http://localhost:9000:roster-test'

xuesichao: The goal of this change is to further separate the front end and back end. 

Before we applies proxy to all requests. We skip proxy for browser requests and then serve `meeting.html` for them. Like what we did before in:
```js
 bypass: function (req, _res, _proxyOptions) {
  if (req.headers.accept.indexOf('html') !== -1) {
    console.log('Skipping proxy for browser request.');
    return `/${app}.html`;
  }
```

Now the proxy only applies to requests start with `/join` and `/attendee`, for the rest we directly serve the `meeting.html` for them.

ltrung: So from the description here https://webpack.js.org/configuration/dev-server/#devserverhistoryapifallback. It seems it is used as a fallback page instead of 404 so does not seem like what we want.

xuesichao: I tried to find some other way, but didn't find any specific option for this situation. And according to this answer, it seems to be a common use case. 
https://stackoverflow.com/questions/31945763/how-to-tell-webpack-dev-server-to-serve-index-html-for-any-route

If we don't add it, the request will jump directly to a web server page, which is somehow a 404 in my opinion, check the below image.
image

xuesichao:
The Roster Test App doesn't really test the `useAudioVideo` hook, since removed it from the hook coverage.

integration/app/test-demo/webpack.config.js
ltrung: Got it.

ltrung (approved):

PR #681, opened 2021-11-23
Set up Selenium and Mocha

xuesichao:
**Issue #:** 
N/A

**Description of changes:**
* Import Selenium WebDriver and Mocha for integration testing.
* `BasePage.js` encapsulates some common WebDriver operations.
* `BaseTestPage.js` encapsulates opening demo URL, joining the meeting and leaving the meeting operations and corresponding checks.
* `DriverFactory.js ` offers the ability to create WebDriver and start testing.
* `config.js` stores the browser config of Firefox and Chrome when testing using Sauce Labs

**Testing**
N/A

Write an integration testing for roster related feature, and confirm they work well.

N/A


ltrung (commented):
Need to update the github action to run this test right?

integration/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "integration-tests",
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "test": "npm install && mocha test/*",
+    "test:fast": "mocha test/*"
+  },
+  "author": "",
+  "license": "ISC",
```
ltrung, line 10: why ISC?

integration/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "integration-tests",
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "test": "npm install && mocha test/*",
+    "test:fast": "mocha test/*"
+  },
+  "author": "",
```
ltrung, line 9: probably want to remove it and update the description. The version should probably be `1.0.0`

integration/package.json
xuesichao: Forgot to check these info, will update them.

integration/pages/BaseTestPage.js
```diff
@@ -0,0 +1,68 @@
+const BasePage = require('./BasePage');
+const assert = require('assert');
+const { By } = require('selenium-webdriver');
+
+const ELEMENTS = {
```
ltrung, line 5: Should this be RosterTestPage?

integration/pages/BaseTestPage.js
```diff
@@ -0,0 +1,68 @@
+const BasePage = require('./BasePage');
```
ltrung, line 1: Should this be RosterTestPage since this is for the Roster Test App?

integration/utils/config.js
```diff
@@ -0,0 +1,57 @@
+module.exports = {
+  baseUrl: 'http://localhost:9000',
+  host: process.env.HOST || 'chrome',
+  sauce: {
```
nainkunal933, line 4: There should be a config for Safari as well here. We can exclude safari from tests if there are compatibility issues but by default we should have the support.

integration/pages/BasePage.js
```diff
@@ -0,0 +1,52 @@
...
+  constructor(driver) {
+    this.driver = driver;
+  }
+
+  async visit(route) {
+    await this.driver.get(config.baseUrl + route);
+  }
+
+  find(locator) {
+    return this.driver.findElement(locator);
```
ltrung, line 14: Isn't this an async function so find should be async too right?

xuesichao: `this.driver.findElement(locator)` is NOT an async function,
`this.driver.findElements(locator)` is an async function.

It's weird but correct.

integration/pages/BaseTestPage.js
xuesichao: There is a `RosterTestPage.js` file, my next PR will include it. Since this PR is to setup Mocha and Selenium I didn't include that file. 

This `BasePage.js` encapsulates some common WebDriver operations, e.g. `find`, `findAll`, `isDisplayed`, ...
These are very basic operations, every tests will use them.

The `BaseTestPage.js` encapsulates opening demo URL, joining the meeting and leaving the meeting operations and corresponding checks. These are common operations that every tests will use too.

Their relation is:
`RosterTestPage extends BaseTestPage`
`BaseTestPage extends BasePage`

integration/pages/BasePage.js
ltrung: From their main doc  https://www.selenium.dev/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_WebDriver.html. You can see they use then.
image

integration/utils/config.js
xuesichao: Yeah, make sense, I will add Safari in my revision. 👍

ltrung: It should be a separate function to get config for Chrome and Firefox etc. You can add Safari later if we want (We do not run Safari in JS SDK either).

integration/pages/BasePage.js
xuesichao: Yeah, but if you check their doc and compare `findElement` and `findElements`:
https://www.selenium.dev/documentation/webdriver/web_element/#find-element
image

xuesichao: @ltrung check here, they are same:
https://www.selenium.dev/selenium/docs/api/javascript/module/selenium-webdriver/lib/webdriver_exports_WebElementPromise.html

So I think it's better to make it `async` to keep functions consistent. Will update later

integration/utils/config.js
xuesichao: Yeah, I can delete this file and make it into a function like how we did in JS SDK. So I won't add Safari for now.

integration/pages/BaseTestPage.js
```diff
@@ -0,0 +1,68 @@
...
+      10000
+    );
+    assert(isDisplayed, 'App was not loaded successfully');
+  }
+
+  async checkIfAttendeeHasLeftMeeting() {
+    const isMatched = await this.waitUntil(async () => {
+      const meetingStatus = await this.find(ELEMENTS.meetingStatus).getText();
+      return meetingStatus === 'Meeting Status: Left';
+    }, 10000);
```
ltrung, line 62: 10s is really long. Can it be shorter?

integration/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "license": "ISC",
+  "dependencies": {
+    "chromedriver": "^96.0.0",
+    "geckodriver": "^2.0.4",
+    "mocha": "^9.1.3",
+    "selenium-webdriver": "^4.1.0"
+  },
+  "mocha": {
+    "timeout": 120000
```
ltrung, line 18: what is this for?

integration/pages/BaseTestPage.js
ltrung: Got it. Is there a reason we can't have everything in this PR?

integration/pages/BaseTestPage.js
xuesichao: This is not a fixed wait time. It means it will wait until the change appears but no longer than 10s. I don't have enough data to choose a proper timeout, so make it 10s. In my local test, it usually need 0.3s, but not sure how it goes when running on Sauce Labs.
Do you have any suggestions?

ltrung (changes requested):
Need to include GitHub action and an actual test :)

integration/pages/BasePage.js
```diff
@@ -0,0 +1,52 @@
...
+    await this.find(locator).click();
+  }
+
+  async type(locator, inputText) {
+    await this.find(locator).sendKeys(inputText);
+  }
+
+  async isDisplayed(locator, timeout) {
+    if (timeout) {
+      await this.driver.wait(until.elementLocated(locator), timeout);
```
ltrung, line 31: Do we need this or can we just check for elementIsVisible?

integration/pages/BaseTestPage.js
ltrung: Yeah i got that but 10s is really long. Can we just set it to 5 for now? If it takes longer than 5s then we probably want to fail the test :D

integration/pages/BasePage.js
ltrung: @LeviSklaroff Maybe we could use this instead of writing our own loop :D

integration/package.json
xuesichao: The maximum time for this test to run.
In this case setting the timeout is 120s. This way a test will be given 120s to run, then will quit (timeout).

integration/pages/BasePage.js
xuesichao: > Need to include GitHub action and an actual test :)

These will be in separate PRs not this one. Vidya told me to break down tasks and PRs.

xuesichao:
> Need to include GitHub action and an actual test :)

These will be in separate PRs not this one. Vidya told me to break down tasks and PRs.

integration/package.json
xuesichao: I don't have enough data to choose a proper timeout. And this may not be even necessary if sauce labs handles the timeout stuff. I will update it later according to the real execution time when running on the Sauce Labs.

integration/pages/BaseTestPage.js
xuesichao: Just Vidya wants me to break down PRs and tasks. So I raised the PR based on the SIM tasks. 😃

integration/pages/BasePage.js
ltrung: But i mean can we just wait for it to be visible which then should include that it exists?

integration/package.json
ltrung: What is the default timeout?

integration/pages/BaseTestPage.js
ltrung: Yeah but each tests should be self-contain and is testable. I am fine with pushing the demo but this needs to have something to test on.

xuesichao: Got it, I will address these comments first and push other commits here later.

integration/package.json
xuesichao: I think it's 30s.

integration/pages/BasePage.js
xuesichao: With `timeout` we actually wait until it exists.

> can elementIsVIsible mean it is located?

No. The `elementIsVisible(WebElement)` need a `WebElement` input. We have to first `find` (locate) this element. This is why we need `await this.driver.wait(until.elementLocated(locator), timeout)` to first check if the element is able to located or not.

xuesichao: The alternative solution could be: In a previous test we already check  the element exists, and then we just need to check if it is displayed.

integration/package.json
xuesichao: Done

integration/package.json
xuesichao: Done

integration/pages/BasePage.js
xuesichao: Changing the `findElement` to `async` breaks current test.

Taking clicking element as an example:
Now
```
// click() is async funciton
await driver.findElement({id: 'my-button'}).click();
```

After changing `findElement` to `async`, we have to update all our code to:
```
const myButton = await driver.findElement({id: 'my-button'});
await myButton.click();
//

or await (await driver.findElement({id: 'my-button'})).click();
```

There is no benefit in doing so.  Thus keep it as it is.

integration/utils/config.js
xuesichao: updated. check the `configure(host)` function in `DriverFactory`

xuesichao:
@ltrung, now the roster test can run both locally or on Sauce Labs.
The only issue now is when running on Sauce Labs, the attendee can not join the meeting due to WebSocket disconnection. I expect this issue to be solved when the GitHub Action hosting the demo URL. Will raise another PR to add GitHub Action and related stuff after thanks giving.

.github/script/get-prev-version
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env node
+
+const exec = require('child_process').execSync;
+
+const versions = exec('npm view amazon-chime-sdk-js versions --json').toString().trim().split("\n");
```
michhyun1, line 5: amazon chime js sdk? Or react library?

integration/pages/BasePage.js
```diff
@@ -0,0 +1,52 @@
...
+      try {
+        return await this.find(locator).isDisplayed();
+      } catch (error) {
+        return false;
+      }
+    }
+  }
+
+  async isDisappeared(locator, timeout) {
+    return await this.driver.wait(until.stalenessOf(locator), timeout);
```
michhyun1, line 44: we have to be careful with these, since we'll base all of oru tests on this base methods. Is "stale" the correct method to use here? Are there other indications that a element is no longer on the page?

integration/pages/BasePage.js
```diff
@@ -0,0 +1,52 @@
...
+  find(locator) {
+    return this.driver.findElement(locator);
+  }
+
+  async findAll(locator) {
+    return await this.driver.findElements(locator);
+  }
+
+  async click(locator) {
+    await this.find(locator).click();
```
michhyun1, line 22: what if the locator does not exist? Do we have to wait for the locator before clicking?

.github/script/get-prev-version
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env node
```
michhyun1, line 1: where do we use this script?

.github/script/get-prev-version
xuesichao: Roster test doesn't need this script. Deleted the file.

.github/script/get-prev-version
xuesichao: Roster test doesn't need this script. Deleted the file.

integration/pages/BasePage.js
xuesichao: The duty of this function is just to click on the element. It is used in the scenario of clicking an element that already exists. This means that the user has to make sure that the element exists before using this method.

From my opinion, the duty you mentioned here is actually the `clickAfterElementAppears`. The flow of this scenario is:
1. The element does not exist at the beginning.
2. The element appears later.
3. Click the element.

for step 1, 2 we can use below method to check:

```js
  async isDisplayed(locator, timeout) {
    if (timeout) {
      await this.driver.wait(until.elementLocated(locator), timeout);
      await this.driver.wait(until.elementIsVisible(this.find(locator)), timeout);
      return true;
    } else {
      try {
        return await this.find(locator).isDisplayed();
      } catch (error) {
        return false;
      }
    }
  }
```

xuesichao: This method doesn't seem necessary, its only purpose is to reduce the amount of code, but the functionality it provides is not complete and has risk to cause trouble.

I will remove `click()` and `type()` functions.

integration/pages/BasePage.js
xuesichao: Yes, this is the official condition for this situation.

>Creates a condition that will wait for the given element to become stale. An element is considered stale once it is removed from the DOM, or a new page has loaded.

integration/pages/BasePage.js
xuesichao: >what if the locator does not exist? Do we have to wait for the locator before clicking?

There will be an `error.NoSuchEementError`.

integration/pages/RosterTestPage.js
```diff
@@ -0,0 +1,155 @@
...
+    return hookState[property];
+  }
+
+  async getLengthOfAttendeeStates() {
+    const hookStates = this.find(ELEMENTS.hookStates);
+    const attendeeStates = await hookStates.findElements(ELEMENTS.attendeeState);
+    return attendeeStates.length;
+  }
+
+  async checkRosterLength(targetLength) {
```
devalevenkatesh, line 65: So for all the `check..` APIs, do we want to name them as `assetRosterLength..`? Currently, it is not clear what this returns or if fails or asserts, if we are going to `asset` inside these `check` methods I think `assertRosterLength` makes it clear IMO. Do you have any early discussion or what do you think?

integration/pages/RosterTestPage.js
```diff
@@ -0,0 +1,155 @@
...
+    assert(isMatched, `Wrong microphone icon when muted is ${muted}`);
+  }
+
+  async checkVideoIcon(attendeeName, videoEnabled) {
+    const isMatched = await this.waitUntil(async () => {
+      const rosterCell = await this.getAttendeeRosterCell(attendeeName);
+      const childElements = await rosterCell.findElements(By.xpath('*'));
+      const videoIcon = childElements[2];
+      const d = await videoIcon.findElement(By.css('path')).getAttribute('d');
+      return videoEnabled === d.startsWith('M19');
```
devalevenkatesh, line 137: What is M19 here? The video icons SVG I guess?
NIt: Request to add a comment.

integration/test/roster_test.js
```diff
@@ -0,0 +1,285 @@
...
+          await localPage.checkVideoIcon(localAttendeeName, false);
+        });
+
+        it('remote attendee has disabled video icon', async function () {
+          await remotePage.checkVideoIcon(localAttendeeName, false);
+        });
+      });
+
+      describe('hook', async function () {
+        it('`videEnabled` of local attendee is `false`', async function () {
```
devalevenkatesh, line 222: nit: `videEnabled` -> `videoEnabled`. Here and in below `it` as well.

integration/pages/RosterTestPage.js
xuesichao: The only way that I can find to check the icon diff...
It's from the data of path define of the `<svg>`.

image

integration/test/roster_test.js
```diff
@@ -0,0 +1,285 @@
...
+          await remotePage.checkAttendeeState(localAttendeeName, 'sharingContent', true);
+        });
+      });
+    });
+
+    describe('stop', async function () {
+      before(async function () {
+        await localPage.toggleContent();
+      });
+      describe('component', async function () {
```
devalevenkatesh, line 258: Remove below?
```
describe('component', async function () {

      });
```

integration/utils/DriverFactory.js
```diff
@@ -0,0 +1,58 @@
...
+class DriverFactory {
+  configure(host) {
+    let builder = new Builder();
+    const sauceLabsURL = 'https://ondemand.saucelabs.com/wd/hub';
+
+    switch (host) {
+      case 'sauce-firefox':
+        builder.usingServer(sauceLabsURL);
+        builder.withCapabilities({
+          ...config.firefoxOptions,
```
devalevenkatesh, line 15: Checking for the order here: If `sauceOptions` have some attributes in common with `firefoxOptions` are we going to override any `firefoxOptions`? OR this is not the case.

integration/test/roster_test.js
```diff
@@ -0,0 +1,285 @@
...
+
+  describe('Attendee Presence', async function () {
+    describe('#length', async function () {
+      describe('component', async function () {
+        it('has length of 2', async function () {
+          await localPage.checkRosterLength(2);
+        });
+      });
+
+      describe('hook', async function () {
```
devalevenkatesh, line 71: Checking if can we use the actual name of the hook in place of 'hook'?

integration/utils/DriverFactory.js
xuesichao: `sauceOptions` stores Sauce Labs related config. `firefoxOptions` and `chromeOptions` store the browser related config. So this is not the case.

```js
  sauceOptions: {
    browserVersion: process.env.BROWSER_VERSION || 'latest',
    platformName: process.env.PLATFORM_NAME || 'Windows 10',
    'sauce:options': {
      username: process.env.SAUCE_USERNAME,
      accessKey: process.env.SAUCE_ACCESS_KEY,
      tunnelIdentifier: process.env.JOB_ID,
      noSSLBumpDomains: 'all',
      extendedDebugging: true,
    }
  },
```
```js
  firefoxOptions: {
    browserName: 'firefox',
    resolution: '1920x1080',
    'moz:firefoxOptions': {
      args: [
        '-start-debugger-server',
        '9222'
      ],
      prefs: {
        'media.navigator.streams.fake': true,
        'media.navigator.permission.disabled': true,
        'media.peerconnection.video.h264_enabled': true,
        'media.webrtc.hw.h264.enabled': true,
        'media.webrtc.platformencoder': true,
        'devtools.chrome.enabled': true,
        'devtools.debugger.prompt-connection': false,
        'devtools.debugger.remote-enabled': true
      },
    },
  },
  chromeOptions: {
    browserName: 'chrome',
    resolution: '1920x1080',
    'goog:chromeOptions': {
      args: [
        '--use-fake-ui-for-media-stream',
        '--use-fake-device-for-media-stream',
      ],
    },
  },
```

integration/script/run-test
```diff
@@ -0,0 +1,64 @@
...
+curr=`pwd`
+
+start_test_demo
+wait_for_demo_to_start 9000
+cd $curr
+
+# Start test and store result in /logs
+case $1 in
+  roster)
+    mocha test/roster_test.js > $curr/logs/$uuid-result 2>&1
```
devalevenkatesh, line 47: For understanding:
- How are these logs helpful to us? Are these just for the temporary runs on the VM or SauceLabs will capture these logs?
- Will adding the test-name help to the `$uuid-result` ($uuid-roster-test-result)?

integration/test/roster_test.js
xuesichao: I forgot adding the tests for component check 😅... Thanks for catching this

integration/script/run-test
xuesichao: Good question.
The mocha runs the test one by one and while printing the test result, like how the unit test does.
But sometimes these test results mess up with the logs of API calls. Cause they share the same console.
For example:
It should be:
```
* test result
* test result
* test result
* test result
* test result
* test result
```

But it could be:
```
* test result
* test result
log of API calling
* test result
log of API calling
* test result
* test result
```

So I disabled this behavior and write the test result in a file to avoid this mess-up. 
At the end of the test, a) check the `passing` and `failing` string in the file, to check the whole testing result, b) print the whole test results in that file at the end of test for easy checking.

integration/pages/RosterTestPage.js
xuesichao: I think `checkXxxxx` should be clear enough. The duty of these APIs is to check if the conditions are met. Their return type is void.

For example, `checkRosterLength(targetLength)` means to check if the length of roster is equal to `targetLength`. It might be clearer to name them as `checkIfRosterLengthEqualsTo(targetLength)`, but it's too redundant in my opinion. Sometimes we need to check if something is not equals to some value as well.

In JS SDK integration testing, we call these steps as `xxxxCheck` and we put them under the `integration/js/checks` folder.

integration/utils/DriverFactory.js
devalevenkatesh: Yup I checked the config later in the review. Thanks for explanation.

integration/script/run-test
devalevenkatesh: Ok so this is just for checking whether a test failed or passed so should be fine it seems.

integration/pages/RosterTestPage.js
devalevenkatesh: Hmm yeah fine to go ahead with it, just wanted to have some discussion over it. But just `check` was not clear enough until that assert line comes. It does not inherently tell what is returned when the failure happens whether it returns a boolean of check passed or failed. The action still remains unclear when one is not familiar with the code and wants to use it. 

Hope this article makes it clear (our discussion is between Assert and Check which is close though :) ): https://artoftesting.com/difference-between-assert-and-verify

integration/pages/RosterTestPage.js
devalevenkatesh: Yup thought so.

integration/script/run-test
xuesichao: Yes, the main goal is to check the test results and make it easier do it, similar to how we did in JS SDK integration test.

> Are these just for the temporary runs on the VM or SauceLabs will capture these logs?
It's the terminal log, so SauceLabs doesn't capture these logs. The GitHub runner will store them I think, but I don't know if we can access it.

> Will adding the test-name help to the $uuid-result ($uuid-roster-test-result)?
It helps just a little bit. Cause anyway you have to find the corresponding uuid to target the exact log. So I don't think there's a need to do it.

integration/pages/BasePage.js
michhyun1: Then we should call this isStale? Dissappeared could mean different things right?

integration/pages/BasePage.js
michhyun1: We should probably implement some wait here - people will use this click function without waiting for the element to be there. Better safe than sorry, right?

integration/script/run-test
devalevenkatesh: Got it thanks!

integration/pages/RosterTestPage.js
xuesichao: Added the comment.

integration/test/roster_test.js
xuesichao: Addressed, thanks for catching it.

integration/pages/BasePage.js
xuesichao: Updated.

integration/test/roster_test.js
xuesichao: Added and tested.

integration/test/roster_test.js
xuesichao: Updated.

integration/pages/BaseTestPage.js
```diff
@@ -0,0 +1,66 @@
...
+  leaveMeetingButton: By.css('[data-testid=\'button\'][label=\'Leave\']'),
+  meetingStatus: By.id('meeting-status'),
+};
+
+class BaseTestPage extends BasePage {
+  constructor(driver) {
+    super(driver);
+  }
+
+  async load(route) {
```
devalevenkatesh, line 19: Can one use `visit` directly over `load`? This looks like an un-necessary extra method on the extended class.

integration/pages/BasePage.js
xuesichao: You are right. I think we should avoid this by design, rather than dealing with it in function logic. We will have other `WebElement` related function in the future, e.g., `select()`, `check()`, ... , it's apparently wrong to handle this kind of wait situation inside each of them.

For now I will delete `click()`, `type` and probably the `isDisplayed()` function and try to find an alternative way later.

integration/test/roster_test.js
```diff
@@ -0,0 +1,312 @@
...
+    await chromeDriverFactory.build('Roster Test - Local', 'chrome');
+    localPage = new RosterTestPage(chromeDriverFactory.driver);
+
+    await firefoxDriverFactory.build('Roster Test - Remote', 'firefox');
+    remotePage = new RosterTestPage(firefoxDriverFactory.driver);
+  });
+
+  afterEach(function () {
+    const state = this.currentTest.state;
+    if (state === "failed") {
```
devalevenkatesh, line 26: nit: Use single quotes for consistency.

integration/pages/BaseTestPage.js
xuesichao: Removed `load()` and use `visit()` instead.

integration/test/roster_test.js
xuesichao: Checked the double quotes again as a whole and fixed them. Should be find now.

integration/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "integration-tests",
+  "version": "1.0.0",
+  "description": "Integration testing for Amazon Chime SDK React Components Library",
+  "scripts": {
+    "test": "npm ci && ./script/run-test",
```
nainkunal933, line 6: @xuesichao I was trying to run the demo and when I run `npm run test` in `integration` folder I am getting this error: 
```
Starting the test demo
Waiting for test demo to start

> test-demo@0.1.0 start
> concurrently "npm run start:client" "npm run start:backend"

sh: concurrently: command not found
```

Seems like since concurrently is not a dependency in integration folder's `package.json` the test command is failing. You might want to add a `npm install` in `run-test` to resolve this.

nainkunal933: Please refer to the `package.json` in JS SDK as that is calling `npm run start` in test demo and test demo's start is installing deps here: https://github.com/aws/amazon-chime-sdk-js/blob/master/integration/js/app/package.json#L10

nainkunal933: There are two options either: `npm install` in `run-test` or updated `package.json`

xuesichao: Thanks for catching this! I added `npm install` in the run-test script, after `cd integration/app/test-demo`

.github/script/get-current-version
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env node
+
+const pjson = require('../../package.json');
+console.log(pjson.version);
```
ltrung, line 4: Is this used?

ltrung: Oh didn't know that you can read from console.log

.github/workflows/roster-integration.yml
```diff
@@ -0,0 +1,62 @@
...
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+      - name: Setup Sauce Connect
+        uses: saucelabs/sauce-connect-action@v1
+        with:
+          username: ${{ secrets.SAUCE_USERNAME }}
+          accessKey: ${{ secrets.SAUCE_ACCESS_KEY }}
+          noSSLBumpDomains: all
+          tunnelIdentifier: ${{ steps.create-job-id.outputs.uuid }}
+      - name: Setup Node.js - 16.x
```
ltrung, line 55: Don't need this step. The default in GitHub should already be 16

integration/package.json
ltrung: Can we add a comment here? Also 2 min seems a bit low, can we change it to 5 min?

package.json
```diff
@@ -35,7 +35,9 @@
     "build-storybook": "build-storybook",
     "jest:integration": "jest -c jest-snapshot.config.js --verbose ./tst/snapshots/*",
     "test:snapshots-path": "jest -c jest-snapshot.config.js --verbose",
-    "test:snapshots": "start-server-and-test start http-get://localhost:9009 jest:integration"
+    "test:snapshots": "start-server-and-test start http-get://localhost:9009 jest:integration",
+    "test:integration": "cd ./integration && npm run test",
```
ltrung, line 39: As discussed yesterday, can we move this to under the integ folder?

integration/utils/Window.js
```diff
@@ -0,0 +1,27 @@
+class Window {
+  constructor(driver) {
+    this.driver = driver;
+  }
+
+  static async getDefaultWindow(driver, name) {
```
ltrung, line 6: Can we comment on what the different between this 2 methods? I also think it is better to have a method to create Window and another one to move to another window.

integration/utils/DriverFactory.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+    return builder;
+  }
+
+  async build(testName, host = this.config.host) {
+    this.host = host;
+    this.driver = await this.configure(host).build();
+    const { id_ } = await this.driver.getSession();
+    this.sessionId = id_;
+    this.driver.executeScript('sauce:job-name=' + testName);
```
ltrung, line 41: Do we have to check for `    if (this.host.startsWith('sauce')) {` here?

integration/pages/BasePage.js
```diff
@@ -0,0 +1,44 @@
...
+    } else {
+      try {
+        return await this.find(locator).isDisplayed();
+      } catch (error) {
+        return false;
+      }
+    }
+  }
+
+  async isStale(locator, timeout) {
```
ltrung, line 35: Do we use this anywhere?

integration/pages/BasePage.js
```diff
@@ -0,0 +1,44 @@
...
+
+  find(locator) {
+    return this.driver.findElement(locator);
+  }
+
+  async findAll(locator) {
+    return await this.driver.findElements(locator);
+  }
+
+  async isDisplayed(locator, timeout) {
```
ltrung, line 21: Seem like we always pass timeout when using this function? Do we have a use case for the else part?
Also can we just set 5000 as the default value for timeout here instead of needing to always passing it in?

integration/script/run-test
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash -i
+
+function start_test_demo {
```
ltrung, line 3: Can we use node script instead of bash :D

integration/utils/DriverFactory.js
xuesichao: Yes, so the test can run either locally or on Sauce Labs.
The video URL only exists when the test running on Sauce Labs.

In `configure(host)` function, if host starts with 'sauce', e.g. 'sauce-chrome'. The test will run on Sauce Labs using latest Chrome. Other wise if hose is `chrome` it will run locally.

integration/test/roster_test.js
```diff
@@ -0,0 +1,310 @@
...
+  let localWindow;
+  let remoteWindow;
+  let failureCount = 0;
+
+  const meetingId = uuidv4();
+  const localAttendeeName = 'local-attendee';
+  const remoteAttendeeName = 'remote-attendee';
+
+  before(async function () {
+    await driverFactory.build('Roster Test', 'sauce-chrome');
```
ltrung, line 19: So we only test with Chrome right now right? How do i run this locally with local grid?
Can we make this a parameter?

xuesichao: It is a parameter now:
```js
async build(testName, host = this.config.host) {
  ...
}
```
So if you want to run the test locally, just update it to:

```js
    await driverFactory.build('Roster Test', 'chrome');
````

It would be great to have a `enum` here, but it's `.js` file, so can't do it

Comment (xuesichao) [reply integration/test/roster_test.js:19]: >So we only test with Chrome right now right? 

We can test with Firefox as well:
```js
await driverFactory.build('Roster Test', 'firefox'); // Locally
await driverFactory.build('Roster Test', 'sauces-firefox'); // On Sauce Labs
```

integration/script/run-test
xuesichao: I prefer to do this later. 

Since the JS SDK using the bash, I can refer to that one. If changing to node script, I will need to spend more time to learn and update, since no reference I assume.

.github/workflows/roster-integration.yml
xuesichao: Removed.

package.json
xuesichao: Updated.

integration/pages/BasePage.js
xuesichao: Deleted.

integration/utils/Window.js
xuesichao: >I also think it is better to have a method to create Window and another one to move to another window.

1. `createNewWindow(driver, name)` creates a new window. 
2. The `run(commands)` function automatically switches the window to the target one before executing the commands. So no need to have a method to switch the window.
  ```js
    async run(commands) {
      await this.driver.switchTo().window(this.handle);
      await commands();
    }
  ```

integration/pages/BasePage.js
xuesichao: Refactored `isDisplayed` function a little bit. 
And set the default `timeout` to 5000 for all wait until functions.

```js
  // Check if the element is able to be found
  async isFound(locator, timeout = 5000) {
    return await this.driver.wait(until.elementLocated(locator), timeout);
  }

  // Check if the element is `displayed`
  async isDisplayed(locator, timeout = 5000) {
    await this.driver.wait(until.elementIsVisible(this.find(locator)), timeout);
  }
```

So to check if an element appears and displayed, for example:
```js
  async checkIfTestAppHasLoaded() {
    const isFound = await this.isFound(ELEMENTS.appHeader);
    assert(isFound, 'The header of Test App was not found, Test App was not loaded successfully');

    const isDisplayed = await this.isDisplayed(ELEMENTS.appHeader);
    assert(isDisplayed, 'The header of Test App was not displayed, Test App was not loaded successfully');
  }
```

integration/package.json
xuesichao: Updated.

integration/pages/BasePage.js
xuesichao: Deleted `click()` and `type()` funtion. Refactored `isDisplayed` function a little bit. 
And set the default `timeout` to 5000 for all wait until functions.

```js
  // Check if the element is able to be found
  async isFound(locator, timeout = 5000) {
    return await this.driver.wait(until.elementLocated(locator), timeout);
  }

  // Check if the element is `displayed`
  async isDisplayed(locator, timeout = 5000) {
    return await this.driver.wait(until.elementIsVisible(this.find(locator)), timeout);
  }
```

So to check if an element appears and displayed, for example:
```js
  async checkIfTestAppHasLoaded() {
    const isFound = await this.isFound(ELEMENTS.appHeader);
    assert(isFound, 'The header of Test App was not found, Test App was not loaded successfully');

    const isDisplayed = await this.isDisplayed(ELEMENTS.appHeader);
    assert(isDisplayed, 'The header of Test App was not displayed, Test App was not loaded successfully');
  }
```

integration/test/roster_test.js
xuesichao: >If so, you may need to update the screenshot in the description.

Updated.

xuesichao: > How do i run this locally with local grid?

Simply remove `sauce-`
For example, to test on Sauce Labs
```
await driverFactory.build('Roster Test', 'sauce-firefox'); // On Sauce Labs with Firefox
// Or
await driverFactory.build('Roster Test', 'sauce-chrome'); // On Sauce Labs with Firefox
```

To test locally:

````
await driverFactory.build('Roster Test', 'firefox'); // Locally with Firefox
// Or
await driverFactory.build('Roster Test', 'chrome'); // Locally with Firefox
```

ltrung: Thats what I meant. It is not good though. We need to have a way to do this in common line and don't need to change code.

xuesichao: > Thats what I meant. It is not good though. We need to have a way to do this in common line and don't need to change code.

Got it. Could you explain this a little bit more? How should it works like finally? Like how should it works, and what is the use cases.

e.g.
Should we run command once and then test is running against all platforms and browsers. Or we could run the command one by one to do that?

xuesichao: Can we merge this first and improve it in another PR, since it's an improvement.

ltrung: I disagree. It is a basic requirement and we should just do it :)

integration/script/run-test
ltrung: Node script is easier to do and which we all know well. I don't think we should create any new bash script.

xuesichao: Updated.

integration/test/roster_test.js
xuesichao: Updated. 
In `/integration` folder, simply call `npm run test:roster` to trigger roster integration test.
By default the `host` is local chrome browser, the `platform` is `Windows 10`. The `browser version` is `latest`.

To specify the `host`, `platform` and `browser version`. Just call the commend line with named argument.
```sh
npm run test:roster -- --host='sauce-chrome' --browser-version='latest' --platform-name='Mac'
```

integration/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "scripts": {
+    "test": "./script/run-test",
+    "test:fast": "mocha ./test/*"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "child_process": "^1.0.2",
+    "chromedriver": "^96.0.0",
+    "fs-extra": "^10.0.0",
+    "geckodriver": "^2.0.4",
```
ltrung, line 14: What if we want to run Safari locally? Is there a webkit driver?

integration/script/run-test
```diff
@@ -0,0 +1,211 @@
+#!/usr/bin/env node
+
+const { argv, env, kill, exit } = require('process');
+const { spawnSync, spawn } = require('child_process');
+const args = require('minimist')(argv.slice(2));
+const path = require('path');
```
ltrung, line 6: Do we need this? From the npm description `This is an exact copy of the NodeJS ’path’ module published to the NPM registry.` ??

integration/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "integration-tests",
+  "version": "1.0.0",
+  "description": "Integration testing for Amazon Chime SDK React Components Library",
+  "scripts": {
+    "test": "./script/run-test",
+    "test:fast": "mocha ./test/*"
```
ltrung, line 7: what is this option for? Doesn't seem it is faster than the default option?

xuesichao: The `test` script starts the demo, run the test, check the result and terminate the demo.
The `test:fast` just runs the test directly, it's mostly for development and debugging.
So when you develop/debug the test, you just need to start the test demo your self and run this script.

ltrung: but does just run all tests?

xuesichao: It only runs all integration tests. Since this `package.json` file is in the integration folder, I think there's no need to name them as `test:integration` or `test:integration-fast`. There should be enough context here, I assume.

integration/package.json
xuesichao: As we discussed earlier, we don't need to test Safari now, so I didn't add it. Do you want me to add Safari? It's simple.

integration/script/run-test
xuesichao: We could `path` from `package.json`, as well as the `child_process`. I just checked the Node.js doc, we could just `require` them.

ltrung: Yeah then we could remove this dependency?

integration/test/roster_test.js
ltrung: Nice work

integration/package.json
ltrung: We don't need to enable it in GitHub action but we should add it so we can run it locally for testing and stuff :D

.github/workflows/roster-integration.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Setup Sauce Connect
+        uses: saucelabs/sauce-connect-action@v1
+        with:
+          username: ${{ secrets.SAUCE_USERNAME }}
+          accessKey: ${{ secrets.SAUCE_ACCESS_KEY }}
+          noSSLBumpDomains: all
+          tunnelIdentifier: ${{ steps.create-job-id.outputs.uuid }}
+      - name: Clean Install
+        run: npm ci
+      - name: Run Roster Integration Test
```
ltrung, line 57: nit: Can we say the browser name here `Run Roster Integration Test`

integration/script/run-test
xuesichao: Yes, I deleted them in my local branch, will update the PR soon. Thanks for catching this.

.github/workflows/roster-integration.yml
xuesichao: Sure, so the browser used in this GitHub Action is Chrome, should we change it to `Run Roster Integration Test - Chrome`? Sounds good?

xuesichao: Updated

integration/script/run-test
xuesichao: Updated

integration/package.json
xuesichao: Add the support for Safari

ltrung (approved):

ltrung (approved):

Issue #668, opened 2021-11-16
Overriding `activeSpeakerPolicy` requires undocumented steps


kaizau:
### What are you trying to do?

Override `ActiveSpeakerPolicy` with a custom instance.

### How can the documentation be improved to help your use case?

The docs from [MeetingManager > activeSpeakerPolicy](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/docs/sdk-providers-meetingprovider--page) are incomplete.

In addition to passing`meetingConfig` to `<MeetingProvider>`, it's necessary to also pass it to `meetingManager.join({meetingManagerConfig: meetingConfig})`.

Otherwise the meeting continues to use the `DefaultActiveSpeakerPolicy`.

### What documentation have you looked at so far?

- https://aws.github.io/amazon-chime-sdk-js/interfaces/activespeakerpolicy.html
- https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/docs/sdk-providers-meetingprovider--page

Ultimately, I was only about to figure this out by digging through the source.

Inserting a breakpoint at [`setupActiveSpeakerDetection()`](https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/providers/MeetingProvider/MeetingManager.ts#L491) revealed that `this.meetingManagerConfig.activeSpeakerPolicy` was still the default, despite passing a custom override via `<MeetingProvider>`.

I'm not sure if this is intended, so this one is arguably a bug. Either way, following the example given didn't work.

devalevenkatesh:
@kaizau thanks for reporting this issue and sorry for the delay, we missed adding props through the MeetingProvider, added those now and you should be able to use the custom active speaker policy. 

Please note that if you pass a meetingManagerConfig prop through the meetingManager.join then that overrides any existing config passed through the MeetingProvider. That is already documented. 

We have merged the PR now and will be available in next release.

jing-chen1:
Closing the issue since the fix is in released version.

PR #683, opened 2021-11-24
Add prop to provide custom active speaker policy through MeetingProvider prop

devalevenkatesh:
**Issue #:** 
- #668 

As pointed the documentation is not pointing the correct way to use it, hence we have two approaches:
1. Improve on the documentation and the builders will have to provide the `activeSpeakerPolicy` with MeetingManager.join()
2. Allow builders to pass `activeSpeakerPolicy` through MeetingProvider props.
- This approach allows passing `activeSpeakerPolicy`  through both via `join` and `MeetingProvider` props and the documentation does not need to be updated.

**Description of changes:**
- Follows second approach discussed above.
- Storybook shows the new prop on the MeetingProvider prop.
- Along with this, also added the `videoUplinkBandwidthPolicy` prop to `MeetingProvider` props so it gets passed down to `MeetingManager`.

**Testing**
Yes.

1. As per documentation created a custom Speaker policy in demo meeting app and tested in context whether we see the custom one and not the Default one.
2. Also tested if nothing is passed does the MeetingManager still uses the `DefaultActiveSpeakerPolicy` then.
3. Tested using breakpoints as well as said in the issue.

No, the props changes are auto-picked by storybook hence they auto-update.


src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -222,7 +222,7 @@ export class MeetingManager implements AudioVideoObserver {
     meetingManagerConfig,
   }: MeetingJoinData) {
     if (meetingManagerConfig) {
-      this.meetingManagerConfig = meetingManagerConfig;
+      this.meetingManagerConfig = {...this.meetingManagerConfig, ...meetingManagerConfig};
```
devalevenkatesh, line 225: This is necessary, cause if MeetingProvider creates a MeetingManager with few props and later at join time if builders provide a custom meetingManagerConfig we override the existing one, rather override the attributes that are passed and not all.

xuesichao: Just want to confirm. For example:
```js
this.meetingManagerConfig = {
    a: 0,
    b: 0,
}

meetingManagerConfig = {
    b: 1,
    c: 1
};

this.meetingManagerConfig = { ...this.meetingManagerConfig, ...meetingManagerConfig };

console.log(this.meetingManagerConfig);
// {
//   "a": 0, // case 1
//   "b": 1, // case 2
//   "c": 1, // case 3
// } 
```
So what we want is to:
* case 1: keep the existing properties which `meetingManagerConfig` doesn't have
* case 2: override the existing properties which `meetingManagerConfig` does have
* case 3: add the new properties which`meetingManagerConfig` dose have

devalevenkatesh: Yes, I think that is better over the existing direct replace approach, what do you think? I was hesitant on doing so because its easier to just say that the MeetingManagerConfig from join overrides the existing MeetingProvider props but then why would we need the MeetingProvider props in first place.

devalevenkatesh: Ok, so I checked few last PRs and we said if you are passing the props through it will just override the MeetingManagerConfig:
Source: Check "meetingManager.join()` under documentation [here](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/sdk-providers-meetingmanager--page).
```
// Set the meeting config. This will override the props from MeetingProvider.
  meetingManagerConfig: MeetingManagerConfig;
```

Lets follow that and just update the documentation should be enough. I will re-work this.

xuesichao: I think it's a breaking change. Because it is not guaranteed to override the whole `meetingManagerConfig` anymore, unless the builders overrides all the properties.

devalevenkatesh: As discussed will not be doing so.

ltrung (approved):

PR #689, opened 2021-12-01
Add Amazon Chime Echo Reduction for React SDK

wchmzn:
**Issue #:** 

**Description of changes:**
Add Amazon Chime Echo Reduction for React SDK

**Testing**

Testing steps:
- Enable Web Audio. Enable Echo Reduction. Join the meeting
- Select one input audio device, and then click on "Enable voice focus". 
- Expect: ns_es model get download and exactly once.
- demo PR: https://github.com/aws-samples/amazon-chime-sdk/pull/109
Echo Reduction On:
Screen Shot 2021-12-01 at 9 23 56 PM
Echo Reduction Off:
Screen Shot 2021-12-01 at 9 24 13 PM
Default Voice Focus On:
Screen Shot 2021-12-06 at 9 51 23 PM


CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Add `activeSpeakerPolicy` and `videoUplinkBandwidthPolicy` to `MeetingProvider` props.
-
+- Add `createMeetingResponse` to `VoiceFocusProvider` props.
```
xuesichao, line 16: What you really did is adding Amazon Chime Echo Reduction, right?

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -110,6 +113,11 @@ const AudioInputVFControl: React.FC<Props> = ({
             setIsLoading(true);
             const receivedDevice = audioInputSelectionToDevice(device.deviceId);
             const currentDevice = await addVoiceFocus(receivedDevice);
+
+            console.info('ObserveMeetingAudio:', audioVideo);
+            await (currentDevice as VoiceFocusTransformDevice).observeMeetingAudio(audioVideo as AudioVideoFacade);
+            console.info('ObserveMeetingAudio:', audioVideo);
```
xuesichao, line 119: Do we need to keep these two `console.info`?

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -164,6 +172,10 @@ const AudioInputVFControl: React.FC<Props> = ({
         if (typeof device === 'string') {
           const currentDevice = audioInputSelectionToDevice(device);
           current = await addVoiceFocus(currentDevice);
+
+          // console.info('ObserveMeetingAudio:', audioVideo);
+          // await (current as VoiceFocusTransformDevice).observeMeetingAudio(audioVideo as AudioVideoFacade);
+          // console.info('ObserveMeetingAudio:', audioVideo);
```
xuesichao, line 178: Remove them if they are useless.

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -2,14 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import {
+  // AudioMixObserver,
+  // AudioVideoFacade,
```
xuesichao, line 6: Same here.

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -21,6 +22,7 @@ interface Props {
    * For more info, you can go to https://aws.github.io/amazon-chime-sdk-js/interfaces/voicefocusdeviceoptions.html
    */
   options?: VoiceFocusDeviceOptions;
+  createMeetingResponse?: any;
```
xuesichao, line 25: Could we avoid `any`?

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -32,18 +34,25 @@ interface VoiceFocusState {
 
 const VoiceFocusContext = createContext<VoiceFocusState | null>(null);
 
-const VoiceFocusProvider: React.FC<Props> = ({ spec, options, children }) => {
+const VoiceFocusProvider: React.FC<Props> = ({ 
+  spec, 
+  options, 
+  createMeetingResponse,
+  children
```
xuesichao, line 41: Add trailing comma here.

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -32,18 +34,25 @@ interface VoiceFocusState {
...
+}) => {
   const [isVoiceFocusSupported, setIsVoiceFocusSupported] = useState<
     boolean | undefined
   >(undefined);
   const [voiceFocusDevice, setVoiceFocusDevice] =
     useState<VoiceFocusTransformDevice | null>(null);
   const [voiceFocusTransformer, setVoiceFocusTransformer] =
     useState<VoiceFocusDeviceTransformer | null>(null);
+    const [initialized, setInitialized] =
+    useState<Boolean>(false);
```
xuesichao, line 51: Could we make it one line?

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -32,18 +34,25 @@ interface VoiceFocusState {
...
 
   // Make sure that minor changes to the spec don't result in recomputation:
   // Any value of `{}` and undefined are all considered the same.
-  const vfSpec = useMemoCompare(
+  let vfSpec = useMemoCompare(
```
xuesichao, line 55: Why do we need to change it to `let` instead of `const`?

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -132,9 +141,10 @@ const VoiceFocusProvider: React.FC<Props> = ({ spec, options, children }) => {
   async function createVoiceFocusDeviceTransformer(
     spec: VoiceFocusSpec | undefined,
     options: VoiceFocusDeviceOptions | undefined,
+    createMeetingResponse: any | undefined,
```
xuesichao, line 144: Could we avoid `any` here?

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -132,9 +141,10 @@ const VoiceFocusProvider: React.FC<Props> = ({ spec, options, children }) => {
...
     canceled: () => boolean
   ): Promise<VoiceFocusDeviceTransformer> {
-    const fetch = VoiceFocusDeviceTransformer.create(spec, options);
+    const fetch = VoiceFocusDeviceTransformer.create(spec, options, undefined, createMeetingResponse);
```
xuesichao, line 147: Could we change it to:
```js
const fetch = VoiceFocusDeviceTransformer.create(
  spec,
  options,
  undefined,
  createMeetingResponse
);

```

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -165,21 +175,25 @@ const VoiceFocusProvider: React.FC<Props> = ({ spec, options, children }) => {
   async function initVoiceFocus(
     vfSpec: VoiceFocusSpec | undefined,
     options: VoiceFocusDeviceOptions | undefined,
+    createMeetingResponse: any | undefined,
```
xuesichao, line 178: Could we avoid `any` here?

xuesichao (changes requested):
1. Since you update the `package.json`, I think you need to include `package-lock.json` as well.
2. You need to update the storybook of `VoiceFocusProvider` and `AudioInputVFControl` for your change.

Should we also mention the difference between Echo Reduction and normal Voice Focus? If we need to charge builder for Echo Reduction, could we add the link to the billing rules page?

CHANGELOG.md
richnew10: It's fine for a CHANGELOG entry to be factual, but it should also be useful.

```
- Allow builders to supply the response from a `CreateMeeting` or `CreateMeetingWithAttendees`
  call when adding a `VoiceFocusProvider` to the component tree. This enables optional features
  like Amazon Chime Echo Reduction to be added to devices when turning on Amazon Chime
  Voice Focus.
```

package.json
```diff
@@ -79,7 +79,7 @@
     "@typescript-eslint/eslint-plugin": "^4.33.0",
     "@typescript-eslint/parser": "^4.33.0",
     "add": "^2.0.6",
-    "amazon-chime-sdk-js": "^2.21.1",
+    "amazon-chime-sdk-js": "^2.23.0",
```
richnew10, line 82: Please commit `package-lock.json`.

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
richnew10: They should be removed.

src/providers/VoiceFocusProvider/index.tsx
richnew10: We have a type for this, and we should use it.

src/providers/VoiceFocusProvider/index.tsx
richnew10: Lines up to 120 characters are OK.

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -165,21 +175,25 @@ const VoiceFocusProvider: React.FC<Props> = ({ spec, options, children }) => {
...
+    createVoiceFocusDeviceTransformer(vfSpec, options, createMeetingResponse, canceled);
   }
 
   useEffect(() => {
     let canceled = false;
-    initVoiceFocus(vfSpec, options, () => canceled);
+    if (createMeetingResponse && !initialized) {
+      initVoiceFocus(vfSpec, options, createMeetingResponse, () => canceled);
+      setInitialized(true);
+    }
```
richnew10, line 192: But `initVoiceFocus` can fail, and is asynchronous. This logic doesn't seem right. Try it: turn your wifi off after loading the page but before mounting the component. You'll end up with `setInitialized(true)` being called, but the transformer is not initialized.

richnew10 (changes requested):
1. Does this require changes to the demo app? I imagine it does — you need to pass in the `CreateMeetingResponse`! Those need to be staged alongside this change.
2. Testing this involves making sure that echo reduction works, not just that the correct model is loaded. You also need to test the negative case: that you get original Voice Focus when echo reduction isn't enabled. And you need to construct and test the volitional negative case: when the builder turns on echo reduction for the meeting, but chooses not to use it in their application. Please provide evidence of testing these things.

CHANGELOG.md
wchmzn: Thanks. updated.

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
wchmzn: removed.

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
wchmzn: removed

src/providers/VoiceFocusProvider/index.tsx
wchmzn: removed

src/providers/VoiceFocusProvider/index.tsx
wchmzn: The type is removed. Same for JS SDK. https://code.amazon.com/packages/ChimeNoiseSuppressionWeb/commits/15e3f7ad06598e5beb9b98c06aae046647a9c2eb#
https://github.com/aws/amazon-chime-sdk-js/blob/147e7bb8f0b38e13410a50956c921d54cd872f44/src/voicefocus/VoiceFocusDeviceTransformer.ts#L138

src/providers/VoiceFocusProvider/index.tsx
wchmzn: added

src/providers/VoiceFocusProvider/index.tsx
wchmzn: sure

src/providers/VoiceFocusProvider/index.tsx
wchmzn: change to const

src/providers/VoiceFocusProvider/index.tsx
wchmzn: Keep same with JS SDK. 
https://github.com/aws/amazon-chime-sdk-js/blob/147e7bb8f0b38e13410a50956c921d54cd872f44/src/voicefocus/VoiceFocusDeviceTransformer.ts#L138

src/providers/VoiceFocusProvider/index.tsx
wchmzn: done

src/providers/VoiceFocusProvider/index.tsx
wchmzn: Keep same with JS SDK.
https://github.com/aws/amazon-chime-sdk-js/blob/147e7bb8f0b38e13410a50956c921d54cd872f44/src/voicefocus/VoiceFocusDeviceTransformer.ts#L138

src/providers/VoiceFocusProvider/index.tsx
wchmzn: `initVoiceFocus` won't be called until `createMeetingResponse` is not null. I tried to replace boolean `initialized` with checking existence of transformer, due to `initVoiceFocus` is async, so I see this useEffect is called multiple times before transformer is actually created, which results in different complexity models getting downloaded multiple times.

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -110,6 +113,9 @@ const AudioInputVFControl: React.FC<Props> = ({
             setIsLoading(true);
             const receivedDevice = audioInputSelectionToDevice(device.deviceId);
             const currentDevice = await addVoiceFocus(receivedDevice);
+
+            await (currentDevice as VoiceFocusTransformDevice).observeMeetingAudio(audioVideo as AudioVideoFacade);
```
devalevenkatesh, line 117: From the logic I see, you want to call `observeMeetingAudio` on the selected `currentDevice` if it is `VoiceFocusTransformDevice`.

This component code currently is already subscribing to change in audio input transform device. so the moment it changes the `device` state is also updated. In React, its a good practice to separate the concern into its own hooks. [Reference](https://reactjs.org/docs/hooks-effect.html#tip-use-multiple-effects-to-separate-concerns).

Based on this, I think it would be more cleaner to understand below code and you wont have to duplicate it in `onVFCheckboxChange` and here.

```tsx
useEffect(() => {
  if (!audioVideo) {
    return;
  }
  // This ensures that the selected current device is an Amazon Voice Focus transform device.
  if (isVoiceFocusEnabled) {
    await (currentDevice as VoiceFocusTransformDevice).observeMeetingAudio(audioVideo as AudioVideoFacade);
  }
}, [audioVideo, isVoiceFocusEnabled, device]);
```

This hook can go after the third `useEffect` in the existing code I think since in React the order of the hooks also matters.

Request you to update this to implement this using an extra `useEffect` hook. Tagging @jing-chen1 more help on this change.

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -32,14 +32,20 @@ interface VoiceFocusState {
 
 const VoiceFocusContext = createContext<VoiceFocusState | null>(null);
 
-const VoiceFocusProvider: React.FC<Props> = ({ spec, options, children }) => {
+const VoiceFocusProvider: React.FC<Props> = ({ 
+  spec, 
+  options, 
+  createMeetingResponse,
```
devalevenkatesh, line 38: Will the VF provider now accepting `createMeetingResponse` can be re-used with multiple meetings? We separated out the VF provider because this can used for multiple meetings, can you once check with Richard on this, tagging @jing-chen1 for better context on this.

wchmzn: @richnew10 does default voice focus support multiple meetings? how does that work? as far as I know, there was no mixing logic for multiple microphone streams.

richnew10: It's separate because:

1. You typically want to initialize VF prior to joining a meeting, and perhaps even before knowing which meeting you're joining. With echo reduction you can no longer do the latter, but you can do the former (just grab the `CreateMeetingResponse` from the server first — it's not the same as joining).
2. You might want to use the same microphone input as input to multiple meetings without running noise suppression multiple times. E.g., COVID testing supervision where a doctor will read instructions to six participants in six 1:1 meetings. This will work with echo reduction: request ER for all meetings, supply the `CreateMeetingResponse` of any of them when creating the device, and [observe the meeting audio for all meetings](https://github.com/aws/amazon-chime-sdk-js/blob/master/guides/09_Amazon_Voice_Focus.md#2-enabling-echo-reduction-at-the-client-level), which will use all of the output streams mixed together as the echo signal.

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -165,21 +177,25 @@ const VoiceFocusProvider: React.FC<Props> = ({ spec, options, children }) => {
...
     setVoiceFocusTransformer(null);
     setVoiceFocusDevice(null);
-    createVoiceFocusDeviceTransformer(vfSpec, options, canceled);
+    createVoiceFocusDeviceTransformer(vfSpec, options, createMeetingResponse, canceled);
   }
 
   useEffect(() => {
     let canceled = false;
-    initVoiceFocus(vfSpec, options, () => canceled);
+    if (createMeetingResponse && !initialized) {
```
devalevenkatesh, line 191: What if the builders want to use the VF without ER? Say if `createMeetingResponse` is not passed will the VF work?

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
devalevenkatesh: Is there a way to `unobserveMeetingAudio`? Suggesting removing the observer once the selected device is not an VoiceFocusTransformDevice anymore. Else, multiple observers are going to get added here.

src/providers/VoiceFocusProvider/index.tsx
wchmzn: figured out the root cause here. reason why this useeffect is called multiple times is that vfSpec is not deep compared. Adding `JSON.stringify(prev) === JSON.stringify(next)` solved the problem. Will removed `setInitialized(true)`.

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
wchmzn: Observer is [Set](https://github.com/aws/amazon-chime-sdk-js/blob/147e7bb8f0b38e13410a50956c921d54cd872f44/src/audiomixcontroller/DefaultAudioMixController.ts#L22), which has dedup logic itself. same for [addFarEndStream](https://github.com/aws/amazon-chime-sdk-js/blob/147e7bb8f0b38e13410a50956c921d54cd872f44/src/voicefocus/VoiceFocusTransformDevice.ts#L259).

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -88,7 +93,6 @@ const VoiceFocusProvider: React.FC<Props> = ({ spec, options, children }) => {
     }
 
     if (!isVoiceFocusSupported) {
-      console.debug('Not supported, not creating device.');
```
devalevenkatesh, line 91: Any reason to remove?

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -99,6 +102,18 @@ const AudioInputVFControl: React.FC<Props> = ({
     }
   }, [device]);
 
+  useEffect(() => {
+    if (!audioVideo) {
+      return;
+    }
+    if (device instanceof VoiceFocusTransformDevice) {
+      if (isVoiceFocusEnabled) {
+        console.info('VoiceFocusTransformDevice observeMeetingAudio.');
```
devalevenkatesh, line 111: You probably want to add which device you are adding the meeting audio observer to help while debugging.

```
console.info(`Add observeMeetingAudio on ${JSON.stringify(device)} VoiceFocusTransformDevice.`);
```

devalevenkatesh (changes requested):
Few questions:

1. Do we want to support an additional option on the AudioInputVFControl which enables-disables ER on the meeting? (This PR logic implicitly does the enabling but not disabling as far as I am understanding it)
2. What way we are suggesting builders to disable ER?
-  I do not know what use-case it would be may be CPU consumption etc, but want to know already discussed.
- This will cover [disabling-echo-reduction-from-your-application](https://github.com/aws/amazon-chime-sdk-js/blob/master/guides/09_Amazon_Voice_Focus.md#disabling-echo-reduction-from-your-application).

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
devalevenkatesh: Got it thanks for the explanation! Then the multiple observers wont be an issue, but I posted a general comment on how we plan to disable it? Will the feature support enabling-disabling ER when VF is enabled-disabled?

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -99,6 +102,18 @@ const AudioInputVFControl: React.FC<Props> = ({
     }
   }, [device]);
 
+  useEffect(() => {
+    if (!audioVideo) {
+      return;
+    }
+    if (device instanceof VoiceFocusTransformDevice) {
+      if (isVoiceFocusEnabled) {
```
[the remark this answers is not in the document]
wchmzn: no this does not work. you can try it. exception will pop up.

src/providers/VoiceFocusProvider/index.tsx
```diff
@@ -165,21 +175,24 @@ const VoiceFocusProvider: React.FC<Props> = ({ spec, options, children }) => {
...
     setVoiceFocusTransformer(null);
     setVoiceFocusDevice(null);
-    createVoiceFocusDeviceTransformer(vfSpec, options, canceled);
+    createVoiceFocusDeviceTransformer(vfSpec, options, createMeetingResponse, canceled);
   }
 
   useEffect(() => {
     let canceled = false;
-    initVoiceFocus(vfSpec, options, () => canceled);
+    if (createMeetingResponse) {
```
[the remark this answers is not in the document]
wchmzn: no will only initVoiceFocus when there is createMeetingResponse, because only when createMeetingResponse is defined we could see if the transformer is VF or RES.

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
michhyun1: doesnt the meetingManger have a logger reference instead of using console.info?

src/providers/VoiceFocusProvider/index.tsx
michhyun1: lets not post the code.amazon.com links - those are internal.

src/providers/VoiceFocusProvider/index.tsx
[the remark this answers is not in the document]
michhyun1: So if we have no createMeetingResponse, then we dont need to initVoiceFocus?

src/providers/VoiceFocusProvider/index.tsx
devalevenkatesh: Thanks for explanation @richnew10 ! We just have to test and document the behavior for multiple meetings to help builders understand it better. Took some time to imagine the scenario here :)

src/providers/VoiceFocusProvider/index.tsx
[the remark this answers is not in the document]
devalevenkatesh: So, this means if the builders wont provide the `createMeetingResponse` the voice focus wont be supported which will be a backward incompatible change. We want to support both:
1. Builders should be able to continue VF use without ER. If this is not supported with this logic then it will be a breaking change. Builders will upgrade to the new React component library version and their app will break if they are using VFProvider cause, now the init needs `createMeetingResponse`.
2. Builder should be able to use ER with VF.

PR #714, opened 2022-01-12
Add new release script

ltrung:
**Description of changes:**
Change to get version from git in `Versioning.ts` like in JS SDK.

Replace our current release script:
- Include new options for handling major and pre-build versions (beta).
- Include new option for redeploy demo in release or hotfix branch so that we could do incorporate fix and redeploy test demo without rerun the entire script.
- Change release to bump version at the beginning of any new version.

Also update our publish script to tag pre-release (beta) version.

**Testing:**
- Test the release script by manually executing.

*Can these tested using a demo application? Please provide reproducible step-by-step instructions.*
No

**Checklist:**

1. Have you successfully run `npm run build:release` locally? N/A

2. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? No

3. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved? No

### DIFF
--- a/.github/script/get-pre-release-name.js
+++ b/.github/script/get-pre-release-name.js
@@ -1,0 +1,9 @@
+#!/usr/bin/env node
+
+const version = require('../../package.json').version;
+const preRelease = version.split('-');
+if (preRelease[1]) {
+  console.log(preRelease[1].split('.')[0]);
+} else {
+  console.log('');
+}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -2,7 +2,9 @@ name: Deploy Components Demo App Workflow
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      - 'release-**.x'
 
 jobs:
   deploy:
@@ -30,20 +32,41 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 16
+      - name: Get branch name
+        id: branch_name
+        run: |
+          branch=${{github.ref}}
+          branch=${branch#refs/heads/}
+          echo $branch
+          echo ::set-output name=branch_name::$branch
       - name: Checkout amazon-chime-sdk repo
         uses: actions/checkout@v2
         with:
           repository: 'aws-samples/amazon-chime-sdk'
           path: amazon-chime-sdk
+          ref: ${{steps.branch_name.outputs.branch_name}}
       - name: Install latest version of React library and JS SDK in meeting demo
         run: |
           npm install
           node ./scripts/update-sdk-version.js
       - name: Deploy the Serverless Demo App
+        if: ${{ steps.branch_name.outputs.branch_name == 'main' }}
         run: |
           cd $GITHUB_WORKSPACE/amazon-chime-sdk/apps/meeting/serverless
           node ./deploy.js -r us-east-1 -b chime-sdk-components-demo-canary -s chime-sdk-components-demo-canary -l
-      ##- name: Publish Storybook Documentation
-        ##run: npm run deploy-storybook -- --ci
-        ##env:
-          ##GH_TOKEN: chime-sdk-js-bot:${{ secrets.GITHUBPAGES}}
+      - name: Deploy the Serverless Demo App for Previous Major Version
+        if: ${{ steps.branch_name.outputs.branch_name != 'main' }}
+        run: |
+          cd $GITHUB_WORKSPACE/amazon-chime-sdk/apps/meeting/serverless
+          node ./deploy.js -r us-east-1 -b chime-sdk-components-demo-canary-prev -s chime-sdk-components-demo-canary-prev -l
+      - name: Check if need to deploy storybook
+        id: need_deploy
+        run: |
+          result=$(.github/script/need-deploy-storybook.js)
+          echo $result
+          echo ::set-output name=need_deploy::$result
+      - name: Publish Storybook Documentation
+        if: startsWith(steps.need_deploy.outputs.need_deploy, 'Deploy storybook')
+        run: npm run deploy-storybook -- --ci
+        env:
+          GH_TOKEN: chime-sdk-js-bot:${{ secrets.GITHUBPAGES}}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -21,7 +21,19 @@ jobs:
         run: npm install
       - name: NPM run build
         run: npm run build
-      - name: Publish to NPM
-        run: npm publish --access public
+      - name: Get npm tag name if needed
+        id: npm_tag
+        run: |
+          pre_release_name=$(.github/script/get-pre-release-name.js)
+          echo "Pre release name:" $pre_release_name
+          echo ::set-output name=npm_tag::$pre_release_name  
+      - name: Publish to NPM latest
+        if:  steps.npm_tag.outputs.npm_tag == ''
+        run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
+      - name: Publish to NPM with tag
+        if: steps.npm_tag.outputs.npm_tag != ''
+        run: npm publish --tag ${{ steps.npm_tag.outputs.npm_tag }}
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -1,211 +1,183 @@
 #!/usr/bin/env node
 
 const prompt = require('prompt-sync')();
+const fs = require('fs');
+const path = require('path');
+
+const {
+  currentVersion,
+  getLatestNPMJSSdkVersion,
+  getMajorVersion,
+  isCurrentMajorVersion,
+  isPreRelease,
+  versionBump,
+} = require('./version-utils');
 const {
   logger,
   spawnOrFail,
   process,
   shouldContinuePrompt,
   checkWarning,
-  packDependency,
   updateDependency
 } = require('./utilities');
-const fs = require('fs');
-const path = require('path');
-const util = require('util');
 
-//  Go to root dir
-process.chdir(path.join(__dirname, '..'));
+const currentBranch = (spawnOrFail('git', [' branch --show-current'], { skipOutput: true })).trim();
 
-let package_json = JSON.parse(fs.readFileSync('package.json', 'utf-8'));
-const originalVersion = package_json['version'];
-const version = originalVersion.split('.').map(x => parseInt(x));
+const latestNPMJSSdkVersion = getLatestNPMJSSdkVersion(currentVersion);
 
-logger.log(`
-Choose one of the following bumping version options:
-    1. Patch
-    2. Minor
-    3. Major
-    4. Manual
-    5. Hotfix.`);
-const userInput = prompt('');
-if (!userInput || !['1', '2', '3', '4', '5'].includes(userInput)) {
-  logger.error('Exiting...', 'You must make selection to proceed!');
-  process.exit(1);
-}
-let versionString = '';
-switch (userInput) {
-  case '1':
-  case '5':
-    version[2] += 1;
-    versionString = version.join('.');
-    break;
-  case '2':
-    version[1] += 1;
-    version[2] = 0;
-    versionString = version.join('.');
-    break;
-  case '3':
-    version[0] += 1;
-    version[1] = 0;
-    version[2] = 0;
-    versionString = version.join('.');
-    break;
-  case '4':
-    versionString = prompt(
-      util.format(
-        // Color prompt
-        '\x1b[34m%s\x1b[0m',
-        'Specify the version in Semantic Versioning format:'
-      )
-    )
-      .toString()
-      .trim();
-    break;
-  default:
-    process.exit(1);
-}
-const release_option = userInput;
+const deployDemo = (version) => {
+  logger.log('Deploying unique release candidate Meeting Demo URL...');
+  // Checkout the meeting demo
+  process.chdir(path.join(__dirname, '../..'));
 
-package_json['version'] = versionString;
-const versionFile = 'src/versioning/Versioning.ts';
-const changeLog = 'CHANGELOG.md';
-const tag = `amazon-chime-sdk-component-library-react@${versionString}`;
-logger.warn(`
-    Warning: you are bumping the version
-    `);
-logger.log(
-  `        From: ${util.format('\x1b[32m%s\x1b[0m', originalVersion)} `
-);
-logger.log(
-  `        
-            To:   ${util.format('\x1b[31m%s\x1b[0m', versionString)} `
-);
+  let demoBranch = 'main';
+  if (!isCurrentMajorVersion()) {
+    demoBranch = `release-${getMajorVersion(currentVersion)}.x`;
+  }
 
-shouldContinuePrompt();
+  if (!fs.existsSync('amazon-chime-sdk')) {
+    spawnOrFail('git', [`clone --branch ${demoBranch} https://github.com/aws-samples/amazon-chime-sdk.git`]);
+  } else {
+    spawnOrFail('rm', ['-rf amazon-chime-sdk']);
+    spawnOrFail('git', [`clone --branch ${demoBranch} https://github.com/aws-samples/amazon-chime-sdk.git`]);
+  }
 
-spawnOrFail('git', ['fetch origin']);
-if (release_option !== '5') {
-  logger.warn(`
-    Warning: are you sure to reset the HEAD to origin/main?
-    Current staged and local changes will be lost!\n`);
-  shouldContinuePrompt();
-  logger.log('Reseting HEAD to origin/main');
-  spawnOrFail('git', ['reset --hard origin/main']);
-}
-spawnOrFail('git', [' clean -ffxd .']);
+  // Update the demo to the most up to date version of JS SDK
+  process.chdir(path.join(__dirname, '../../amazon-chime-sdk/apps/meeting'));
+  updateDependency(`amazon-chime-sdk-js@${latestNPMJSSdkVersion}`);
+  process.chdir(path.join(__dirname, '../../amazon-chime-sdk/apps/meeting'));
+  updateDependency('amazon-chime-sdk-component-library-react', `../../../amazon-chime-sdk-component-library-react/amazon-chime-sdk-component-library-react-${version}.tgz`);
 
+  process.chdir(path.join(__dirname, '../../amazon-chime-sdk/apps/meeting/serverless'));
+  const formattedVersion = version.replace(/\./g, '-');
+  spawnOrFail('node', [`./deploy.js -r us-east-1 -b chime-sdk-components-demo-${formattedVersion} -s chime-sdk-components-demo-${formattedVersion}`], { printErr: true });
 
-logger.log(`Updating ${versionFile} with ${versionString}`);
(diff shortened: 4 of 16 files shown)


.github/workflows/publish.yml
```diff
@@ -21,7 +21,19 @@ jobs:
         run: npm install
       - name: NPM run build
         run: npm run build
-      - name: Publish to NPM
-        run: npm publish --access public
+      - name: Get npm tag name if needed
```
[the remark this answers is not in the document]
xuesichao: Also this step seems to be executed anyway, should we remove "if needed" in its name? And add it to the next step "Publish to NPM latest if needed(possible/necessary/available)"

.github/script/get-pre-release-name.js
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env node
+
+const version = require('../../package.json').version;
+const preRelease = version.split('-');
+if (preRelease[1]) {
+  console.log(preRelease[1].split('.')[0]);
+} else {
+  console.log('');
+}
```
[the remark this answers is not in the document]
ltrung: Similar to jS SDK it is to tell whether it is a beta version (v.3.0.0-beta.0).

.github/workflows/publish.yml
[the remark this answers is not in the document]
ltrung: so for semver they call alpha/beta/etc to pre-release. We so far go with beta but i keep the script which is just to extract the pre-release part to be generic in case we need to change name later. In this case we tag if if it is a pre-release. The tag name is the prelease name.

.github/script/get-pre-release-name.js
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env node
+
+const version = require('../../package.json').version;
+const preRelease = version.split('-');
```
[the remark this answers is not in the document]
ltrung: I have a comment here https://github.com/aws/amazon-chime-sdk-component-library-react/pull/714/files#diff-c1445cf73423b12f427da3cd9ce51e312f70b28d57fff1567f2733d680801ef1R16

.github/script/get-pre-release-name.js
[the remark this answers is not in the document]
xuesichao: I didn't find the similar script in JS SDK, could you please show me the way?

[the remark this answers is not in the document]
xuesichao: I didn't find the similar script in JS SDK, are your referring to this?
https://github.com/aws/amazon-chime-sdk-js/blob/main/script/release#L49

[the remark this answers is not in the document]
ltrung: I just means It is copied from a similar PR in JS SDK https://github.com/aws/amazon-chime-sdk-js/pull/1905/files

scripts/release.js
```diff
@@ -1,211 +1,167 @@
...
+  spawnOrFail('npm', [`install amazon-chime-sdk-js@${latestNPMJSSdkVersion} styled-components styled-system`]);
+  checkWarning('npm', [`install -q ../amazon-chime-sdk-component-library-react/amazon-chime-sdk-component-library-react-${currentVersion}.tgz`], null, 'amazon-chime-sdk-component-library-react');
   process.chdir(path.join(__dirname, '..'));
+  spawnOrFail('rm', ['-rf ../dependency-check-app']);
 }
 
-spawnOrFail('git', ['add -A']);
-spawnOrFail('git', [`commit -m 'Publish ${tag}'`]);
-logger.log(`Created commit called: "Publish ${tag}"`);
+const cleanUp = (remoteBranch) => {
```
michhyun1, line 89: resetToHead

scripts/version-utils.js
```diff
@@ -0,0 +1,144 @@
...
+  spawnOrFail('npm', [`version ${newVersion} --no-git-tag-version`]);
+  updateChangelog(newVersion);
+
+  logger.log('Committing version bump...');
+  spawnOrFail('git', ['add -A']);
+  spawnOrFail('git', [`commit -m "Version bump for amazon-chime-sdk-js@${newVersion}"`]);
+  logger.log(`Do you want to upload these files to ${branchName} branch?\n`);
+  shouldContinuePrompt();
+  spawnOrFail('git', [`push origin HEAD:${branchName} -f`]);
+  logger.log('Please create a pull request to merge the version bump to main.');
```
michhyun1, line 136: is there something besides logger.log that will make this show in bold? Otherwise its easy to miss.

scripts/release.js
ltrung: As mention in the other PR in JS SDK, i want to keep this function generic as it can be whatever and not just reset to HEAD (we also have another command to clear out untracked changes too).

scripts/version-utils.js
ltrung: We can change to log.warn() but this will be a step in the MCM so not sure if it is needed.

scripts/release.js
```diff
@@ -1,211 +1,168 @@
...
+    spawnOrFail('mkdir', ['dependency-check-app']);
+  } else {
+    spawnOrFail('rm', ['-rf dependency-check-app']);
+    spawnOrFail('mkdir', ['dependency-check-app']);
+  }
+  process.chdir(path.join(__dirname, '../../dependency-check-app'));
+  spawnOrFail('npm', ['init -y']);
+  spawnOrFail('npm', ['install react react-dom']);
+  spawnOrFail('npm', [`install amazon-chime-sdk-js@${latestNPMJSSdkVersion} styled-components styled-system`]);
+  checkWarning('npm', [`install -q ../amazon-chime-sdk-component-library-react/amazon-chime-sdk-component-library-react-${currentVersion}.tgz`], null, 'amazon-chime-sdk-component-library-react');
```
devalevenkatesh, line 85: What is this `-q` here?

devalevenkatesh, line 85: What is this `-q` here?

scripts/release.js
```diff
@@ -1,211 +1,168 @@
...
-logger.log(`Created commit called: "Publish ${tag}"`);
+const cleanUp = (remoteBranch) => {
+  logger.warn(`Warning: Resetting HEAD${remoteBranch ? ` to ${remoteBranch}` : ''}.\nAll current staged and local changes will be lost.`);
+  shouldContinuePrompt();
+  spawnOrFail('git', [`reset --hard ${remoteBranch ? remoteBranch : ''}`]);
+  spawnOrFail('git', [' clean -ffxd .']);
+};
+
+const release = () => {
+  spawnOrFail('git', ['fetch origin'], { skipOutput: true });
```
devalevenkatesh, line 98: Should we move this `get fetch origin` inside the `getCurrentRemoteBranch()`?

devalevenkatesh, line 98: Should we move this `get fetch origin` inside the `getCurrentRemoteBranch()`?

scripts/release.js
```diff
@@ -1,211 +1,168 @@
...
+      cleanUp(remoteBranch);
+      spawnOrFail('npm', ['install']);
+      deployDemo(currentVersion);
+      break;
+    default:
+      if (option) {
+        logger.error('Invalid option');
+      }
+      process.exit(1);
+  }
```
michhyun1, line 165: Can you remidn me again why we need this? When would we use it?

scripts/version-utils.js
```diff
@@ -0,0 +1,149 @@
...
+    case 1: // Patch
+      if (isBeta) {
+        logger.error(`ERROR: Cannot increase patch in pre-release version `);
+        return undefined;
+      }
+      verArr[2] = Number(verArr[2]) + 1;
+      return verArr.join('.');
+    case 2: // Minor
+      if (isBeta) {
+        logger.error(`ERROR: Cannot increase patch in pre-release version `);
```
michhyun1, line 33: minor version not patch?

scripts/version-utils.js
```diff
@@ -0,0 +1,149 @@
+#!/usr/bin/env node
+
+const { logger, spawnOrFail, shouldContinuePrompt, process } = require('./utilities.js');
+const prompt = require('prompt-sync')();
+const path = require('path');
+const fs = require('fs');
+
+const currentVersion = require('../package.json').version;
+const latestNPMJSSdkVersion = spawnOrFail('npm', [`show amazon-chime-sdk-js version`], { skipOutput: true }).trim();
```
devalevenkatesh, line 9: This `latestNPMJSSdkVersion` is been used in `release` script to update the JS SDK dependency. Now, this `npm show` command does not return `3.0.0-beta.1` which is published to NPM but `2.29.0` hence, the component library will be stuck at `2.x` versions, is this intended? Should we handle based on which version we are releasing either from `2.x` (current major) or `main` (next major -> 3.x) and accordingly get the beta tag for no?

devalevenkatesh, line 9: This `latestNPMJSSdkVersion` is been used in `release` script to update the JS SDK dependency. Now, this `npm show` command does not return `3.0.0-beta.1` which is published to NPM but `2.29.0` hence, the component library will be stuck at `2.x` versions, is this intended? Should we handle based on which version we are releasing either from `2.x` (current major) or `main` (next major -> 3.x) and accordingly get the beta tag for no?

scripts/release.js
```diff
@@ -1,211 +1,168 @@
...
+};
+
+const buildAndPack = () => {
+  logger.log('Building package...');
+  spawnOrFail('npm', [`run build:publish`]);
+  logger.log('Packaging ...');
+  spawnOrFail('npm', ['pack'], { printErr: true });
+};
+
+const updateJSSdkDep = () => {
```
devalevenkatesh, line 55: Request to check comment: https://github.com/aws/amazon-chime-sdk-component-library-react/pull/714/files#r826419554.


Should we also take `isPreRelease` version into consideration while installing JS SDK?

devalevenkatesh, line 55: Request to check comment: https://github.com/aws/amazon-chime-sdk-component-library-react/pull/714/files#r826419554.


Should we also take `isPreRelease` version into consideration while installing JS SDK?

scripts/version-utils.js
```diff
@@ -0,0 +1,149 @@
...
+    logger.log('  3. Major');
+    logger.log('  4. Beta');
+    option = Number(prompt(''));
+  }
+  const newVersion = getNewVersion(currentVersion, option);
+  if (!newVersion) {
+    process.exit(1);
+  }
+
+  branchName = branchName ? branchName : `version-bump-${newVersion}`;
```
michhyun1, line 102: why is it called version-bump-{}? I tihnk for the js sdk, we just do release-3.0.0-beta.1 and for PMV we have release-2.x right?

scripts/version-utils.js
```diff
@@ -0,0 +1,149 @@
...
+  spawnOrFail('npm', [`version ${newVersion} --no-git-tag-version`]);
+  updateChangelog(newVersion);
+
+  logger.log('Committing version bump...');
+  spawnOrFail('git', ['add -A']);
+  spawnOrFail('git', [`commit -m "Version bump for amazon-chime-sdk-js@${newVersion}"`]);
+  logger.log(`Do you want to upload these files to ${branchName} branch?\n`);
+  shouldContinuePrompt();
+  spawnOrFail('git', [`push origin HEAD:${branchName} -f`]);
+  if (branchName === 'version-bump') {
```
michhyun1, line 138: wouldnt it be 'version-bump-{newVersion}' ?

scripts/version-utils.js
ltrung: to avoid conflict when we have dual release in one week (same as release branch) so it would be version-bump-3.0.0-beta-2 for v3 and version-bump-2.30.0 for v2

scripts/release.js
ltrung: Sorry need what?

michhyun1: why do we need the 'deploy demo' option

ltrung: So just when we need to include new commits in the release or QA finds bugs and we need to include fix, we will do cherry-pick those commits and just redeploy and skip all the previous steps of creating a release branch and bumping new dev versions.

scripts/release.js
ltrung: This is from the existing release script so not sure. I think this is what Lu/Jing added to check on dependency installation. However, I don't think it still work (I tried it out with npm 8 and it just auto resolves the conflict by updating js sdk version). Guess it is out of scope of this so we can come back to this later.

.github/workflows/deploy.yml
```diff
@@ -35,15 +37,29 @@ jobs:
...
+          cd $GITHUB_WORKSPACE/amazon-chime-sdk/apps/meeting/serverless
+          node ./deploy.js -r us-east-1 -b chime-sdk-components-demo-canary-prev -s chime-sdk-components-demo-canary-prev -l
+      - name: Get Pre Release Tag If Exists
+        id: npm_tag
+        run: |
+          pre_release_name=$(.github/script/get-pre-release-name.js)
+          echo "Pre release name:" $pre_release_name
+          echo ::set-output name=npm_tag::$pre_release_name
       - name: Publish Storybook Documentation
+        if:  steps.npm_tag.outputs.npm_tag == ''
```
devalevenkatesh, line 62: So this means, we will be keeping the storybook to 2.x version only as of now?

scripts/release.js
```diff
@@ -1,211 +1,170 @@
...
+const cleanUp = (remoteBranch) => {
+  logger.warn(`Warning: Resetting HEAD${remoteBranch ? ` to ${remoteBranch}` : ''}.\nAll current staged and local changes will be lost.`);
+  shouldContinuePrompt();
+  spawnOrFail('git', [`reset --hard ${remoteBranch ? remoteBranch : ''}`]);
+  spawnOrFail('git', [' clean -ffxd .']);
+};
+
+const release = () => {
+  const remoteBranch = getCurrentRemoteBranch();
+  if (!remoteBranch || (remoteBranch !== 'origin/main' && (/^origin\/release-[0-9]+\.x$/).test(remoteBranch))) {
```
devalevenkatesh, line 102: Should this also be `!(/^origin\/release-[0-9]+\.x$/).test(remoteBranch)` ?

scripts/release.js
```diff
@@ -1,211 +1,170 @@
...
-process.chdir(path.join(__dirname, '..'));
-spawnOrFail('rm', ['-rf dependency-check-app']);
-spawnOrFail('rm', [`-rf amazon-chime-sdk-component-library-react-${versionString}.tgz`]);
-
-if (release_option === '5') {
-  spawnOrFail('git', ['push origin HEAD:hotfix -f']);
-} else {
-  spawnOrFail('git', ['push origin HEAD:release -f']);
-}
+  //Bump next development version
```
devalevenkatesh, line 118: nit: Add a space after `//` in comment.

scripts/version-utils.js
```diff
@@ -0,0 +1,158 @@
...
+    // Switch back to the local branch
+    spawnOrFail('git', [`checkout ${currentBranch}`]);
+  }
+
+  spawnOrFail('npm', [`version ${newVersion} --no-git-tag-version`]);
+  updateChangelog(newVersion);
+
+  logger.log('Committing version bump...');
+  spawnOrFail('git', ['add -A']);
+  spawnOrFail('git', [`commit -m "Version bump for amazon-chime-sdk-js@${newVersion}"`]);
```
devalevenkatesh, line 143: This should be component library and not amazon-chime-sdk-js right?

scripts/version-utils.js
```diff
@@ -0,0 +1,158 @@
...
+  logger.warn(`  From: ${currentVersion}\n`);
+  logger.warn(`  To: ${newVersion}\n`);
+  if (prevReleaseBranch) {
+    logger.warn(`  This will also create ${prevReleaseBranch} branch.`);
+  }
+  shouldContinuePrompt();
+
+  if (prevReleaseBranch) {
+    const currentBranch = (spawnOrFail('git', [' branch --show-current'], { skipOutput: true })).trim();
+    spawnOrFail('git', [`checkout -b ${prevReleaseBranch}`]);
```
devalevenkatesh, line 127: Do we want to do a `git fetch --all` here to get all remote branches, in case if this `prevReleaseBranch` is not present locally?

Also, could you please once explain why we are updating the same branch? If this is to merge a fix after cutting the branch say some bug fix after testing then the commit message is a bit confusing.

scripts/update-sdk-version.js
```diff
@@ -1,28 +1,26 @@
 const path = require('path');
 const fs = require('fs');
+
+const { currentVersion, getLatestNPMJSSdkVersion }  = require('./version-utils');
```
devalevenkatesh, line 4: Does not look like we are using `getLatestNPMJSSdkVersion` below. Do we need this `update-sdk-version` script?

devalevenkatesh (changes requested):

.github/workflows/deploy.yml
```diff
@@ -2,7 +2,9 @@ name: Deploy Components Demo App Workflow
 
 on:
   push:
-    branches: [main]
+    pull_request:
```
LeviSklaroff, line 5: Is this syntax right?

.github/workflows/deploy.yml
```diff
@@ -35,15 +37,29 @@ jobs:
         with:
           repository: 'aws-samples/amazon-chime-sdk'
```
LeviSklaroff, line 38: I'm assuming this is going to change now that we're putting the demo in the integration folder? Otherwise this won't work for either main or 2x

scripts/utilities.js
```diff
@@ -14,24 +14,21 @@ const logger = {
 };
 
 const spawnOrFail = (command, args, options) => {
-  options = {
-    ...options,
-    shell: true
-  };
-  const cmd = exec(command, args, options);
+  const cmd = exec(command, args, { shell: true });
```
LeviSklaroff, line 17: Why are we no longer using the passed options?

.github/workflows/deploy.yml
ltrung: Since we use submodule now this is no longer a concern.

.github/workflows/deploy.yml
ltrung: This is just for testing.

scripts/utilities.js
ltrung: Because there is no callers that actually set it. I repurpose that `options` parameters to control whether we want to print out the command output instead.

scripts/release.js
```diff
@@ -1,211 +1,183 @@
...
+
+const latestNPMJSSdkVersion = getLatestNPMJSSdkVersion(currentVersion);
+
+const deployDemo = (version) => {
+  logger.log('Deploying unique release candidate Meeting Demo URL...');
+  // Checkout the meeting demo
+  process.chdir(path.join(__dirname, '../..'));
+
+  let demoBranch = 'main';
+  if (!isCurrentMajorVersion()) {
```
devalevenkatesh, line 34: `isCurrentMajorVersion` is a bit confusing. Would suggest renaming to `isNextMajorVersion`? I checked the implementation then could understood it.

ltrung: Actually `isCurrentMajorVersion` is the correct name as we only have current and previous. Right now for example, current version is v3 (even though it is still in beta phase). Next major version is v4 :)

devalevenkatesh (approved):

scripts/release.js
devalevenkatesh: Yeah I think whole wording of previous and current for me is confusing as actually for the builders the current stable will be our previous and next major version beta is the next major version. Since, we follow prev, current elsewhere this is fine.

Issue #745, opened 2022-02-15
Add test to verify BackgroundBlurProvider does not render unnecessarily when props are changed


richhx:
**Description of changes:**
We should have a test to verify that the `BackgroundBlurProvider` does not re-render unnecessarily in the case that its props (either options or spec) are changed by the downstream dependency on `BackgroundBlurVideoFrameProcessor` in the `amazon-chime-sdk-js`. This is essentially a follow-up on https://github.com/aws/amazon-chime-sdk-component-library-react/commit/21ba0cfa808c6694c62fc223975fb01f47110728 to add coverage and ensure we don't run into the same problem again.

**Testing:**
1. Have you successfully run npm run build:release locally?
Yes

2. How did you test these changes?
This pull request adds a test and verifies that it passes. Also verifies that all existing unit tests pass by running `npm run test`.

3. If you made changes to the component library, have you provided corresponding documentation changes?
No documentation changes needed since no changes to the component library (except a minor typo fix to the console log).

PR #746, opened 2022-02-15
Add BackgroundBlurProvider unit test to verify prop changes (#745)

richhx:
**Issue #:** #745 

**Description of changes:**
We should have a test to verify that the BackgroundBlurProvider does not re-render unnecessarily in the case that its props are changed by the downstream dependency on BackgroundBlurVideoFrameProcessor in the amazon-chime-sdk-js. This is essentially a follow-up on https://github.com/aws/amazon-chime-sdk-component-library-react/commit/21ba0cfa808c6694c62fc223975fb01f47110728 to add coverage and ensure we don't run into the same problem again.

Note that this also adds `jest-useragent-mock` as a dev dependency for ease of testing (needed to mock user agents).

**Testing**
Yes

This pull request adds a test and verifies that it passes. Also verifies that all existing unit tests pass by running npm run test.

No documentation changes needed since no changes to the component library (except a minor typo fix to the console log).

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     "jest": "^26.6.3",
     "jest-image-snapshot": "^4.4.0",
     "jest-puppeteer-docker": "^1.4.2",
+    "jest-useragent-mock": "^0.1.1",
     "mini-css-extract-plugin": "^0.7.0",
     "postcss-loader": "^3.0.0",
     "postcss-preset-env": "^6.6.0",
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,5 +1,6 @@
 module.exports = {
   testRegex: './*\\.test\\.tsx$',
+  setupFiles: ['jest-useragent-mock'],
   setupFilesAfterEnv: ['./tst/setupTests.js'],
   moduleFileExtensions: ['ts', 'tsx', 'js', 'json', 'jsx', 'node'],
   transform: {
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,6 +59,7 @@
         "jest": "^26.6.3",
         "jest-image-snapshot": "^4.4.0",
         "jest-puppeteer-docker": "^1.4.2",
+        "jest-useragent-mock": "^0.1.1",
         "mini-css-extract-plugin": "^0.7.0",
         "postcss-loader": "^3.0.0",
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -91,6 +91,7 @@
     "jest": "^26.6.3",
     "jest-image-snapshot": "^4.4.0",
     "jest-puppeteer-docker": "^1.4.2",
+    "jest-useragent-mock": "^0.1.1",
```
ltrung, line 94: Do we need to have a dependency just for this? It is pretty easy to mock user Agent (See https://github.com/aws/amazon-chime-sdk-js/blob/main/test/browserbehavior/DefaultBrowserBehavior.test.ts#L59)

richhx: Hmm I actually tried that before, but kept getting `TypeError: Cannot set property userAgent of [object Navigator] which has only a getter` even when doing `@ts-ignore`. I tried a few other approaches too but didn't seem to work :/ Open to recommendations/thoughts.

richhx:
See #747 instead.

package.json
ltrung: Regarding this you probably need to override the browser implementation of navigator (In JS SDK we have DOMMockBuilder object https://github.com/aws/amazon-chime-sdk-js/blob/main/test/browserbehavior/DefaultBrowserBehavior.test.ts#L8) .
Alternatively I think you can do something like suggested here https://stackoverflow.com/questions/1307013/mocking-a-useragent-in-javascript.

richhx: I actually tried the StackOverflow suggestion and I got a similar error (i.e. `__defineGetter__` does not exist). I can go ahead and override the implementation of navigator if that is the preferred approach by the team. I think this is easier / simpler to understand for developers though.

ltrung: The problem is the dependency is version 0 which means unstable and can have a lot of breaking changes. Also there are not many of downloads from that library so not sure how well supported that is.
Lastly, more dependencies means we have to keep update them -> more maintenance efforts. You probably have not seen the large amount of patches we have to do for all these dependencies :)

richhx: See changes in https://github.com/aws/amazon-chime-sdk-component-library-react/pull/747

PR #749, opened 2022-02-18
Bump the version of react component library in integration test demo

xuesichao:
**Issue #:** 

**Description of changes:**
Update the version for `amazon-chime-sdk-component-library-react" from 2.11.1 to 2.15.0 in test demo.

**Testing**
N/A

Yes, confirm the test demo works fine.

N/A

### DIFF
--- a/integration/app/test-demo/package-lock.json
+++ b/integration/app/test-demo/package-lock.json
@@ -8,7 +8,7 @@
       "name": "test-demo",
       "version": "0.1.0",
       "dependencies": {
-        "amazon-chime-sdk-component-library-react": "^2.11.1",
+        "amazon-chime-sdk-component-library-react": "^2.15.0",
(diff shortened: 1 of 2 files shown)


devalevenkatesh (approved):

ltrung:
Do we need to do this for every version? Suggest to update the script then.
Another alternative solution is to use relative path and point to the source folder similar to JS SDJ

xuesichao:
> Do we need to do this for every version? Suggest to update the script then.
> Another alternative solution is to use relative path and point to the source folder similar to JS SDJ

@ltrung I tried to do that this afternoon. But after the change, there were errors after running the `npm run start`.
Maybe it's because the `npm run build` script in the root directory is different for react sdk and js sdk.

The error is:
```
Uncaught Error: Invalid hook call. Hooks can only be called inside of the body of a function component.
```

PR #750, opened 2022-03-02
Update channel list component to display a more detailed version of the channel including last channel message and its timestamp

jingwy:
**Description of changes:**
This change is to update channel list component to display a more detailed version of the channel including last channel message and its timestamp.

Original ChannelList component:
image

Updated ChannelList component:
image

**Testing**

3. How did you test these changes? Tested by adding unit tests and load the change in storybook. The UI is shown below:
image

5. If you made changes to the component library, have you provided corresponding documentation changes? Yes.


nainkunal933:
@jingwy Can you please remove the `package-lock.json` from the PR commit?

src/components/ui/Chat/DetailedChannelList/DetailedChannelList.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+import DetailedChannelList from './';
+import { _DetailedChannelList } from './DetailedChannelList.stories.tsx';
+import { lightTheme } from '../../../../theme/';
+import { GlobalStyles } from '../../../../theme/GlobalStyles';
+import PopOverItem from '../../PopOver/PopOverItem';
+
+# DetailedChannelList
+
+The DetailedChannelList component displays a column of chat channels along with last message's content and timestamp.
```
nainkunal933, line 12: Can you increase the documentation on the `DetailedChannelList` component? We already have a Channel List component and good documentation is important for builders to know the difference between the two. Please add more details like you added in the changelog.

> Add a new chat UI component DetailedChannelList which displays a list of channels with channel name, last message content and  timestamp, as well as unread channel badge.

src/components/ui/Chat/DetailedChannelList/DetailedChannelItem.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const {
+    name,
+    channelMessage,
+    channelMessageTime,
+    actions,
+    isSelected,
+    onClick,
+    unread,
+    unreadBadgeLabel,
+  } = props;
```
michhyun1, line 44: needs `...rest` prop

src/components/ui/Chat/DetailedChannelList/DetailedChannelItem.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <StyledDetailedChannelItem
+      {...props}
+      className={classnames({ 'ch-selected': isSelected, 'ch-unread': unread })}
+    >
+      <div className={'ch-channel-button'} onClick={() => onClick()}>
+        <div className="ch-label">{name}</div>
+        <div className="channel-message">{channelMessage}</div>
+        <div className="channel-message-time">{channelMessageTime}</div>
```
michhyun1, line 53: ch-channel-button-label
ch-channel-button-message
ch-channel-button-message-time

src/components/ui/Chat/DetailedChannelList/DetailedChannelItem.tsx
```diff
@@ -0,0 +1,77 @@
...
+      </div>
+      {unread && unreadBadgeLabel && (
+        <Badge value={unreadBadgeLabel} className="ch-unread-badge" />
```
michhyun1, line 56: ch-channel-unread-badge

src/components/ui/Chat/DetailedChannelList/DetailedChannelList.mdx
michhyun1: I guess my main concern is, is there any way we can make a more generic component to combine the two - there is a lot of code and documentation here for very similar components. Imagine the maintenance of both components compared to just one. Is it possible to add optional props to the Channel List component to make it bcomaptible?

michhyun1 (changes requested):
So this looks to me like we can just add on to the existing component - is there a reason why we can't extend the funtionality to the existing Channel List component? I don't think its worth creating a completely new component for this when we can just add the functionality to the Channel List component. It seems very do-able. Less documentation and code to maintain that way.

src/components/ui/Chat/ChannelList/ChannelItem.tsx
```diff
@@ -26,21 +26,65 @@ export interface ChannelItemProps
   unread?: boolean;
   /** Content of the badge signaling unread messages, such as a badge displaying the number. */
   unreadBadgeLabel?: string;
+  /** Boolean to indicate if we want to display the channel with last channel message and timestamp */
+  displayLastChannelMessageAndTimestamp?: boolean;
```
michhyun1, line 30: instead of having this extra prop, maybe you can just check if lastChannelMessage or lastChannelMessageTimestamp has some value or not

michhyun1: They should be able to enable timestamp OR preview message, right? This makes it seem like they need to always enable both at the same time

src/components/ui/Chat/ChannelList/ChannelItem.tsx
```diff
@@ -26,21 +26,65 @@ export interface ChannelItemProps
...
+        <Button className="ch-channel-button" label={name} onClick={onClick} />
+      )}
+      {displayLastChannelMessageAndTimestamp && unread && unreadBadgeLabel && (
+        <Badge value={unreadBadgeLabel} className="ch-unread-badge-detailed" />
+      )}
+      {!displayLastChannelMessageAndTimestamp && unread && unreadBadgeLabel && (
         <Badge value={unreadBadgeLabel} className="ch-unread-badge" />
       )}
-      {actions && isSelected && (
+      {displayLastChannelMessageAndTimestamp && actions && isSelected && (
```
michhyun1, line 71: You could probably combine these instead of having 4 conditionals

src/components/ui/Chat/DetailedChannelList/DetailedChannelList.mdx
jingwy: Thanks for the feedback. Discussed offline and made the change as suggested to update the existing ChannelList component instead of creating a new one.

src/components/ui/Chat/ChannelList/ChannelItem.tsx
jingwy: Ah right, will try it out. Thanks.

src/components/ui/Chat/DetailedChannelList/DetailedChannelItem.tsx
jingwy: Thanks addressed in the latest revision.

src/components/ui/Chat/ChannelList/ChannelItem.tsx
```diff
@@ -22,25 +22,80 @@ export interface ChannelItemProps
...
+  /** Signifies if there are unread message in the channel. */
   unread?: boolean;
   /** Content of the badge signaling unread messages, such as a badge displaying the number. */
   unreadBadgeLabel?: string;
+  /** Indicate if we want to display the detailed channel view with last channel message or timestamp */
+  displayDetailedChannelView?: boolean;
+  /* Content of the last channel message to be displayed */
+  lastChannelMessage?: string;
+  /* Timestamp of the last channel message to be displayed */
+  lastChannelMessageTimestamp?: string;
```
simmkyu, line 34: I just noticed that the `timestamp` prop in the component library has a string type. Usually, the timestamp should have either `Date` or `number`, and the SDK should decide which format to display.

src/components/ui/Chat/ChannelList/ChannelItem.tsx
```diff
@@ -22,25 +22,80 @@ export interface ChannelItemProps
...
   return (
     <StyledChannelItem
       {...props}
       className={classnames({ 'ch-selected': isSelected, 'ch-unread': unread })}
     >
-      <Button className="ch-channel-button" label={name} onClick={onClick} />
-      {unread && unreadBadgeLabel && (
-        <Badge value={unreadBadgeLabel} className="ch-unread-badge" />
+      {displayDetailedChannelView ? (
+        <div className={'ch-detailed-channel'} onClick={onClick}>
```
simmkyu, line 64: Use `"ch-detailed-channel"` to match others.

src/components/ui/Chat/ChannelList/ChannelItem.tsx
```diff
@@ -22,25 +22,80 @@ export interface ChannelItemProps
...
   onClick: () => void;
-  /** Signifies of there are unread message in the channel. */
+  /** Signifies if there are unread message in the channel. */
   unread?: boolean;
   /** Content of the badge signaling unread messages, such as a badge displaying the number. */
   unreadBadgeLabel?: string;
+  /** Indicate if we want to display the detailed channel view with last channel message or timestamp */
+  displayDetailedChannelView?: boolean;
+  /* Content of the last channel message to be displayed */
+  lastChannelMessage?: string;
```
simmkyu, line 32: From a builder perspective, these props do not need to be the *last* message or timestamp. i.e., the `ChannelItem` component does not check whether these props are really last messages or have any special message like `Last message: ${lastChannelMessage}`.

Rather, you can simply add one more prop, say `detailView`. If `detailView` exists, show it in a detailed view.

```
detailView: ReactNode // ReactNode is ReactElement, string, or any other element
```

```
detailView={
  <div className="ch-detailed-channel-message">
    {lastChannelMessage}
  </div>
  <div className="ch-detailed-channel-message-time">
    {lastChannelMessageTimestamp}
  </div>
}
```

simmkyu: Actually, NVM.. I just reviewed screenshots again. You can't just pass down to show the date and time on the right side.

michhyun1, line 32: I was thinking of props more like:

previewString?: string,
timestamp?: number (or Date)

and then in the implementation below:
```
if timestamp -> render the timestamp and actions dropdown icon, otherwise, render just the action dropdown

if previewString -> render a div with the channel name and previewString , otherwise render only the channel button
```

PR #761, opened 2022-03-18
Remove meetingRegion from MeetingManager

ltrung:
**Description of changes:**
- Remove `meetingRegion` from `MeetingManager` because we do not use that value anywhere in the library. Application can still get it from the `createMeeting` API response in aws-sdk.
- Fix some minor bugs that remove the `meetingStatus.Ended` as well as `DeviceLabels.None` check.

**Testing**

2. How did you test these changes? Manually test with latest meeting demo.

3. If you made changes to the component library, have you provided corresponding documentation changes? Yes.

### DIFF
--- a/src/components/migrationToV3.stories.mdx
+++ b/src/components/migrationToV3.stories.mdx
@@ -20,97 +20,38 @@ npm install --save amazon-chime-sdk-component-library-react@beta amazon-chime-sd
 ## What's New
 
 - `MeetingManager`'s `join()` method no longer takes an `EventReporter` and instead takes an `EventController`.
+- Add `MeetingSessionConfiguration` as a required parameter to `MeetingManager.join()` method. With this change the builders have direct access to `MeetingSessionConfiguration`, this will allow more flexibility to customize the `MeetingSession`.
+- Add `MeetingManagerJoinOptions` as a new interface for the `options` parameter of the `MeetingManager.join` method.
+- Add `deviceLabels`, `eventController`, `logLevel`, `postLoggerConfig`, `logger`, `enableWebAudio`, and `activeSpeakerPolicy` to `MeetingManagerJoinOptions` interface.
+- Remove `MeetingSessionConfiguration` properties from `MeetingProvider` props.
+- Remove `deviceLabels`, `eventController`, `logLevel`, `postLogConfig`, `logger`, `enableWebAudio`, and `activeSpeakerPolicy` from `MeetingProvider` props.
+- Remove `meetingRegion` from `MeetingManager`. Previously we just cache this value from meetingInfo returned by
+[CreateMeeting](https://docs.aws.amazon.com/chime/latest/APIReference/API_CreateMeeting.html) but it is not used
+anywhere in the library.
 - Deprecate `useBandwidthMetrics` hook in favor of `useMediaStreamMetrics`.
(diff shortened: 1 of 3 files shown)


nainkunal933 (approved):

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -60,8 +60,6 @@ export class MeetingManager implements AudioVideoObserver {
 
   meetingId: string | null = null;
 
-  meetingRegion: string | null = null;
```
devalevenkatesh, line 63: Can you also remove it when we do cleanup and set this to null?

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -308,6 +305,14 @@ export class MeetingManager implements AudioVideoObserver {
   audioVideoDidStop = (sessionStatus: MeetingSessionStatus): void => {
     const sessionStatusCode = sessionStatus.statusCode();
     switch (sessionStatusCode) {
+      case MeetingSessionStatusCode.MeetingEnded:
```
devalevenkatesh, line 308: How did I miss this, I thought only `AudioAttendeeEnded` was removed :(. Thanks for the fix.

src/types/index.ts
```diff
@@ -69,7 +69,7 @@ export enum DevicePermissionStatus {
 }
 
 export enum DeviceLabels {
-  None,
+  None = 1,
```
devalevenkatesh, line 72: Does this need to be documented in migration? Or should be just fix the `if` check?

src/providers/MeetingProvider/MeetingManager.ts
LeviSklaroff: Was this supposed to replace AudioAttendeeEnded? I looked back at my change and this wasn't there before either.

ltrung: So MeetingEnded and AudioAttendeeEnded are both = 6 :) Thats why we should replace instead of removing it.

src/providers/MeetingProvider/MeetingManager.ts
ltrung: I already did no?

devalevenkatesh (approved):

src/components/migrationToV3.stories.mdx
```diff
@@ -20,97 +20,38 @@ npm install --save amazon-chime-sdk-component-library-react@beta amazon-chime-sd
...
-  - `logger`
-  - `enableWebAudio`
-  - `activeSpeakerPolicy`
+- `deviceLabels`
+- `eventController`
+- `logLevel`
+- `postLogConfig`
+- `logger`
+- `enableWebAudio`
+- `activeSpeakerPolicy`
```
xuesichao, line 52: Why we remove the indention? I think we should either remove dash for "MeetingSessionConfiguration properties:" and "Other props:" or add intention to their children bullet points.

xuesichao: It looks like this now:

- `MeetingSessionConfiguration` properties:
- `VideoDownlinkBandwidthPolicy`
- `VideoUplinkBandwidthPolicy`
- `simulcastEnabled`
- `reconnectTimeoutMs`
- `keepLastFrameWhenPaused`
- Other props:
- `deviceLabels`
- `eventController`
- `logLevel`
- `postLogConfig`
- `logger`
- `enableWebAudio`
- `activeSpeakerPolicy`

src/providers/MeetingProvider/MeetingManager.ts
LeviSklaroff: Gotcha, I should have looked more into what it was actually doing

Issue #529, opened 2021-06-07
PreviewVideo also stops LocalVideo without setting isVideoEnabled to false


alabianca:
**Describe the bug**

`audioVideo.stopVideoPreviewForInput` also ends LocalVideo, but `isVideoEnabled` of `useLocalVideo` still remains true.

**To Reproduce**
1. Run meeting demo in this forked repo https://github.com/alabianca/amazon-chime-sdk-component-library-react
2. Enter the call and enable your local video
3. In the ControlBar click the `Open for Settings` button
4. Close the modal that contains the VideoPreview
5. Notice how the local video stream ended, but the isVideoEnabled is still true. It's easy to spot due to the Video icon in the controlbar being set to 'on'

**Expected behavior**
1. Either the local video not stopping after PreviewVideo unmounts
2. Or... isVideoEnabled to be set to false. I could create a useEffect to turn the local video on again if it is set to false and but selectedDevice of `useVideoInputs` is not null.

**Screenshots**

<img width="979" alt="Screen Shot 2021-06-07 at 5 39 02 PM" src="https://user-images.githubusercontent.com/25507518/121096683-4c1fe280-c7b8-11eb-8f04-033a33d4032d.png">

**Desktop (please complete the following information):**
 - OS: macOS Catalina 10.15.7
 - Browser Chrome
 - Version 91.0.4472.77 (Official Build) (x86_64)

**Smartphone (please complete the following information):**
 - Device: N/A
 - OS: N/A
 - Browser N/A
 - Version N/A

**Additional context**
Additional background context:
Our app currently allows a user to open up a modal during the call in which the user can make some device selections. In the modal we display a preview of their video (in case they change their video input). When the user closes the modal, it also turns off their local video, but the Video Input icon is set to 'on'.

ltrung:
This is because JS SDK does not support to have video preview and local video to be on at the same time. Could you try a similar workaround as suggested in https://github.com/aws/amazon-chime-sdk-js/issues/502 (basically have another meeting provider with a different device controller)?

xuesichao:
Hi @alabianca, thanks for bringing us this issue, we raised this PR #556 to fix the issue.

>Either the local video not stopping after PreviewVideo unmounts

Due to the limitation of JS SDK, it's hard to do achieve this. `LocalVideo` and `PreviewVideo` shares the same stream from selected camera, unmounting `PreviewVideo` closes `LocalVideo` as well.

[...]

PR #773, opened 2022-04-05
Improve Device Management

xuesichao:
**Issue #:** 
Migrate to JS SDK DeviceController change.
Changes under "Adapt to  JS SDK Device Controller V3 Changes" in design doc.

**Description of changes:**

* Migrate to JS SDK DeviceController change following [migration guide](https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/17_Migration_to_3_0.md#device-controller).
  * Remove duplicate eslint rule `no-unused-vars`because we already have `@typescript-eslint/no-unused-vars`.
  * Remove preset device selection options.
  * Remove `setIsVideoEnabled` from `useLocalVideo` hook and `LocalVideoProvider`. This change is to revert the workaround for issue #529. The issue is fixed in JS SDK V3 so the workaround is no longer needed.

**Testing**
Yes

Install the React SDK in meeting demo and do e2e testing. Confirm it works fine.
No specific test steps. Basically try e2e testing and confirm device related features are working well.

Yes


src/components/sdk/DeviceSelection/CameraSelection/index.tsx
```diff
@@ -14,25 +13,18 @@ interface Props extends BaseSdkProps {
...
   notFoundMsg = 'No camera devices found',
   label = 'Camera source',
-  appendSampleDevices = true,
   ...rest
 }) => {
   const meetingManager = useMeetingManager();
-  const videoInputConfig: DeviceConfig = {
-    additionalDevices: appendSampleDevices,
+  const { devices, selectedDevice } = useVideoInputs();
+  const selectVideoInput = async (deviceId: string): Promise<void> => {
```
devalevenkatesh, line 25: Should this `deviceId: string` be `VideoInputDevice` type instead? I know we wont need a transform device here but we also do not want just the `stringId` right?

devalevenkatesh, line 25: Should this `deviceId: string` be `VideoInputDevice` type instead? I know we wont need a transform device here but we also do not want just the `stringId` right?

src/components/sdk/DeviceSelection/MicSelection/index.tsx
```diff
@@ -14,21 +13,15 @@ interface Props extends BaseSdkProps {
...
 }) => {
-  const audioInputConfig: DeviceConfig = {
-    additionalDevices: appendSampleDevices,
-  };
   const selectAudioInput = useSelectAudioInputDevice();
-  const { devices, selectedDevice } = useAudioInputs(audioInputConfig);
+  const { devices, selectedDevice } = useAudioInputs();
 
   return (
     <DeviceInput
```
devalevenkatesh, line 27: Will now the prop on the `DeviceInput` be not called `selectedDeviceId` anymore since we are removing the concept of `deviceId` and moving to use `Device` (`AudioInputDevice` | `VideoInputDevice`)?

devalevenkatesh, line 27: Will now the prop on the `DeviceInput` be not called `selectedDeviceId` anymore since we are removing the concept of `deviceId` and moving to use `Device` (`AudioInputDevice` | `VideoInputDevice`)?

src/components/sdk/DeviceSelection/CameraSelection/index.tsx
xuesichao: Yeah, I'm working on it in a separate PR. This PR is for the first task in design doc - Adapt to  JS SDK Device Controller V3 Changes. Rest of tasks will be in another PR.

src/components/sdk/DeviceSelection/MicSelection/index.tsx
xuesichao: Yeah, I'm renaming it to `selectedDevice`. Will be out in next PR.

src/components/sdk/DeviceSelection/CameraSelection/index.tsx
xuesichao: As discussed in standup, I will not raise another PR but push commits to this one. So this change will be in later.

src/components/sdk/DeviceSelection/MicSelection/index.tsx
xuesichao: As discussed in standup, I will not raise another PR but push commits to this one. So this change will be in later.

src/components/sdk/DeviceSelection/CameraSelection/index.tsx
devalevenkatesh: Got it!

src/components/sdk/MeetingControls/AudioInputControl.tsx
```diff
@@ -22,46 +24,56 @@ interface Props extends BaseSdkProps {
...
+            mutedTitle={mutedIconTitle}
+            unmutedTitle={unmutedIconTitle}
+          />
+        }
+        onClick={toggleMute}
+        label={muted ? unmuteLabel : muteLabel}
+        popOver={dropdownOptions}
+        {...rest}
+      />
+      {error && (
```
devalevenkatesh, line 68: Good that you handle it, but, as discussed in meetings logging the error is fine for now unless we have  gone through UX to finalize this.

devalevenkatesh: If we are doing so in one component consistency of UX is also important on the rest of controls on UX for SDK connected UI components.

src/hooks/sdk/useSelectAudioOutputDevice.tsx
```diff
@@ -1,16 +1,18 @@
...
 
-export const useSelectAudioOutputDevice = () => {
+export const useSelectAudioOutputDevice = (): ((
+  deviceId: string
+) => Promise<void>) => {
   const meetingManager = useMeetingManager();
 
   const selectDevice = useCallback(async (deviceId: string) => {
-    if (supportsSetSinkId()) {
+    if (new DefaultBrowserBehavior().supportsSetSinkId()) {
```
devalevenkatesh, line 15: Looks like we do use `new DefaultBrowserBehavior().supportsSetSinkId()` now at multiple places so good to keep the device-utils?

src/components/sdk/DeviceSelection/CameraSelection/index.tsx
```diff
@@ -14,25 +13,18 @@ interface Props extends BaseSdkProps {
...
+    await meetingManager.selectVideoInputDevice(deviceId);
```
devalevenkatesh, line 26: Instead of meetingManager.selectVideoInputDevice can we just use the `useSelectVideoInputDevice` hook? Trying to unify the way we select these devices with a common pattern.

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
```diff
@@ -118,22 +116,23 @@ const VideoInputBackgroundBlurControl: React.FC<Props> = ({
...
-              {isLoading && <Spinner width="1.5rem" height="1.5rem" />}
-              {backgroundBlurLabel}
-            </>
-          }
           checked={isVideoTransformDevice(activeVideoDevice)}
           disabled={isLoading}
           onClick={toggleBackgroundBlur}
-        />
+        >
+          <>
```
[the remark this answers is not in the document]
xuesichao: Yeah, we can remove this. Will update later.

src/components/sdk/DeviceSelection/CameraSelection/index.tsx
devalevenkatesh: I see that we do it in SpeakerSelection hence wondering if we can do it or if out of scope for now and needs lot of changes?

src/components/sdk/MeetingControls/AudioInputControl.tsx
xuesichao: Discussed offline, I will log the error message instead of displaying via `Modal` UI component.

src/components/sdk/MeetingControls/AudioOutputControl.tsx
```diff
@@ -26,7 +27,7 @@ const AudioOutputControl: React.FC<Props> = ({
   const { devices, selectedDevice } = useAudioOutputs();
   const { isAudioOn, toggleAudio } = useLocalAudioOutput();
   const audioOutputOnClick = async (deviceId: string): Promise<void> => {
-    if (supportsSetSinkId()) {
+    if (new DefaultBrowserBehavior().supportsSetSinkId()) {
       await meetingManager.selectAudioOutputDevice(deviceId);
```
devalevenkatesh, line 31: Same comment about using the ``useSelectAudioOutputDevice`.

src/hooks/sdk/useSelectAudioOutputDevice.tsx
xuesichao: It's a TODO item to remove it.
```tsx
// TODO: Remove this and use DefaultBrowserBehavior.supportsSetSinkId from JS SDK v2.x
export const supportsSetSinkId = (): boolean => {
  return 'setSinkId' in HTMLAudioElement.prototype;
};
```

src/hooks/sdk/useSelectAudioInputDevice.tsx
```diff
@@ -6,11 +6,13 @@ import { useCallback } from 'react';
...
 
-export const useSelectAudioInputDevice = () => {
+export const useSelectAudioInputDevice = (): ((
+  device: Device | AudioTransformDevice | null
+) => Promise<void>) => {
   const meetingManager = useMeetingManager();
 
   const selectDevice = useCallback(
-    async (device: Device | AudioTransformDevice) => {
+    async (device: Device | AudioTransformDevice | null) => {
```
devalevenkatesh, line 15: Similar comment to using AudioInputDevice from JS SDK?
https://github.com/aws/amazon-chime-sdk-js/blob/main/src/devicecontroller/AudioInputDevice.ts

src/hooks/sdk/useSelectVideoInputDevice.tsx
```diff
@@ -6,7 +6,9 @@ import { useCallback } from 'react';
 import { useLocalVideo } from '../../providers/LocalVideoProvider';
 import { useMeetingManager } from '../../providers/MeetingProvider';
 
-export const useSelectVideoInputDevice = () => {
+export const useSelectVideoInputDevice = (): ((
+  deviceId: string
```
devalevenkatesh, line 10: This should now be supporting both `Device` and `VideoTransformDevice` as well as `null` right?

xuesichao: No, `null` is no longer supported by `selectVideoInputDevice` in V3. The equivalent will be a new API `unselectVideoInputDevice()`.

src/hooks/sdk/useSelectAudioInputDevice.tsx
xuesichao: Yes will be.

src/components/sdk/DeviceSelection/CameraSelection/BackgroundBlurCheckbox.tsx
```diff
@@ -28,20 +28,22 @@ export const BackgroundBlurCheckbox: React.FC<Props> = ({
   const [isLoading, setIsLoading] = useState(false);
   const meetingManager = useMeetingManager();
   // TODO: Move this to the Video Input Provider and expose only one selected Video Input device state
-  const [device, setDevice] = useState<Device | VideoTransformDevice | null>(
-    meetingManager.selectedVideoInputTransformDevice
-  );
+  const [device, setDevice] = useState<
+    Device | VideoTransformDevice | undefined
```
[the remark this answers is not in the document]
xuesichao: Yes, in later commit.

src/components/sdk/DeviceSelection/CameraSelection/index.tsx
xuesichao: I will try to unify them later. I think there should not be too many changes

src/components/sdk/PreviewVideo/index.tsx
```diff
@@ -44,24 +44,17 @@ export const PreviewVideo: React.FC<BaseSdkProps> = (props) => {
     return () => {
       if (videoElement) {
         audioVideo?.stopVideoPreviewForVideoInput(videoElement);
-        setIsVideoEnabled(false);
```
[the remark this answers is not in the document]
xuesichao: >There was an issue about isVideoEnabled state not being set to false once the PreviewVideo is stopped.
https://github.com/aws/amazon-chime-sdk-component-library-react/issues/529

In JS SDK V3, `startVideoPreviewForVideoInpu`t and `stopVideoPreviewForVideoInput` do not affect the local video state anymore. So we can revert the workaround for that issue.

>This change is reverting it, so, wanted to check, do we need to maintain two separate states for LocalVideo and PreviewVideo enabling/disabling status?

I think we can do that. But it is kinda out of scope, will try if have extra time.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -544,22 +528,24 @@ export class MeetingManager implements AudioVideoObserver {
...
+      let deviceId = null;
       if (isAudioTransformDevice(device)) {
-        innerDevice = await device.intrinsicDevice();
+        const innerDevice = await device.intrinsicDevice();
+        deviceId = DefaultDeviceController.getIntrinsicDeviceId(innerDevice);
       } else {
-        innerDevice = device;
+        const innerDevice = device;
+        if (innerDevice) {
+          deviceId = DefaultDeviceController.getIntrinsicDeviceId(innerDevice);
```
devalevenkatesh, line 545: The [`getIntrinsicDeviceId`](https://github.com/aws/amazon-chime-sdk-js/blob/main/src/devicecontroller/DefaultDeviceController.ts#L712) returns `undefined` for `null` `deviceId` so in that case should the `deviceId` be actually `undefined` or `null`? Not sure if that matters. @ltrung do you have any concerns?

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -727,7 +709,7 @@ export class MeetingManager implements AudioVideoObserver {
     this.selectedVideoInputTransformDeviceObservers.push(callback);
   };
 
-  unsubscribeFromSelectedVideoInputTranformDevice = (
+  unsubscribeFromSelectedVideoInputTransformDevice = (
```
devalevenkatesh, line 712: So we still have the notion of separate handling of transform device and normal device right?

xuesichao: For current PR, yes. The `deviceId` type change is under development.

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: I refactored this code in my latest change (not in this PR yet) and `let deviceId = null;` is removed.
`deviceId` should be type of `string | string[] | undefined`.

xuesichao:
```tsx
let innerDevice: Device | null;

if (isVideoTransformDevice(device)) {
  innerDevice = await device.intrinsicDevice();
} else {
  innerDevice = device;
}

const deviceId = DefaultDeviceController.getIntrinsicDeviceId(innerDevice);
```

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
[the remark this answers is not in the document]
xuesichao: Confirmed we can not remove this.

src/hooks/sdk/useSelectVideoInputDevice.tsx
xuesichao: The deviceId type change will be in next PR. the type will be `Device | VideoTransformDevice` (`VideoInputDevice`)

src/components/sdk/DeviceSelection/MicSelection/index.tsx
xuesichao: This will be updated in the next PR.

src/components/sdk/DeviceSelection/CameraSelection/index.tsx
xuesichao: Yes, change will be in next PR.

src/components/migrationToV3.stories.mdx
```diff
@@ -440,4 +431,85 @@ const MyChild = () => {
...
+The detailed changes are:
+
+- Remove `appendSampleDevices` from Props of `CameraSelection`, `MicSelection`, `AudioInputControl`, `AudioInputVFcontrol`, and `VideoInputControl`.
+- Remove `additionalDevices` from Props of `useAudioInputs` and `useVideoInputs` hook.
+- Remove `DeviceConfig` type.
+
+Before in 2.x:
+
+```tsx
+// Take `CameraSelection` as an example.
```
devalevenkatesh, line 460: Can drop these comments, example is clear enough I think.

src/components/migrationToV3.stories.mdx
```diff
@@ -440,4 +431,85 @@ const MyChild = () => {
...
+import React from 'react';
+import {
+  MeetingProvider,
+  CameraSelection,
+} from 'amazon-chime-sdk-component-library-react';
+
+const App = () => {
+  return (
+    <MeetingProvider>
+      {/* If you don't want preset device selection options. */}
```
devalevenkatesh, line 471: This should be "If you want.." right?

src/components/migrationToV3.stories.mdx
```diff
@@ -440,4 +431,85 @@ const MyChild = () => {
...
+
+The detailed changes are:
+
+- Remove `appendSampleDevices` from Props of `CameraSelection`, `MicSelection`, `AudioInputControl`, `AudioInputVFcontrol`, and `VideoInputControl`.
+- Remove `additionalDevices` from Props of `useAudioInputs` and `useVideoInputs` hook.
+- Remove `DeviceConfig` type.
+
+Before in 2.x:
+
+```tsx
```
devalevenkatesh, line 459: Can you once check if we have generally used on `jsx` as well as `tsx` in component library? In JS SDK, we only use `js` and not `typescript` for docs hence want to check.

Here, we do not use type hence change the `tsx` to `jsx`.

src/components/migrationToV3.stories.mdx
```diff
@@ -440,4 +431,85 @@ const MyChild = () => {
...
+  return (
+    <MeetingProvider>
+      {/* If you don't want preset device selection options. */}
+      <CameraSelection appendSampleDevices={true} />
+    </MeetingProvider>
+  );
+};
+```
+
+```jsx
```
devalevenkatesh, line 478: If using `tsx` in other examples, this should be `tsx` since we show the `DeviceConfig` type or else should be fine to drop it.

src/components/sdk/DeviceSelection/SpeakerSelection/index.tsx
```diff
@@ -26,7 +26,7 @@ export const SpeakerSelection: React.FC<Props> = ({
   const { devices, selectedDevice } = useAudioOutputs();
   const selectAudioOutput = useSelectAudioOutputDevice();
 
-  async function selectDevice(deviceId: string) {
+  async function selectDevice(deviceId: string): Promise<void> {
     selectAudioOutput(deviceId);
```
devalevenkatesh, line 30: Not  related to your change, but since `selectAudioOutput` also returns a Promise should we `await it?

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -120,39 +111,39 @@ const AudioInputVFControl: React.FC<Props> = ({
...
-          }
           checked={isVoiceFocusEnabled}
           disabled={isLoading}
           onClick={() => {
             setIsLoading(true);
             setIsVoiceFocusChecked((current) => !current);
           }}
-        />
+        >
+          <>
```
devalevenkatesh, line 142: May be we can remove this Fragment wrapping?

src/components/sdk/PreviewVideo/PreviewVideo.stories.mdx
```diff
@@ -5,9 +5,9 @@ import { PreviewVideo } from './';
 
 # PreviewVideo
 
-The `PreviewVideo` component renders a preview of the local user's video. If there is no available video input, the video will be blank. 
+The `PreviewVideo` component renders a preview of the local video. If there is no available video input, the video will be blank. 
 
-A `PreviewVideo` is typically used on a pre-meeting view, or a device settings page. The React SDK depends on the Amazon Chime SDK for JavaScript, wherein, a single video stream is attached to a video element at a time. Hence, if `PreviewVideo` and `LocalVideo` component both are rendered on the same page and if both are enabled, stopping the video preview in `PreviewVideo` component will result into disconnecting the video stream for `LocalVideo` as well showing a black screen.
+A `PreviewVideo` is typically used on a pre-meeting view, or a device settings page. It doesn not effect the local video state.
```
devalevenkatesh, line 10: nit: Second sentence here:
"It does not affect the local video state."

src/hooks/sdk/useSelectAudioInputDevice.tsx
```diff
@@ -6,11 +6,13 @@ import { useCallback } from 'react';
 
 import { useMeetingManager } from '../../providers/MeetingProvider';
 
-export const useSelectAudioInputDevice = () => {
+export const useSelectAudioInputDevice = (): ((
+  device: Device | AudioTransformDevice | null
```
devalevenkatesh, line 10: In the component specific states I see we have used `undefined` and we take in `null` as well is it due to the fact that JS SDK may return a `undefined` device but can take in `null` device for selection purpose?

What will be the state for the device if we select a `null` device, an empty audio input device?

src/hooks/sdk/useSelectAudioOutputDevice.tsx
devalevenkatesh: I meant, we can still keep the util and move `new DefaultBrowserBehavior().supportsSetSinkId()` under it over writing the whole again each time we want to check, no strong opinion though, up to you.

src/hooks/sdk/useSelectVideoInputDevice.tsx
devalevenkatesh: I think `unselectVideoInputDevice` would not be good as it seems from JS SDK V3 we just start stop the video input device and allow choosing only a valid input device.

src/providers/LocalVideoProvider/docs/LocalVideoProvider.stories.mdx
```diff
@@ -2,31 +2,23 @@
...
-
-The React SDK depends on the Amazon Chime SDK for JavaScript, wherein, a single video stream is attached to a video element at a time. Hence, if `PreviewVideo` and `LocalVideo` component both are rendered on the same page and if both are enabled, stopping the video preview in `PreviewVideo` component will result into disconnecting the video stream for `LocalVideo` as well showing a black screen.
-
-To start/stop a video preview, please check `PreviewVideo` component [documentation](/docs/sdk-components-previewvideo--page).
-To start/stop user's local video, use `toggleVideo` function from the `LocalVideoProvider`.
+The `LocalVideoProvider` provides `tileId` of video tile associated with the local video, `isVideoEnabled` specifying the local video state, `hasReachedVideoLimit` specifying whether or not video limit is reached, and `toggleVideo()` to toggle local video.
 
+The state of the local video and the state of the previewVideo are independent.
```
devalevenkatesh, line 7: May be add a newline after this line. Also, suggest using `LocalVideo` over "local video" and `PreviewVideo` over "previewVideo".

src/providers/LocalVideoProvider/docs/LocalVideoProvider.stories.mdx
```diff
@@ -2,31 +2,23 @@
...
 ## State
 
 ```javascript
 {
-  // The tile ID associated with the local user's video
+  // The tile ID associated with the local video
   tileId: null | number;
 
   // Whether or not the local user has video enabled
-  // This does not tell whether the video preview from PreviewVideo component is enabled or not.
```
devalevenkatesh, line 20: I think this comment still stands and is fine to keep it to avoid builder confusion again with PreviewVideo component.

src/providers/LocalVideoProvider/docs/useLocalVideo.stories.mdx
```diff
@@ -2,31 +2,24 @@
...
 
-The `useLocalVideo` hook returns video tile id associated with the local user's video, `isVideoEnabled` specifying the local video status, a function to set the value of `isVideoEnabled` and a function to toggle user's local video.
+The `useLocalVideo` hook returns `tileId` of video tile associated with the local video, `isVideoEnabled` specifying the local video state, `hasReachedVideoLimit` specifying whether or not video limit is reached, and `toggleVideo()` to toggle local video.
 
-The React SDK depends on the Amazon Chime SDK for JavaScript, wherein, a single video stream is attached to a video element at a time. Hence, if `PreviewVideo` and `LocalVideo` component both are rendered on the same page and if both are enabled, stopping the video preview in `PreviewVideo` component will result into disconnecting the video stream for `LocalVideo` as well showing a black screen.
-
-To start/stop a video preview, please check `PreviewVideo` component [documentation](/docs/sdk-components-previewvideo--page).
-To start/stop user's local video, use `toggleVideo` function from the `LocalVideoProvider`.
+The state of the local video and the state of the previewVideo are independent.
```
devalevenkatesh, line 7: Copying: https://github.com/aws/amazon-chime-sdk-component-library-react/pull/773/files#r846297532

src/providers/LocalVideoProvider/index.tsx
```diff
@@ -68,7 +68,7 @@ const LocalVideoProvider: React.FC = ({ children }) => {
       audioVideo?.stopLocalVideoTile();
```
devalevenkatesh, line 68: Checking if we need to call `stopVideoInput` and then `stopLocalVideoTile` or just `stopLocalVideoTile` is fine?

src/components/migrationToV3.stories.mdx
xuesichao: The comment is correct. The code is wrong. It should be 
```jsx
<CameraSelection appendSampleDevices={false} />
```

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -248,11 +244,10 @@ export class MeetingManager implements AudioVideoObserver {
       this.audioVideo.unbindAudioElement();
 
       try {
-        await this.audioVideo.chooseVideoInputDevice(null);
-        await this.audioVideo.chooseAudioInputDevice(null);
-        await this.audioVideo.chooseAudioOutputDevice(null);
+        this.meetingSession?.deviceController.destroy();
```
devalevenkatesh, line 247: `await` destroy here?

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -583,25 +569,21 @@ export class MeetingManager implements AudioVideoObserver {
...
   ): Promise<void> => {
-    let receivedDevice = device;
-    if (typeof device === 'string') {
-      receivedDevice = videoInputSelectionToDevice(device);
-    }
-    if (receivedDevice === null) {
+    if (device === null) {
       try {
-        await this.audioVideo?.chooseVideoInputDevice(null);
+        await this.audioVideo?.stopVideoInput();
```
devalevenkatesh, line 576: I see the use now for `stopVideoInput`, may be I have commented somewhere in `LocalVideoProvider` about its usage.

src/components/sdk/DeviceSelection/SpeakerSelection/index.tsx
xuesichao: I have fixed this in my latest commit, so it will come with next PR. Would you mind not fixing it in this PR?

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
xuesichao: Removing this will lead to type error. Have to wrap them as a single child.
```jsx
<>
  {isLoading && <Spinner width="1.5rem" height="1.5rem" />}
  {isVoiceFocusEnabled ? voiceFocusOnLabel : voiceFocusOffLabel}
</>
```

src/hooks/sdk/useSelectAudioInputDevice.tsx
xuesichao: So in my later change. The the value of selected device on mount/unmount will be `undefined`. However in JS SDK the `startAudioInput(device: AudioInputDevice)` supports `null` which stands for the empty/fake audio input device. So have to keep it only for audio input device. This will come with my next PR.

src/components/sdk/DeviceSelection/SpeakerSelection/index.tsx
devalevenkatesh: Np at All!

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: That comment is correct and helps us find a bug in JS SDK.

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: Nice catch, updated!

src/providers/LocalVideoProvider/index.tsx
xuesichao: Confirmed with Trung that we should call `stopVideoInput` instead of `stopLocalVideoTile`. In JS SDK `stopVideoInput` kills stream and also stops local video tile. No need to call both of them.

So we find a bug in JS SDK. Currently we just call `stopLocalVideoTile()` here and it's supposed not to kill the media stream, but it does now.That's why I didn't find this code is wrong during my testing.

Updated.

src/providers/LocalVideoProvider/docs/useLocalVideo.stories.mdx
xuesichao: Updated.

src/providers/LocalVideoProvider/docs/LocalVideoProvider.stories.mdx
xuesichao: Added them back.

src/providers/LocalVideoProvider/docs/LocalVideoProvider.stories.mdx
xuesichao: This is something I want to do as well, updated.

src/components/sdk/PreviewVideo/PreviewVideo.stories.mdx
xuesichao: Updated both.

Use `LocalVideo` instead of `<LocalVideo />` to keep consistency.

src/components/sdk/DeviceSelection/SpeakerSelection/index.tsx
xuesichao: Thanks, but it's a small fix, so I updated it : )

src/components/migrationToV3.stories.mdx
xuesichao: Updated all markdown language.
So this is only for syntax highlighting, either `jsx` or `tsx` works. I kinda prefer to use `tsx` for all. But it also makes sense to use the accurate one.

src/components/migrationToV3.stories.mdx
xuesichao: Updated all markdown language.
So this is only for syntax highlighting, either `jsx` or `tsx` works. I kinda prefer to use `tsx` for all. But it also makes sense to use the accurate one.

src/components/migrationToV3.stories.mdx
xuesichao: Updated

src/components/migrationToV3.stories.mdx
xuesichao: Removed.

src/components/sdk/PreviewVideo/PreviewVideo.stories.mdx
```diff
@@ -7,7 +7,7 @@ import { PreviewVideo } from './';
 
 The `PreviewVideo` component renders a preview of the local video. If there is no available video input, the video will be blank. 
 
-A `PreviewVideo` is typically used on a pre-meeting view, or a device settings page. It doesn not effect the local video state.
+A `PreviewVideo` is typically used on a pre-meeting view, or a device settings page. It doesn not effect the `LocalVideo` component state.
```
devalevenkatesh, line 10: nit: Typo "doesn" -> "does"

xuesichao: Updated.

devalevenkatesh (dismissed):

src/components/sdk/MeetingControls/AudioInputControl.tsx
```diff
@@ -22,30 +21,31 @@ interface Props extends BaseSdkProps {
...
 const AudioInputControl: React.FC<Props> = ({
   muteLabel = 'Mute',
   unmuteLabel = 'Unmute',
   mutedIconTitle,
   unmutedIconTitle,
-  appendSampleDevices = true,
   ...rest
 }) => {
-  const meetingManager = useMeetingManager();
+  const selectAudioInput = useSelectAudioInputDevice();
```
ltrung, line 33: what is the difference between `AudioInputControl` and `MicSelection`?

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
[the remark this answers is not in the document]
ltrung: Can we put a comment on why we can't remove it?

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
```diff
@@ -60,80 +59,93 @@ const VideoInputBackgroundBlurControl: React.FC<Props> = ({
...
-  }
+  };
+
+  const handleClick = async (deviceId: string): Promise<void> => {
+    try {
+      // If background blur is on, then re-use the same video transform pipeline, but replace the inner device
+      // If background blur is not on, then do a normal video selection
+      if (isVideoTransformDevice(activeVideoDevice) && !isLoading) {
+        setIsLoading(true);
+        if ('chooseNewInnerDevice' in activeVideoDevice) {
```
ltrung: Don't think we need to check for it

src/components/sdk/PreviewVideo/index.tsx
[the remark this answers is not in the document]
ltrung: We may need to discuss this tomorrow. Kinda weird that start will start the video input stream but stop just detach the video element. If we are out of time, better to keep the same behavior and call `stopVideoInput` manually.

src/constants/index.ts
```diff
@@ -11,6 +11,6 @@ export const KEY_CODES = {
 
 export const VIDEO_INPUT_QUALITY = {
   '360p': '360p (nHD) @ 15 fps (600 Kbps max)',
-  '540p': '540p (qHD) @ 15 fps (1.4 Mbps max)',
+  '540p': '540p (qHD) @ 15 fps (1.0 Mbps max)',
```
ltrung, line 14: Why do we change this?

src/hooks/sdk/useSelectAudioInputDevice.tsx
ltrung: isn't this the same as `AudioInputDevice` from JS SDK? Why dont we just use it?

src/components/sdk/MeetingControls/AudioInputControl.tsx
xuesichao: `MicSelection` is the one in Device Setup page.
`AudioInputControl` is the one in Meeting Control Bar.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -277,18 +247,19 @@ export class MeetingManager implements AudioVideoObserver {
...
-          name,
-          meetingSessionConfiguration,
-          batchSize,
-          intervalMs,
-          url,
-          logLevel
-        )
-      );
+      logger = new MultiLogger(consoleLogger);
+      // const { name, batchSize, intervalMs, url, logLevel } = postLoggerConfig;
```
ltrung, line 251: Assume this will be removed once merged with Venky's PR.

src/components/sdk/MeetingControls/AudioInputControl.tsx
ltrung: Yeah but can't tell from that name and the code is almost similar. So not sure why we need both (actually 3, `MicSelection`, `AudioInputControl`, and `AudioInputVFControl`.

src/components/sdk/DeviceSelection/CameraSelection/BackgroundBlurCheckbox.tsx
```diff
@@ -27,41 +28,51 @@ export const BackgroundBlurCheckbox: React.FC<Props> = ({
...
-          'Background blur processor is not supported yet.'
-        );
-      }
-    } else {
-      if ('chooseNewInnerDevice' in device) {
-        setIsLoading(true);
-        // @ts-ignore
-        const deviceId = await device.intrinsicDevice();
-        await meetingManager.selectVideoInputDevice(deviceId);
+        if ('chooseNewInnerDevice' in device) {
```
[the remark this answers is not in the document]
xuesichao: I don't think it's needed. I was afraid to break something so I didn't change it. Technically I think we can remove this.

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: Yes

src/constants/index.ts
xuesichao: I thought this was a typo. Cause in JS SDK migration guide, the 540p's maximum bandwidth is 1.0 Mbps. I can revert this if it's not needed. Seems to be weird that 540p and 720p requires same maximum bandwidth.

src/hooks/sdk/useSelectAudioInputDevice.tsx
xuesichao: We should. I missed this.

src/constants/index.ts
ltrung: Oh I just chose 1000 as an example because the old `ChooseVideoinputQuality` has that parameter so i just chose a random value. 540p and 720p don't require the same max bandwidth but we don't need to change that behavior (as it is not in scope for this PR).

src/hooks/sdk/useSelectVideoQuality.tsx
```diff
@@ -20,13 +20,16 @@ export function useSelectVideoQuality() {
...
       switch (quality) {
         case '360p':
-          audioVideo.chooseVideoInputQuality(640, 360, 15, 600);
+          audioVideo.chooseVideoInputQuality(640, 360, 15);
+          audioVideo.setVideoMaxBandwidthKbps(600);
           break;
         case '540p':
-          audioVideo.chooseVideoInputQuality(960, 540, 15, 1400);
+          audioVideo.chooseVideoInputQuality(960, 540, 15);
+          audioVideo.setVideoMaxBandwidthKbps(1000);
```
ltrung, line 28: You do not need to call `audioVideo.setVideoMaxBandwidthKbps(1400);` as it is the default value.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -248,11 +217,12 @@ export class MeetingManager implements AudioVideoObserver {
       this.audioVideo.unbindAudioElement();
 
       try {
-        await this.audioVideo.chooseVideoInputDevice(null);
-        await this.audioVideo.chooseAudioInputDevice(null);
-        await this.audioVideo.chooseAudioOutputDevice(null);
+        await this.meetingSession?.deviceController.destroy();
+        await this.audioVideo.chooseAudioOutput(null);
```
ltrung, line 221: Can we reverse the order here? Weird to call destroy then try to call `chooseAudioOutput`.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -721,77 +639,35 @@ export class MeetingManager implements AudioVideoObserver {
...
 
-  private publishSelectedAudioInputTransformDevice = (): void => {
-    for (const observer of this.selectedAudioInputTransformDeviceObservers) {
-      observer(this.selectedAudioInputTransformDevice);
-    }
-  };
-
   subscribeToSelectedAudioOutputDevice = (
-    callback: (deviceId: string | null) => void
+    callback: (deviceId: string | undefined) => void
```
ltrung, line 664: In JS SDK, it is `string | null`. The default value for audio output device is `null` or default of browser. Don't think we have `undefined` value for output.

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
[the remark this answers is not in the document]
xuesichao: If we remove it will be a type error.

```
Type 'string | false | Element' is not assignable to type 'ReactElement<any, string | JSXElementConstructor<any>>'.
Type 'string' is not assignable to type 'ReactElement<any, string | JSXElementConstructor<any>>'
```

```tsx
children?: React.ReactElement<any> | React.ReactElement<any>[];
```

src/hooks/sdk/useSelectAudioOutputDevice.tsx
```diff
@@ -1,17 +1,23 @@
...
       await meetingManager.selectAudioOutputDevice(deviceId);
+    } else {
+      console.error(
```
ltrung, line 18: Should we throw error instead?

src/hooks/sdk/useSelectAudioInputDevice.tsx
xuesichao: I missed it in the second commit but updated it in the third commit. No change is required.

src/hooks/sdk/useSelectVideoQuality.tsx
xuesichao: I think it's needed here.
If user chooses 360p first, the `VideoMaxBandwidthKbps` is set to 600.
if user chooses 540p then, the `VideoMaxBandwidthKbps` is not updated and still 600.

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: `undefined` is not a value for any actual device selection. It's the value before component mount or after component unmount. Will change it to `deviceId: string | null | undefined`

src/hooks/sdk/useSelectAudioOutputDevice.tsx
xuesichao: This change is just to add some more logs. Don't think we need to throw this error.

src/providers/MeetingProvider/MeetingManager.ts
ltrung: `undefined` does not make sense. No selection means `null` as we just use whatever the browser used. (i.e., if you do not make any explicit selection, the value is `null` and not `undefined`.

src/hooks/sdk/useSelectVideoQuality.tsx
ltrung: Oops. Need to fix this in JS SDK demo then :D

src/hooks/sdk/useSelectAudioOutputDevice.tsx
xuesichao: Notice that `chooseAudioOutput` API will throw an error if browser does not support setSinkId. If so the error will be threw in `MeetingManger`.

src/components/sdk/DeviceSelection/CameraSelection/BackgroundBlurCheckbox.tsx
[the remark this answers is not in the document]
xuesichao: You are checking the second commit. In the third commit I remvoed this.

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
xuesichao: So `chooseNewInnerDevice` is only available for `VoiceFocusTransformDevice`. I don't know why they check this for a video input device.
I think this is wrong here. and this if condition can never be passed.

xuesichao: `chooseNewInnerDevice` is only available in `DefaultVideoTransformDevice`. This is to check if the `activeVideoDevice` is a `DefaultVideoTransformDevice`. So it's needed.

src/components/sdk/MeetingControls/AudioInputControl.tsx
xuesichao: They are different UI.
`MicSelection` is `DeviceInput`(`FormField`)
`AudioInputControl` is` ControlBarButton`

src/constants/index.ts
xuesichao: Updated.

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: Updated to:
```js
await this.meetingSession?.deviceController.chooseAudioOutput(null);
await this.meetingSession?.deviceController.destroy();
```

src/providers/MeetingProvider/MeetingManager.ts
xuesichao: Updated to `string | null` for `selectedAudioOutputDevice`

src/components/sdk/PreviewVideo/index.tsx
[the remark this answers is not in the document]
xuesichao: I think we need to update the description for `stopVideoPreviewForVideoInput` API and migration guide.
> Stops the stream for a previously bound video preview and unbinds it from the video element.

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
ltrung: Man this should just be in the interface :)

ltrung: Can you create a SIM to track this in V4

src/components/sdk/DeviceSelection/DeviceInput.tsx
```diff
@@ -12,47 +18,58 @@ interface Props extends BaseSdkProps {
   label: string;
   notFoundMsg: string;
   devices: DeviceType[];
-  selectedDeviceId: SelectedDeviceId;
+  selectedDevice:
```
devalevenkatesh, line 21: Can this be `AudioInputDevice` | `VideoInputDevice` | `undefined` ? Is there a reason for separating the types here?

devalevenkatesh, line 21: Can this be `AudioInputDevice` | `VideoInputDevice` | `undefined` ? Is there a reason for separating the types here?

src/components/sdk/DeviceSelection/DeviceInput.tsx
```diff
@@ -12,47 +18,58 @@ interface Props extends BaseSdkProps {
...
+    | Device
+    | AudioTransformDevice
+    | VideoTransformDevice
+    | null
+    | undefined;
   onChange: (deviceId: string) => void;
```
devalevenkatesh, line 27: So for this `onChange` prop, should we now vend out the `selectedDevice` with type not only `string`?

devalevenkatesh, line 27: So for this `onChange` prop, should we now vend out the `selectedDevice` with type not only `string`?

devalevenkatesh: NVM, I think this `onChange` can only vend selected `string` from select element.

src/components/sdk/DeviceSelection/DeviceInput.tsx
xuesichao: The type is exactly same. The only reason I did not do that is because it is confusing.

`selectedDevice` also applies for `AudioOutputDevice`. If `selectedDevice` is type of `AudioInputDevice | VideoInputDevice | undefined`, builder might be curious where is `AudioOutputDevice`. And they have to check the JS SDK code and figure out that there's no type of `AudioOutputDevice` which is `string | null` and is covered by `AudioInputDevice | VideoInputDevice | undefined`.

 So it's for readability.

xuesichao: The type is exactly same. The only reason I did not do that is because it is confusing.

selectedDevice also applies for AudioOutputDevice. If selectedDevice is type of AudioInputDevice | VideoInputDevice | undefined, builder might be curious where is AudioOutputDevice. And they have to check the JS SDK code and figure out that there's no type of AudioOutputDevice which is string | null and is covered by AudioInputDevice | VideoInputDevice | undefined.

So it's for readability.

src/components/sdk/MeetingControls/AudioInputControl.tsx
```diff
@@ -33,20 +33,32 @@ const AudioInputControl: React.FC<Props> = ({
   const selectAudioInput = useSelectAudioInputDevice();
   const { muted, toggleMute } = useToggleLocalMute();
   const { devices, selectedDevice } = useAudioInputs();
+  const [dropdownOptions, setDropdownOptions] = useState<PopOverItemProps[]>(
+    []
+  );
+
+  useEffect(() => {
```
devalevenkatesh, line 40: I do not think you need a state or useEffect here. If `devices` or `selectedDevice` changes this component needs to re-render anyhow right? So, can we keep this as is like earlier?

src/components/sdk/DeviceSelection/DeviceInput.tsx
xuesichao: Yeah, can't do that because It is based on `FormField` and the value can only be `string`.

xuesichao: Yeah, can't do that because It is based on FormField and the value can only be string.

src/components/sdk/MeetingControls/AudioInputControl.tsx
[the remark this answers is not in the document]
xuesichao: No. `useEffect` is need here cause `isOptionActive` API is `async` now. And we can not use `async` function in React functional component.

xuesichao: No. useEffect is need here cause isOptionActive API is async now. And we can not use async function in React functional component.

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
```diff
@@ -128,47 +115,55 @@ const AudioInputVFControl: React.FC<Props> = ({
         console.error(
           'AudioInputVFControl failed to select audio input device'
         );
+      } finally {
+        setIsLoading(false);
```
devalevenkatesh, line 119: I think we dont want any `isLoading` state update changes in the `else` block under try. May be just keep it as is, not sure yet if that impacts anything.

devalevenkatesh, line 119: I think we dont want any `isLoading` state update changes in the `else` block under try. May be just keep it as is, not sure yet if that impacts anything.

src/components/sdk/MeetingControls/AudioOutputControl.tsx
```diff
@@ -26,28 +25,40 @@ const AudioOutputControl: React.FC<Props> = ({
...
+  const [dropdownOptions, setDropdownOptions] = useState<PopOverItemProps[]>(
+    []
+  );
 
-  const handleClick = async (deviceId: string): Promise<void> => {
-    try {
-      if (new DefaultBrowserBehavior().supportsSetSinkId()) {
-        await selectAudioOutput(deviceId);
-      } else {
+  useEffect(() => {
```
devalevenkatesh, line 32: Any reason for this change? I think commented somewhere on AudioInputControl as well.

devalevenkatesh, line 32: Any reason for this change? I think commented somewhere on AudioInputControl as well.

xuesichao: Replied to that comment. There is a TODO item to remove `supportsSetSinkId()` from device-utils and use `new DefaultBrowserBehavior().supportsSetSinkId()` instead.

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
```diff
@@ -93,67 +75,74 @@ const VideoInputBackgroundBlurControl: React.FC<Props> = ({
...
-      // If background blur is not on, then do a normal video selection
-      if (isVideoTransformDevice(activeVideoDevice) && !isLoading) {
-        setIsLoading(true);
-        if ('chooseNewInnerDevice' in activeVideoDevice) {
-          const transformedDevice =
-            // @ts-ignore
-            activeVideoDevice.chooseNewInnerDevice(deviceId);
-          await selectVideoinput(transformedDevice);
+  useEffect(() => {
+    const handleClick = async (deviceId: string): Promise<void> => {
```
devalevenkatesh, line 84: Now that you have moved `handleClick` inside `useEffect`, you have to also add methods or variables used in `handleClick` to dependency array. For e.g.: `selectVideoinput`

devalevenkatesh, line 84: Now that you have moved `handleClick` inside `useEffect`, you have to also add methods or variables used in `handleClick` to dependency array. For e.g.: `selectVideoinput`

src/components/sdk/MeetingControls/AudioInputVFControl.tsx
xuesichao: Nope. if `await selectAudioInput(currentDevice);` throws the error `setIsLoading(false);` can not be reached and thus add it to `finally`.

Maybe I can remove line 110 to avoid duplicate, how do you think?

xuesichao: Nope. if `await selectAudioInput(currentDevice);` throws the error `setIsLoading(false);` can not be reached and thus add it to `finally`.

Maybe I can remove line 110 to avoid duplicate, how do you think?

src/hooks/sdk/useSelectAudioInputDevice.tsx
```diff
@@ -1,22 +1,19 @@
...
 
-  const selectDevice = useCallback(
-    async (device: Device | AudioTransformDevice | null) => {
-      await meetingManager.selectAudioInputDevice(device);
-    },
-    []
-  );
+  const selectDevice = useCallback(async (device: AudioInputDevice) => {
+    await meetingManager.selectAudioInputDevice(device);
+  }, []);
```
devalevenkatesh, line 16: Can you add meetingManager in the dependency array here? Also same for useSelectVideoInputDevice

devalevenkatesh, line 16: Can you add meetingManager in the dependency array here? Also same for useSelectVideoInputDevice

devalevenkatesh: I know we are missing it but we should add it actually to keep the effect updated with meetingManager if it changes.

devalevenkatesh: I know we are missing it but we should add it actually to keep the effect updated with meetingManager if it changes.

src/components/sdk/MeetingControls/AudioInputControl.tsx
devalevenkatesh: Also, with moving this handleClick, we have to update the dependency array with the variables, methods such as selectAudioInput. I wish we enabled the exhaustive-deps rule and solve this for us.

devalevenkatesh: Also, with moving this handleClick, we have to update the dependency array with the variables, methods such as selectAudioInput. I wish we enabled the exhaustive-deps rule and solve this for us.

src/components/sdk/PreviewVideo/index.tsx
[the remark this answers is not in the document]
xuesichao: Discussed offline we will revert this change.

src/components/sdk/MeetingControls/AudioInputControl.tsx
xuesichao: updated dependencies.

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
xuesichao: updated

src/hooks/sdk/useSelectAudioInputDevice.tsx
xuesichao: good catch! updated.

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
```diff
@@ -47,101 +40,110 @@ const VideoInputBackgroundBlurControl: React.FC<Props> = ({
...
+        deviceOptions.push(videoTransformOptions);
+      }
+      setDropdownWithVideoTransformOptions(deviceOptions);
+    };
+
+    getDropdownWithVideoTransformOptions();
   }, [
     createBackgroundBlurDevice,
-    activeVideoDevice,
+    selectVideoinput,
```
ltrung, line 145: typo: selectVideoInput?

xuesichao: It is `selectVideo` input.
```js
const selectVideoinput = useSelectVideoInputDevice();
```

Will remove these `useSelect*Device` hooks and updated to `meetingManager.start*Device` instead in a later PR.

devalevenkatesh (approved):

ltrung (dismissed):

src/components/sdk/MeetingControls/VideoInputBackgroundBlurControl.tsx
xuesichao: Updated.

ltrung (approved):

PR #842, opened 2022-10-19
Add observer prop to the BG Blur and Replacement Provider

michhyun1:
**Issue #:** 

**Description of changes:**
Add an observer prop to the BG Blur and Replacement provider  to add and remove the observer to the initialized processor on mount/umount.

**Testing**
yes

3. How did you test these changes?
Add a prop to the Background Blur Provider in the meeting demo and confirm that observer functions are executed. (Just console log when the events fired)

4. If you made changes to the component library, have you provided corresponding documentation changes?
Yes

### DIFF
--- a/src/providers/BackgroundBlurProvider/index.tsx
+++ b/src/providers/BackgroundBlurProvider/index.tsx
@@ -5,13 +5,13 @@ import {
   BackgroundBlurOptions,
   BackgroundBlurProcessor,
   BackgroundBlurVideoFrameProcessor,
+  BackgroundBlurVideoFrameProcessorObserver,
   BackgroundFilterSpec,
   ConsoleLogger,
   DefaultVideoTransformDevice,
   Device,
   LogLevel,
   NoOpVideoFrameProcessor,
-  VideoFrameProcessor,
 } from 'amazon-chime-sdk-js';
 import React, {
   createContext,
@@ -33,6 +33,10 @@ interface Props extends BaseSdkProps {
   /** A set of options that can be supplied when creating a background blur video frame processor. For more information, refer to
    * [Amazon Chime SDK for JavaScript Background Filter Guide](https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/15_Background_Filter_Video_Processor.md#adding-a-background-filter-to-your-application). */
   options?: BackgroundBlurOptions;
+  /**
+   * Observer callback functions. The observer will be added to the background blur processor on mount and removed on unmount.
+   */
+  observer?: BackgroundBlurVideoFrameProcessorObserver;
 }
 
 interface BackgroundBlurProviderState {
@@ -46,12 +50,12 @@ const BackgroundBlurProviderContext = createContext<
   BackgroundBlurProviderState | undefined
 >(undefined);
 
-const BackgroundBlurProvider: FC<Props> = ({ spec, options, children }) => {
+const BackgroundBlurProvider: FC<Props> = ({ spec, options, observer, children }) => {
   const logger = useLogger();
   const [isBackgroundBlurSupported, setIsBackgroundBlurSupported] = useState<
     boolean | undefined
   >(undefined);
-  const [processor, setProcessor] = useState<VideoFrameProcessor | undefined>();
+  const [processor, setProcessor] = useState<BackgroundBlurProcessor | undefined>();
 
   const blurSpec = useMemoCompare(
     spec,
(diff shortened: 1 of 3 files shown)


richhx (dismissed):
Thanks!

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -104,6 +111,18 @@ const BackgroundReplacementProvider: FC<Props> = ({
...
+    if (processor && observer) {
+      (processor as BackgroundReplacementProcessor).addObserver(observer);
+    }
+    
+    return () => {
+      if (processor && observer) {
+        (processor as BackgroundReplacementProcessor).removeObserver(observer);
+      }
+    };
+  }, [observer]);
```
devalevenkatesh, line 124: Should we add `processor` dependency as well?

michhyun1: Yes, good catch. When a new processor is intitialized, we want to add the observers to that also.

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -104,6 +111,18 @@ const BackgroundReplacementProvider: FC<Props> = ({
     };
   }, [replacementSpec, replacementOptions]);
 
+  useEffect(() => {
+    if (processor && observer) {
+      (processor as BackgroundReplacementProcessor).addObserver(observer);
```
devalevenkatesh, line 116: Would it be better to update the State type itself based on usage? Believe the processor in this provider will always be `BackgroundReplacementProcessor` and it extends `VideoFrameProcessor`.
https://github.com/aws/amazon-chime-sdk-js/blob/main/src/backgroundreplacementprocessor/BackgroundReplacementProcessor.ts#L10

I think we can do that and is better over typecasting. If `VideoFrameProcessor` state type is needed then here instead of casting should we check for `instanceof` to make sure `processor` here is actually an instance of `BackgroundReplacementProcessor`. What say?

src/providers/BackgroundReplacementProvider/index.tsx
devalevenkatesh: Probably missed adding here yet :)

src/providers/BackgroundReplacementProvider/index.tsx
michhyun1: I tried instanceof, but I guess because BackgroundReplacementProcessor is an interface, it doesnt work.

src/providers/BackgroundReplacementProvider/index.tsx
michhyun1: ops thanks

richhx (approved):

src/providers/BackgroundReplacementProvider/index.tsx
richhx: That's actually a great catch

src/providers/BackgroundReplacementProvider/index.tsx
devalevenkatesh: Ohh I see, missed the interface and took it for class, my bad. Well, in that case should we update the type from `VideoFrameProcessor` to `BackgroundReplacementProcessor`? The reason is say if any other type of processor is passed and at runtime we cannot cast it then that might be an error.

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -88,7 +93,11 @@ const BackgroundBlurProvider: FC<Props> = ({ spec, options, children }) => {
   );
 
   useEffect(() => {
-    initializeBackgroundBlur();
+    initializeBackgroundBlur().then((createdProcessor) => {
+      if (observer) {
+        (createdProcessor as BackgroundBlurProcessor).addObserver(observer);
```
devalevenkatesh, line 98: Do we have to do this? I think the `useEffect` will run based on `processor` state update and should add the observer. Could you verify once?

Also suggest adding log statements if that helps in debugging.

michhyun1: Yeah, thanks for catching that. Now that we have the processor in that useEffect, we wont need this.

src/providers/BackgroundReplacementProvider/index.tsx
michhyun1: done

devalevenkatesh (approved):
Can follow up with the use cases in documentation on how this benefits for both providers or can add in same CR up to you :)

src/providers/BackgroundReplacementProvider/index.tsx
devalevenkatesh: Just to note, I dont think this is a breaking change as we are actually expanding the types and not shrinking it so believe we should be good on that.

Issue #847, opened 2022-10-27
[FEATURE REQUEST] Add Node.js LTS (18.12.0) support


hoolek77:
## Community Note

* Please do not add additional comments such as "+1" or "me too" to existing issues - instead please vote on the issue by adding a 👍 to the issue. The Amazon Chime SDK team will prioritize the request based on the number of 👍 the issue has received.
* Please go through the existing [issues](https://github.com/aws/amazon-chime-sdk-component-library-react/issues) to see if your request has already been recommended.
* Please do not use this template to report bugs. This repository has an [issue template](https://github.com/aws/amazon-chime-sdk-component-library-react/issues/new/choose) for issues and [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/docs/quick-starts--page) to help with your development.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

## Tell us about your request

The package is missing support for the current Node.js LTS version, which recently changed to 18.12.0.

### What do you want us to build?

Add support for Node engine ^18.

### How are you currently solving a problem?

Using previous LTS engine.

manojkalpana24:
Kindly add support for Node.js LTS 18.12.0. Not able to build react application right now, the following message is thrown while executing yarn install,
error amazon-chime-sdk-component-library-react@3.2.0: The engine "node" is incompatible with this module. Expected version "^12 || ^14 || ^15 || ^16". Got "18.12.0"
error Found incompatible module.

xuesichao:
PR for adding Node.js 18 support has been merged: #848 
Please let us know if you have any feedback. Thanks!

manojkalpana24:
> PR for adding Node.js 18 support has been merged: #848 Please let us know if you have any feedback. Thanks!

When this would be released ?

PR #848, opened 2022-10-28
Add support for Node.js 18

xuesichao:
**Issue #847 :** 

**Description of changes:**
* Add support to Node.js 18
* Print log when running `npm run build:release` without a new commit to reduce confusion.
* Run `npm upgrade source-map` to fix this issue: https://github.com/mozilla/source-map/issues/454

**Testing**
Yes

Running scripts in `package.json` using node 18.9.0.

N/A


simmkyu (approved):

scripts/prebuild.js
```diff
@@ -36,6 +36,12 @@ if (!process.env.GITHUB_ACTIONS) {
 }
 
 let commit_files = [];
+
+if (commits.length === 0) {
+  logger.log(`Didn't find any new commit. Exiting process.`);
+  return process.exit(1);
```
devalevenkatesh, line 42: Want to exit with 0? Since it is not actually an error or do we want to treat this as error, I guess not :)

xuesichao: We have to use `exit(0)` because `npm run build:release` actually include these steps:
1. prebuild
2. build
3. postbuild

So if we exit(0) in step 1, step 2 and 3 will still be executed. So have to use `exit(1` to terminate the process.

xuesichao: The difference between `npm run build` and `npm run build:release` is that the latter has `prebuild` script to check if there’s a `CHANGELOG` change included in the commit and `postbuild` scripts checks if there’s any uncommitted files after running test.
So if there’s no commit, it doesn’t make sense to run this script. We can manually run `npm run build` and `npm run test` instead.

devalevenkatesh: Yup makes sense can go ahead and merge.

PR #865, opened 2023-01-09
Revert changes to Background Blur and Background Replacement to fix bug

michhyun1:
**Issue #:** 

**Description of changes:**
Need to revert the changes from the previous commit - I removed this unnecessary initalizing of the Background blur and Background replacement on the mounting of the provider, but it turns out that we need to call `initializeBackgroundBlur`  or `initializeBackgroundReplacement` in order to initialize the `isBackgroundBlurSupported` or `isBackgroundReplacementSupported` state of the corresponding providers. Otherwise, this could be a breaking change, as the initial value of the `isBackgroundBlurSupported` or `isBackgroundReplacementSupported` for the providers will always be false, when in reality it should be true.

**Testing**
yes

Tested on the react demo - make sure you can enable background blur and background replacement.

Not needed, reverting changes. I've added comments to highlight the importance of initializing the background processors on the providers' mount event.

### DIFF
--- a/src/providers/BackgroundReplacementProvider/index.tsx
+++ b/src/providers/BackgroundReplacementProvider/index.tsx
@@ -95,9 +95,12 @@ const BackgroundReplacementProvider: FC<Props> = ({
   );
 
   useEffect(() => {
+    // One reason we need to initialize first, even though we'll destroy this background replacement processor when we create a new device
+    // is because we need to check if background replacement is supported by initializing the background replacement processor to see if the browser supports
+    initializeBackgroundReplacement();
     return () => {
       logger.info(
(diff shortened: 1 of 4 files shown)


src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -95,9 +95,12 @@ const BackgroundReplacementProvider: FC<Props> = ({
   );
 
   useEffect(() => {
+        // One reason we need to initialize first, even though we'll destroy this background replacement processor when we create a new device
```
xuesichao, line 98: need to fix this indention.

xuesichao (approved):

Issue #864, opened 2023-01-07
onDeviceReplacement don't work well in the macbook


andreMycroft:
### What happened and what did you expect to happen?

When I turn of my Wifi for some reason the onDeviceReplacement it's called, maybe because he lost the connection with the deviceList of the macbook.
So this way when i connect back he get's the next device Normally the default.
The big problem is, If my DeviceLabel it's None, he ignore that and turn on my audio.
So i expect that the lib check if the current Attendee have permissions to use audio before do that.

### Have you reviewed our existing documentation?

- [X] Amazon Chime SDK for JavaScript GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] Amazon Chime SDK React Components Library GitHub [issues](https://github.com/aws/amazon-chime-sdk-component-library-react/issues)
- [X] Storybook [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/introduction--page)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-component-library-react#readme)
- [X] How-to [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/how-tos--page)
- [X] React SDK meeting [demo](https://github.com/aws-samples/amazon-chime-sdk/tree/main/apps/meeting#readme---react-meeting-demo)

### Reproduction steps

Join a Meeting with Two Attendee without audio `DeviceLabel.None` 
Turn off the wifi on Macbook
Turn On the Wifi on Macbook

### Amazon Chime SDK React Components Library version

3.3.0

### What browsers are you seeing the problem on?

Google Chrome

### Browser version

Version 108.0.5359.124 (Official Build) (arm64)

### Device Information

Macbook M1 2020

### Important now it's setting `Null` because i change the `onDeviceReplacement` to allways return null.

2023-01-07T22:18:47.790Z [INFO] ChimeSDKReactComponent - Device change event callback is triggered
ConsoleLogger.js:78 2023-01-07T22:18:47.829Z [INFO] ChimeSDKReactComponent - AudioInputProvider - audio inputs updated
ConsoleLogger.js:78 2023-01-07T22:18:47.832Z [INFO] ChimeSDKReactComponent - API/DefaultAudioVideoFacade/6a3741ea-2e57-4526-9d43-9fbf547a0706/7afb821f-8913-ce8d-efa9-ad45f71dd8e8/startAudioInput null
[...]

PR #877, opened 2023-03-20
Do not auto-select audio input device on audio inputs change when DeviceLabels are not explicitly enabled.

devalevenkatesh:
Do not auto-select audio input device when audio inputs change when DeviceLabels are not explicitly enabled.

- Currently, we do not store DeviceLabels anywhere once set through or passed through either meetingManager.invokeDeviceProvider or through meetingManager.join itself.
- Add a ESLint plugin for react hook rules and dependency checks following: https://legacy.reactjs.org/docs/hooks-rules.html#eslint-plugin

Fixes issue:
- https://github.com/aws/amazon-chime-sdk-component-library-react/issues/864

**Issue #:** 

**Description of changes:**

- This change keeps track of DeviceLabels in MeetingManager and checks them to decide whether to auto select the audio input when audio inputs change.
- Fix roster PR check run by fixing the `checkIfPortIsInUse` script while running the test. There was a race condition + incorrect hostname in the method. This may be running into issue now either due to GitHub env changes or browser version changes.
- Fix content share test failure by re-ordering `chromeOptions` args. Arg `--use-fake-device-for-media-stream` must be before `--use-fake-ui-for-media-stream` for test to pass. Found that when comparing the test runs with JS SDK demo.

**Testing**

   1. Packed the changed component library into demo meeting app under Amazon Chime SDK samples.
   1. Run the demo
   1. Join the meeting with no A/V permission
   1. Verified mid meeting audio input device change does not audio input selection.

Same for with A/V permissions, verified that the audio input was selected in that case.


src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -71,6 +83,13 @@ const AudioInputProvider: React.FC<Props> = ({
       audioInputsChanged: async (newAudioInputs: MediaDeviceInfo[]) => {
         logger.info('AudioInputProvider - audio inputs updated');
 
+        if (!shouldAutoSelectAudioInputDevice()) {
```
simmkyu, line 86: Can you check if you get any use-hook dependency warning from the Terminal?

src/providers/DevicesProvider/docs/AudioInputProvider.stories.mdx
```diff
@@ -20,14 +20,20 @@ The `AudioInputProvider` provides a list of the user's available audio input dev
...
 it would be better to use `devices`. When a new "default" device is added, the `devices` list will have changed. We suggest you to
 watch for changes to `devices[0]` - this will indicate when the "default" device has changed. `selectedDevice` will not be updated when
 switching to a new audio device since it stands for the deviceId, and the value of deviceId remains the same, still being `default`.
 The deviceId that was previously in use was `default`. When you switch to a new device, that device becomes "default" device. Since the old
 deviceId was `default` and the new deviceId is still `default`, so `selectedDevice` does not change. But actually they correspond to different physical devices.
 
+2. If you are joining a meeting with `DeviceLabels.None` or overriding the default device label trigger function (Check "Get Device Permissions From the Browser" under the [quickstarts](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/quick-starts--page) for more information),
+then, the `AudioInputProvider` will not auto select a audio input device when audio input device list changes. To keep the auto-selection of audio input deviceon audio inputs change,
```
simmkyu, line 34: nit:
- `will not auto select a audio input device` => `will not automatically select an audio input device`
- `To keep the auto-selection of audio input deviceon audio inputs change,` => `To keep the auto-selection of the audio input device on audio input changes,`

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -108,6 +108,12 @@ export class MeetingManager implements AudioVideoObserver {
 
   private eventDidReceiveRef: EventObserver;
 
+  private _deviceLabels: DeviceLabels | DeviceLabelTrigger;
+  
+  public get deviceLabels() : DeviceLabels | DeviceLabelTrigger {
```
simmkyu, line 113: nit: The component library does not appear to use the getter, except `Versioning.ts`. How about making `getDeviceLabels()`?

src/providers/DevicesProvider/AudioInputProvider.tsx
devalevenkatesh: I checked and updated our eslint with `eslint-plugin-react-hooks` with hook related rules as suggested here and updated the `AudioInputProvider` according to warnings. I did not see though warning for use-hook specifically for `shouldAutoSelectAudioInputDevice` in local terminal + browser console.

https://legacy.reactjs.org/docs/hooks-rules.html#eslint-plugin

src/providers/DevicesProvider/docs/AudioInputProvider.stories.mdx
devalevenkatesh: Thanks, fixed it 🙏

src/providers/MeetingProvider/MeetingManager.ts
devalevenkatesh: Sure, updated to use `getDeviceLabels` over `getter`.

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -42,15 +43,28 @@ const AudioInputProvider: React.FC<Props> = ({
...
+      }
+      return device;
+    },
+    [onDeviceReplacement, meetingManager.selectedAudioInputDevice]
+  );
+
+  const shouldAutoSelectAudioInputDevice = useCallback((): boolean => {
+    if (
+      meetingManager.getDeviceLabels() === DeviceLabels.Audio ||
+      meetingManager.getDeviceLabels() === DeviceLabels.AudioAndVideo
```
michhyun1, line 62: Is it possible to use the `selectedAudioInputDevice` instead of the device labels? If the `selectedAudioInputDevice` is undefined, then we don't want to select the audio device that was just plugged in. Whats the difference between this and using device labels exactly?

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -42,15 +43,28 @@ const AudioInputProvider: React.FC<Props> = ({
   const selectedInputRef = useRef(selectedAudioInputDevice);
   selectedInputRef.current = selectedAudioInputDevice;
 
-  const replaceDevice = async (device: string): Promise<AudioInputDevice> => {
-    if (onDeviceReplacement) {
```
ltrung, line 46: Is this to allow customers to handle their own callback? Do we document it in Storybook? I did not find it (Sorry if I missed it).

src/providers/DevicesProvider/AudioInputProvider.tsx
ltrung: DeviceLabel is in our current way to track device permission. By setting it to None it would mean to not trigger any audio or video permission popup.

src/providers/DevicesProvider/AudioInputProvider.tsx
devalevenkatesh: I think it was added by Jing to handle the device selection when a audio input is a transform device:
https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/sdk-providers-meetingprovider--page

We have documented it as a prop on MeetingProvider.

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -42,15 +43,28 @@ const AudioInputProvider: React.FC<Props> = ({
...
+    if (
+      meetingManager.getDeviceLabels() === DeviceLabels.Audio ||
+      meetingManager.getDeviceLabels() === DeviceLabels.AudioAndVideo
+    ) {
+      return true;
     }
-    return device;
-  };
+    return false;
+  }, [meetingManager]);
```
ltrung, line 67: Could this be meetingManager.deviceLabels similar to selectedAudioInput above?

devalevenkatesh: Hmm, no I added the react hook rule and dependency checking eslint plugin to check the warning for that, but since its a function on `MeetingManager` its not needed. Also, Kyu commented to change the getter to `getDeviceLabels` and it has been updated.

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -42,15 +43,28 @@ const AudioInputProvider: React.FC<Props> = ({
...
+  const shouldAutoSelectAudioInputDevice = useCallback((): boolean => {
+    if (
+      meetingManager.getDeviceLabels() === DeviceLabels.Audio ||
+      meetingManager.getDeviceLabels() === DeviceLabels.AudioAndVideo
+    ) {
+      return true;
     }
-    return device;
-  };
+    return false;
```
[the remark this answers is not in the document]
devalevenkatesh: Sure, when we add the flag as part of actual release I will note to not do device selections as part of providers, mostly, it should be only AudioInputProvider, but yes we can club that change with the flag to skip device selection.

src/providers/DevicesProvider/AudioInputProvider.tsx
ltrung: Sorry if I did not understand the code correctly but if someone calls `setupDeviceLabelTrigger` again and change the device label does it update this?

devalevenkatesh: Yes, so if someone calls `setupDeviceLabelTrigger` then the `meetingManager.deviceLabels` will get updated. Now, the next time `audioInputsChanged` observer is triggered, that callback will call `shouldAutoSelectAudioInputDevice` which will get the latest device labels from meetingManager.

The `useCallback` is more of a best practice and optimization to minimize new function creations as part of component re-renders.

ltrung: Oh, i see where I misunderstood the code. Thanks for the explanation.

src/providers/DevicesProvider/AudioInputProvider.tsx
michhyun1: "selectedAudioInputDevice being undefined is a result of builders explicitly setting DeviceLabels.None"

If this is true, then whats the difference between using `selectedAudioInputDevice` vs DeviceLabel.None? Shouldnt it be the same thing?

ltrung (approved):

src/providers/DevicesProvider/AudioInputProvider.tsx
devalevenkatesh: DeviceLabel is a public facing knob to control whether to list and or select devices. For example: As part of meeting leave, we reset the `selectedAudioInputDevice` to `undefined`. Now the `DeviceLabels` will still be `DeviceLabels.AudioAndVideo` and only overridden when done through `join` or `invokeDeviceProvider` but the `selectedAudioInputDevice` for a new join will be `undefined` to start with and it will depend on `DeviceLabels` passed to next `join` to set `selectedAudioInputDevice` to correct device.

Thus, `selectedAudioInputDevice` is actually dependent on `DeviceLabels` been used. In v4, we will anyhow move away from the public fields on `MeetingManager` as those should actually be private (as applicable but selected devices fields all will be private).

michhyun1 (approved):

michhyun1:
Lol no luck with the roster test huh? Can we squash these commits before merging

devalevenkatesh:
> Lol no luck with the roster test huh? Can we squash these commits before merging

Yes, I will squash with meaning commit message as well, but I think I have fixed it in latest commit 😅 🤞

src/providers/DevicesProvider/AudioInputProvider.tsx
```diff
@@ -71,6 +71,16 @@ const AudioInputProvider: React.FC<Props> = ({
       audioInputsChanged: async (newAudioInputs: MediaDeviceInfo[]) => {
         logger.info('AudioInputProvider - audio inputs updated');
 
+        if (
+          meetingManager.getDeviceLabels() !== DeviceLabels.Audio &&
+          meetingManager.getDeviceLabels() !== DeviceLabels.AudioAndVideo
+        ) {
+          logger.info(
+            'Device labels do not allow audio, thus, not selecting any audio input on audioInputsChanged'
```
ltrung, line 79: nit: `Skipping` sounds better than `not selecting` and more inline with JS SDK

ltrung (approved):

integration/script/run-test
```diff
@@ -161,22 +163,22 @@ const startTestDemo = () => {
...
-    await new Promise((resolve) => setTimeout(resolve, 1000));
-  }
-
-  console.log(red, ' Test demo did not start successfully');
-  terminateTestDemo(pid);
-  exit(1);
+  
+    console.log(red, ' Test demo did not start successfully');
+    terminateTestDemo();
+    exit(1);
```
xuesichao, line 181: Seems indent is not correct here.

xuesichao (approved):

integration/script/run-test
devalevenkatesh: prettified the whole file and renamed to `run-test.js` from just `run-test`.

michhyun1 (approved):

samfenwick:
Hi, are you able to provide info on when version 3.6.0 will be released? This issue is causing a bug for us in production currently, for which we now have a temporary workaround

PR #882, opened 2023-03-22
Add `MeetingStatus.Reconnecting` and `audioVideoDidStartConnecting` observer

devalevenkatesh:
**Issue #:** 
NA

**Description of changes:**

- `audioVideoDidStartConnecting` helps in listening to whether JS SDK is reconnecting.

- Applications depending on Chime SDK component library currently do not have a out of the box `MeetingStatus` to tell whether the JS SDK is in reconnection. Thus, add `audioVideoDidStartConnecting` with a new `MeetingStatus.Reconnecting` to aid such a use case.

- Fix: Publish `MeetingStatus.Failed` only when a non-terminal failure is encountered. Currently, some statuses pass both the `sessionStatus.isFailure()` as well as `sessionStatus.isTerminal()` in JS SDK, thus, for a status if it is in both, then it will be considered as failure and `MeetingStatus.TerminalFailure` will never get set. Check below file for more information on both methods in JS SDK: https://github.com/aws/amazon-chime-sdk-js/blob/main/src/meetingsession/MeetingSessionStatus.ts

**Testing**

   1. Start the demo
   2. Join a meeting
   3. Monitor notifications show for meeting loading and start succeeding once you go into the meeting screen.
   4. Turn off Wifi and check for reconnecting status.
   5. Turn on Wifi and check for meeting start succeeding status again.

I will create the PR for testing this in the component library meeting demo app, once the change is merged in this repo. But here is a recording of a new `MeetingStatusIndicator` component that I will add in meeting demo to showcase the meeting status.

https://user-images.githubusercontent.com/61809205/227061511-99469436-504f-40d4-a1a7-baf59027f152.mov

### DIFF
--- a/src/hooks/sdk/docs/useMeetingStatus.stories.mdx
+++ b/src/hooks/sdk/docs/useMeetingStatus.stories.mdx
@@ -12,7 +12,8 @@ enum MeetingStatus {
   Ended,
   JoinedFromAnotherDevice,
   Left,
-  TerminalFailure
+  TerminalFailure,
+  Reconnecting,
 }
 ```
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Applications depending on Chime SDK component library currently do not have a out of the box `MeetingStatus` to tell whether the JS SDK is in reconnection. Thus, add `audioVideoDidStartConnecting` with a new `MeetingStatus.Reconnecting` to aid such a use case.
+
 ### Removed
 
 ### Changed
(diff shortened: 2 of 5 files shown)


src/hooks/sdk/docs/useMeetingStatus.stories.mdx
```diff
@@ -8,6 +8,7 @@ The `useMeetingStatus` hook returns the meeting status from the enum below.
 enum MeetingStatus {
   Loading,
   Succeeded,
+  Reconnecting,
```
simmkyu, line 11: Adding a new enum will break the builder's application with this ESLint option.
https://typescript-eslint.io/rules/switch-exhaustiveness-check

I wonder if we should consider it a breaking change. What do you think?

ltrung (approved):

src/hooks/sdk/docs/useMeetingStatus.stories.mdx
devalevenkatesh: Interesting, thanks for bringing it up. Definitely have to check on breaking change rules on enums in typescript.

devalevenkatesh: I could not find a rule for enums or more information on breaking changes on enums, but like you said Union type rule may apply, in that case it would be hard to extend the enum. I think, to avoid completely breaking, we can improve this in next versions where either we move to `type` or have `string` valued `enums` over `number` incremented one.

With this change for to be safe, I will add the enum at the end of enums values over doing it in middle. That will help us avoid breaking application if depending on index of enum.

simmkyu: For example, if you add the `Reconnected` enum, the `else` case is no longer valid in a builder's application.

```ts
enum MeetingStatus {
  Failed,
  Succeeded,
}
 
function log(status: MeetingStatus) {
  if (status === MeetingStatus.Failed) {
    console.log('Failed!');
  } else {
    console.log('Everything is good');
  }
}
```

Regardless, adding an enum won't break the builder's application unless they have enabled `switch` linting rule.

simmkyu (approved):

src/hooks/sdk/docs/useMeetingStatus.stories.mdx
devalevenkatesh: Right, but its very valid point of concern and two learnings from this:
1. Union types are preferred over enums generally in typescript.
2. If using enums, have string valued enums over indexed ones to have correct logical ordering in enums and have non-breaking changes if someone is using indexed enum.

Issue #888, opened 2023-04-18
Incorrect error being thrown when starting video input device from MeetingManager


Nanosync:
### What happened and what did you expect to happen?

I am using `meetingManager.startVideoInputDevice()` and `meetingManager.stopVideoInputDevice` to start/stop the video input device, and if an error is thrown, the error name is used to display an appropriate error message.

**Current Result:**

These 2 functions are wrapping the original Error coming from `handleGetUserMedia` (in device controller from the JS sdk) and throwing a generic Error.

**Expected Result:**

The original Error should be thrown for these 2 functions, and possibly all the other functions (`startAudioInputDevice`, `stopAudioInputDevice`, `startAudioOutputDevice`, `stopAudioOutputDevice`).

I also note that in `setupDeviceLabelTrigger`, the error is being wrapped (`throw new Error(error)`), it should `throw error` directly.

### Have you reviewed our existing documentation?

- [X] Amazon Chime SDK for JavaScript GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] Amazon Chime SDK React Components Library GitHub [issues](https://github.com/aws/amazon-chime-sdk-component-library-react/issues)
- [X] Storybook [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/introduction--page)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-component-library-react#readme)
- [X] How-to [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/how-tos--page)
- [X] React SDK meeting [demo](https://github.com/aws-samples/amazon-chime-sdk/tree/main/apps/meeting#readme---react-meeting-demo)

### Reproduction steps

1. Block browser permissions or if on Windows, ensure that browser permissions are granted and the video device is being used in another application.
2. Call `await meetingManager.startVideoInputDevice()`.
3. When `PermissionDeniedError` / `NotReadableError` is thrown, the Error name is "Error" with reason "MeetingManager failed to select video input device"

### Amazon Chime SDK React Components Library version

3.5.0

### What browsers are you seeing the problem on?

Chrome

### Browser version

Version 112.0.5615.49 (Official Build) (arm64)

### Device Information

macOS Monterey 12.6.1

### Browser console logs

```
MeetingManager failed to select video input device PermissionDeniedError: Permission denied by browser
[...]

PR #891, opened 2023-05-04
MeetingManager throws original error name from getUserMedia call

xuesichao:
**Issue #888 :** 
We wrap the original error from `getUserMedia` with the general `Error`. It make it difficult for builders to handle the actual error. With this change, `startAudioInputDevice`, `startVideoInputDevice`, and `setupDeviceLabelTrigger` of `MeetingManager` now throw errors with the original error `name` returned by the `getUserMedia` call.

**Description of changes:**
`startAudioInputDevice`, `startVideoInputDevice`, and `setupDeviceLabelTrigger` of `MeetingManager` now throw errors with the original error `name` returned by the `getUserMedia` call.

For example, before this change the `error` thrown by `startVideoInputDevice` was:
```
{ 
  name: Error, 
  message: MeetingManager failed to select video input device
}
 ```
No it is:
```
{ 
  name: PermissionDeniedError, 
  message: MeetingManager failed to select audio input device. Permission denied by browser
}
```

**Testing**
Yes

    1. Modified the roster-test-app to add buttons to call `startAudioInputDevice`, `startVideoInputDevice` manually.
    2. Block the browser device permission.
    3. Using roster test app to join a meeting.
    4. Verify the error log for `setupDeviceLabelTrigger`.
    5. Click button to trigger `startAudioInputDevice`, `startVideoInputDevice` and verify the error log.

7. If you made changes to the component library, have you provided corresponding documentation changes?
No doc needs to be changed.

### CODE REVIEW

File: src/providers/MeetingProvider/MeetingManager.ts [hunk 1]
```diff
@@ -109,8 +109,8 @@ export class MeetingManager implements AudioVideoObserver {
   private eventDidReceiveRef: EventObserver;
 
   private deviceLabels: DeviceLabels | DeviceLabelTrigger;
-  
-  public getDeviceLabels() : DeviceLabels | DeviceLabelTrigger {
```
xuesichao, line 113: Removed accessibility modifier `public` because we have below eslint rule:
https://github.com/aws/amazon-chime-sdk-component-library-react/blob/b3d5389686419758cd2a327399f93098539690e0/.eslintrc.json#L62-L70

michhyun1 (dismissed):

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -478,11 +478,10 @@ export class MeetingManager implements AudioVideoObserver {
       this.selectedAudioInputDevice = device;
       this.publishSelectedAudioInputDevice();
     } catch (error) {
-      console.error(
-        'MeetingManager failed to select audio input device',
-        error
-      );
-      throw new Error('MeetingManager failed to select audio input device');
+      error.message =
```
devalevenkatesh, line 481: Do we need to type check `error` before assigning `error.message`? If all error that are caught here are instance of `Error` then `error.message` is fine otherwise can throw runtime error.

xuesichao: From JS SDK this `error` is 100% an instance of `Error`. But React SDK does not have this info so it's good to do a type check here. Good call! 

Strange that typescript does not give me an error for this.

src/providers/MeetingProvider/MeetingManager.ts
```diff
@@ -368,7 +368,7 @@ export class MeetingManager implements AudioVideoObserver {
           console.error('MeetingManager failed to get device permissions');
           this.deviceLabelTriggerStatus = DeviceLabelTriggerStatus.DENIED;
           this.publishDeviceLabelTriggerStatus();
-          throw new Error(error);
```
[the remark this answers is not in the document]
xuesichao: We discussed this earlier and agreed that it's fine to make this change. The `error` here is the original `getUserMedia` errors, for example `NotReadableError`. These errors are of type `DOMException` which is inherited from `Error` prototype, so it's not a breaking change. 

We wrap these errors with a general `Error` type which is not helpful for builder because they are not able to check the name of the error (`error.name` is always "Error"). And this is why we return it directly here.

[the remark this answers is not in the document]
devalevenkatesh: Cool thanks for the context.

devalevenkatesh (approved):

Issue #923, opened 2024-01-05
[FEATURE REQUEST] Add Node.js LTS (20.x) support


mystafer:
## Community Note

* Please do not add additional comments such as "+1" or "me too" to existing issues - instead please vote on the issue by adding a 👍 to the issue. The Amazon Chime SDK team will prioritize the request based on the number of 👍 the issue has received.
* Please go through the existing [issues](https://github.com/aws/amazon-chime-sdk-component-library-react/issues) to see if your request has already been recommended.
* Please do not use this template to report bugs. This repository has an [issue template](https://github.com/aws/amazon-chime-sdk-component-library-react/issues/new/choose) for issues and [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/docs/quick-starts--page) to help with your development.
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.

## Tell us about your request

The package is missing support for the current Node.js LTS version, which recently changed to 20.10.0.

What do you want us to build?

Add support for Node engine ^20.

How are you currently solving a problem?

Using previous LTS engine.

vitiaaap29:
Guys, we really need NodeJS 20 support. Please add!

Issue #905, opened 2023-07-21
Support npm 9


mcqj:
### What happened and what did you expect to happen?

package.json specifically required npm 5, 6, 7, 8 but not 9.
So, errors occur if you are using npm 9

### Have you reviewed our existing documentation?

- [X] Amazon Chime SDK for JavaScript GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] Amazon Chime SDK React Components Library GitHub [issues](https://github.com/aws/amazon-chime-sdk-component-library-react/issues)
- [X] Storybook [documentation](https://aws.github.io/amazon-chime-sdk-component-library-react/?path=/story/introduction--page)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-component-library-react#readme)
- [X] React SDK meeting [demo](https://github.com/aws-samples/amazon-chime-sdk/tree/main/apps/meeting#readme---react-meeting-demo)

### Reproduction steps

npm install with npm 9

### Amazon Chime SDK React Components Library version

3.7.0

### What browsers are you seeing the problem on?

NA

### Browser version

NA

### Device Information

NA

### Browser console logs

Not applicable

OrrTytocare:
Any progress on this issue ?

suula:
`node 20` and `npm 10` should be added

Tjs002:
@suula is this issue solved if not can I work on this?

suula:
> @suula is this issue solved if not can I work on this?

By all means

ziyiz-amzn:
The https://github.com/aws/amazon-chime-sdk-component-library-react/pull/935 solved this.

PR #935, opened 2024-03-06
Add support for node 20 and drop support for node < 18

xuesichao:
**Issue #:** 
#923 
#905 

**Description of changes:**
Add support for node 20 and drop support for node < 18

**Testing**
Yes

Install in meeting demo and verify e2e works.

N/A


tst/providers/MeetingProvider/index.test.tsx
```diff
@@ -18,17 +19,21 @@ import { MeetingManagerJoinOptions } from '../../../src/providers/MeetingProvide
...
+} from '../../../src/providers/MeetingProvider';
 
 describe('Meeting Provider', () => {
   it('events are received correctly', async () => {
     // Mock the user agent to ensure the MeetingManager event subscription is
     // set up. Otherwise, the amazon-chime-sdk-js will not detect a valid
     // browser to use.
     const userAgentGet = jest.spyOn(navigator, 'userAgent', 'get');
-    userAgentGet.mockReturnValue('Chrome/96.0');
+    userAgentGet.mockReturnValue(
```
xuesichao, line 34: Only change mock user agent to fix a unit test error. Rest change in this file are linting.

tst/providers/MeetingProvider/index.test.tsx
```diff
@@ -18,17 +19,21 @@ import { MeetingManagerJoinOptions } from '../../../src/providers/MeetingProvide
...
+      'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/122.0.0.0 Safari/537.36'
```
georgezy-amzn, line 35: Any reason why this particular value is being used? maybe add a comment here so that other engineers know.

.github/workflows/continuous-integration.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: [20.x]
```
georgezy-amzn, line 19: Are there any doc(like the requirements for deployment the lib) we need to update for node version upgrade?

georgezy-amzn (commented):
Looks like there are some format only changes, are we using a consistent format across the team?

ltrung (approved):

tst/providers/MeetingProvider/index.test.tsx
xuesichao: This is just for testing. Any user agent works here. The reason for this change is that the previous mockValue `'Chrome/96.0'` is partial, unit test reports error when parsing this user agent. So I just replaced it with a complete example.

xuesichao:
> Looks like there are some format only changes, are we using a consistent format across the team?

Yes, we are using Eslint to formate the code. The format changes in this file, I think it's because we haven't touched this file for a long time. Our linting only applies to new files/changes I believe. There're lots of errors and we haven't fixed those so can use lint to fix all formatting issue at one time.

georgezy-amzn (approved):

PR #958, opened 2024-07-31
Add changeBackgroundReplacementImage to BackgroundReplacementProvider

dingyishen-amazon:
**Issue #:** 

**Description of changes:**
Add `changeBackgroundReplacementImage` function in the `BackgroundReplacementProvider` to enable the functionality of changing the background replacement image.

**Testing**
Yes
Make corresponding changes in the demo app, and test with it. Demo Video attached: 
https://github.com/user-attachments/assets/20c92ebf-2d25-4375-870b-b4e00bf2cdbd

No, will add the documentation change after the demo changes are merged since the documentation example should be similar with demo.


src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -179,10 +182,26 @@ export const BackgroundReplacementProvider: FC<
     }
   };
 
+  const changeBackgroundReplacementImage = async (
+    imageBlob: Blob
+  ) => {
+    if (!backgroundReplacementProcessor) {
+      logger.info(`BackgroundReplacementProcessor has not been initialized yet.`);
```
georgezy-amzn, line 189: nit: maybe warning or error instead?

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -179,10 +182,26 @@ export const BackgroundReplacementProvider: FC<
...
+      return;
+    }
+    try {
+      await backgroundReplacementProcessor.setImageBlob(imageBlob);
+      logger.info(`Background replacement image changed to new imageBlob: ${imageBlob}`);
+    } catch (error) {
+      logger.error(`Failed to change the background replacement image: ${error}`);
```
georgezy-amzn, line 196: Under what condition this can fail? should the consumer be notified?

michhyun1:
Might need to squash the two commits

src/providers/BackgroundReplacementProvider/index.tsx
dingyishen-amazon: If in the js library, it fails to load image with the blob https://github.com/aws/amazon-chime-sdk-js/blob/bec2e70f2488658645494d01c141a6fb2939fa47/src/backgroundreplacementprocessor/BackgroundReplacementFilter.ts#L116, then it will fail. 
By notifying the consumer, do you mean throw an error?

georgezy-amzn: Yes. Also do we have a follow up task to update the demo and docs?

georgezy-amzn: Are there any restrictions on the image? If yes, let's also make sure the doc covers that.

dingyishen-amazon: Yes, we have a follow up, there is no restriction if you pass a Blob, and we will put the example in demo and the doc.

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -179,10 +182,28 @@ export const BackgroundReplacementProvider: FC<
...
+  ) => {
+    if (!backgroundReplacementProcessor) {
+      logger.warn(`BackgroundReplacementProcessor has not been initialized yet.`);
+      return;
+    }
+    try {
+      await backgroundReplacementProcessor.setImageBlob(imageBlob);
+      logger.info(`Background replacement image changed to new imageBlob: ${imageBlob}`);
+    } catch (error) {
+      throw new Error(
```
georgezy-amzn, line 196: I’m just asking for you opinion if we should notify consumer, nor enforcing you. 😂 You may want to use Promise if throwing error.

dingyishen-amazon: What is the criteria for throwing error/logging error? As for this use case, if there is error, then we keep use the previous image. Then I think we do not need to throw error.

georgezy-amzn: 2 factors I can think of:
- check how other api handles error in this repo
- Is there any related customer ask for the failure notification, customer may want to notify the end-users if failed to apply the image.

michhyun1 (dismissed):

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
georgezy-amzn, line 10: Don't think React component use `Unreleased`, the change should add to `3.9.0`.

src/providers/BackgroundReplacementProvider/index.tsx
georgezy-amzn: Don't think you are using Promise correctly...

dingyishen-amazon: I have tested the code with my changes in demo app, what do you mean by not correct?

dingyishen-amazon (commented):
Add logger.error on line 196 as @xuesichao requested

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -40,6 +40,9 @@ interface BackgroundReplacementProviderState {
   ) => Promise<DefaultVideoTransformDevice>;
   isBackgroundReplacementSupported: boolean | undefined;
   backgroundReplacementProcessor: BackgroundReplacementProcessor | undefined;
+  changeBackgroundReplacementImage: (
+    imageBlob: Blob
+  ) => {};
```
georgezy-amzn, line 45: nit: return Promise<void>

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -179,10 +182,29 @@ export const BackgroundReplacementProvider: FC<
     }
   };
 
+  const changeBackgroundReplacementImage = async (
+    imageBlob: Blob
+  ): Promise<void> => {
+    if (!backgroundReplacementProcessor) {
+      logger.warn(`BackgroundReplacementProcessor has not been initialized yet.`);
+      return;
```
georgezy-amzn, line 190: Did you verify this works?

georgezy-amzn: Nvm, I think it works.

georgezy-amzn (approved):

xuesichao (approved):

src/providers/BackgroundReplacementProvider/index.tsx
dingyishen-amazon: Will update.

ltrung (commented):
Do we need to update the story book documentation too?

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -40,6 +40,9 @@ interface BackgroundReplacementProviderState {
   ) => Promise<DefaultVideoTransformDevice>;
   isBackgroundReplacementSupported: boolean | undefined;
   backgroundReplacementProcessor: BackgroundReplacementProcessor | undefined;
+  changeBackgroundReplacementImage: (
```
ltrung, line 43: Can we add unit test for this file?

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -179,10 +182,29 @@ export const BackgroundReplacementProvider: FC<
     }
   };
 
+  const changeBackgroundReplacementImage = async (
```
ltrung, line 185: So this only allows to change background image and they could not change back to background color right?

dingyishen-amazon: It allows to change back to color, I set the name to `changeBackgroundReplacementImage` as the method in the js library is `setImageBlob`, color is also a kind of image.

ltrung: Kinda confusing but we could clarify that in the storybook

dingyishen-amazon: Storybook updated.

PR #985, opened 2025-01-27
Fix conditions where BackgroundReplacementProvider needs to be re-rendered, Fix to not log the entire options object

ziyiz-amzn:
**Issue #:** 

**Description of changes:**
1. Fix the conditions where `BackgroundReplacementProvider` needs to be re-rendered. Remove the image size comparison because now it supports dynamically change replacement image in an active processor. The processor recreation when image changes will cost to fetch assets again.
2. Avoid to log the entire `options` props in `BackgroundReplacementProvider`, the optional `logger` prop could potentially be nested JSON.

**Testing**
Yes

3. How did you test these changes?
4. Locally and private build tested by QA

5. If you made changes to the component library, have you provided corresponding documentation changes?

### DIFF
--- a/src/providers/BackgroundBlurProvider/index.tsx
+++ b/src/providers/BackgroundBlurProvider/index.tsx
@@ -110,7 +110,11 @@ export const BackgroundBlurProvider: FC<React.PropsWithChildren<Props>> = ({
     logger.info(
       `Initializing background blur processor with, spec: ${JSON.stringify(
         spec
-      )}, options: ${JSON.stringify(options)}`
+      )}, options: { blurStrength: ${
+        options?.blurStrength
+      }, filterCPUUtilization: ${
+        options?.filterCPUUtilization
+      }, reportingPeriodMillis: ${options?.reportingPeriodMillis} }`
     );
 
     try {
--- a/src/providers/BackgroundReplacementProvider/index.tsx
+++ b/src/providers/BackgroundReplacementProvider/index.tsx
@@ -82,8 +82,7 @@ export const BackgroundReplacementProvider: FC<
         return true;
       }
       if (
-        prev?.imageBlob?.size === next?.imageBlob?.size ||
-        prev?.filterCPUUtilization === next?.filterCPUUtilization ||
+        prev?.filterCPUUtilization !== next?.filterCPUUtilization ||
         prev?.logger !== next?.logger ||
         prev?.reportingPeriodMillis !== next?.reportingPeriodMillis
       ) {
@@ -111,7 +110,11 @@ export const BackgroundReplacementProvider: FC<
     logger.info(
       `Initializing background replacement processor with, ${JSON.stringify(
         spec
-      )}, ${JSON.stringify(options)}`
+      )}, options: { filterCPUUtilization: ${
+        options?.filterCPUUtilization
+      }, reportingPeriodMillis: ${
+        options?.reportingPeriodMillis
+      }}`
     );
 
     try {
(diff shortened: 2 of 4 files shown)


src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -110,7 +110,11 @@ export const BackgroundBlurProvider: FC<React.PropsWithChildren<Props>> = ({
     logger.info(
       `Initializing background blur processor with, spec: ${JSON.stringify(
         spec
```
michhyun1, line 112: Do you want to break this object out also?

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -111,7 +110,11 @@ export const BackgroundReplacementProvider: FC<
     logger.info(
       `Initializing background replacement processor with, ${JSON.stringify(
         spec
```
michhyun1, line 112: same here?

michhyun1 (dismissed):

src/providers/BackgroundBlurProvider/index.tsx
ziyiz-amzn: I think `spec` is fine, it only contains string and number.

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -82,8 +82,7 @@ export const BackgroundReplacementProvider: FC<
         return true;
       }
       if (
-        prev?.imageBlob?.size === next?.imageBlob?.size ||
```
xuesichao, line 85: Do we need to keep the comparison for `imageBlob?.size`?

ziyiz-amzn: I prefer to remove this comparison. 
1. image in the constructor should only take effect when initiate the component, changing image should always through API `processor.setImageBlob(imageBlob)`. 
2. Also, I verified that assets will be re-fetched if BackgroundReplacementProvider re-renders, we may want to avoid it. 

I will merge this PR only after QA's signoff.

ziyiz-amzn: > Is it a breaking change that removing the comparison for imageBlob seems to affect the second flow?

Before this change, the provider will not re-create even if pass in a different image in constructor. correct me if wrong.

xuesichao: >1. image in the constructor should only take effect when initiate the component, changing image should always through API processor.setImageBlob(imageBlob).
>2. Also, I verified that assets will be re-fetched if BackgroundReplacementProvider re-renders, we may want to avoid it.

I agree this is best practice. However before previous release, `processor.setImageBlob` was not available in React SDK. And builders can only rely on the options props change to re render entire BackgroundReplacementProvider to update the processor. Though there's a bug to make it never works, so this 
 change may not be a breaking one. 

My thought is the behavior of BGB and BGR provider should align meaning any change of their configuration in the provider props should cause the provider to re-render.
- https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/15_Background_Filter_Video_Processor.md#configuration

michhyun1: What use case is there for the BGBlur or BGR Provider to re-render? When would they need to kill the old processor and create a new one? 

I think the preferred behavior is that the BGBlur and BGR provider just take in an initial props, but don't re-render on those prop changes. 

Instead, we should just expose functions to dynamically change the processor with a new image blob, or new blur strength.

ziyiz-amzn: Currently, if cx changes replacement image in props, process will not be re-created. It will be a breaking change if it allows re-creation upon props change.

Thus, I think removing image.blob check keeps it non-breaking, meanwhile, it pushes cx to use processor.setImageBlob()

xuesichao (approved):

PR #998, opened 2025-04-22
Add support for multiple content shares

xuesichao:
**Issue #:** 

**Description of changes:**

- Added support for multiple content shares in meetings through the `ContentShareProvider`. The provider now accepts a `maxContentShares` prop (default: 1, range 1-2) to specify the maximum number of concurrent content shares allowed.
  - Added new collections in `ContentShareState` to track multiple content shares: `tiles`, `tileIdToAttendeeId`, and `attendeeIdToTileId`.
  - Added optional `tileId` prop to the `ContentShare` component to specify which content share to render.
  - Added `canStartContentShare` state to control when content sharing is allowed based on the current number of shares and configured maximum.
  - Maintained backward compatibility by keeping `tileId` and `sharingAttendeeId` properties, which now point to the most recently started content share when multiple shares are present.

**Testing**

Yes

Update meeting demo to enable multiple content shares and verify the hook values, ContentTile components are working as expected.

Yes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added support for multiple content shares in meetings through the `ContentShareProvider`. The provider now accepts a `maxContentShares` prop (default: 1, range 1-2) to specify the maximum number of concurrent content shares allowed.
+  - Added new collections in `ContentShareState` to track multiple content shares: `tiles`, `tileIdToAttendeeId`, and `attendeeIdToTileId`.
+  - Added optional `tileId` prop to the `ContentShare` component to specify which content share to render.
+  - Added `canStartContentShare` state to control when content sharing is allowed based on the current number of shares and configured maximum.
+  - Maintained backward compatibility by keeping `tileId` and `sharingAttendeeId` properties, which now point to the most recently started content share when multiple shares are present.
+
 ### Removed
 
 ### Changed
--- a/src/components/sdk/ContentShare/ContentShare.mdx
+++ b/src/components/sdk/ContentShare/ContentShare.mdx
@@ -5,11 +5,15 @@ import { ContentShare } from './';
 
 # ContentShare
 
-The `ContentShare` component renders a `ContentTile` for the active content share video, remote or local.
+The `ContentShare` component renders a `ContentTile` for a content share video, remote or local.
 
 If used within the `VideoGrid` component, it will automatically place the active tile in the featured grid slot. It takes precedence over the featured video tile.
 
 Once a meeting session has been started, a user can start and stop content sharing by using the `useContentShareControls` hook.
+
+## Multiple Content Shares
+
+With the support for multiple content shares, you can now specify which content share tile to render by providing the `tileId` prop. If no `tileId` is provided, the component will render the default content share tile from (`const { tileId } = useContentShareState()`).
 
 ## Importing
 
@@ -19,12 +23,14 @@ import { ContentShare } from 'amazon-chime-sdk-component-library-react';
 
 ## Usage
 
+### With Single Content Share
+
 ```jsx
 import React from 'react';
 import {
   MeetingProvider,
   ContentShare,
-  useContentShareControls
+  useContentShareControls,
 } from 'amazon-chime-sdk-component-library-react';
 
 const App = () => {
@@ -32,9 +38,66 @@ const App = () => {
 
   return (
     <MeetingProvider>
-      <ContentShare nameplate='Content share' />
+      <ContentShare nameplate="Content share" />
       <button onClick={toggleContentShare}>Toggle content share</button>
     </MeetingProvider>
+  );
+};
+```
+
+### With Multiple Content Shares
+
+```jsx
+import React from 'react';
+import {
+  MeetingProvider,
+  ContentShare,
+  ContentShareProvider,
+  useContentShareState,
+  useContentShareControls,
+} from 'amazon-chime-sdk-component-library-react';
+
+const App = () => {
+  return (
+    <MeetingProvider>
+      <ContentShareProvider maxContentShares={2}>
+        <ContentShareView />
+        <ContentShareControls />
+      </ContentShareProvider>
+    </MeetingProvider>
+  );
+};
+
+const ContentShareView = () => {
+  const { tiles, tileIdToAttendeeId } = useContentShareState();
+
+  return (
+    <div>
+      {tiles.length === 0 ? (
+        <p>No content is being shared</p>
+      ) : (
+        <div className="content-share-container">
+          {tiles.map((tileId) => (
+            <ContentShare
+              key={tileId}
+              tileId={tileId}
+              nameplate={`Shared by: ${tileIdToAttendeeId[tileId.toString()]}`}
+            />
+          ))}
+        </div>
+      )}
+    </div>
+  );
+};
+
+const ContentShareControls = () => {
+  const { toggleContentShare } = useContentShareControls();
+  const { canStartContentShare } = useContentShareState();
+
+  return (
+    <button onClick={toggleContentShare} disabled={!canStartContentShare}>
+      Share content
+    </button>
   );
 };
 ```
--- a/src/components/sdk/ContentShare/index.tsx
+++ b/src/components/sdk/ContentShare/index.tsx
@@ -10,37 +10,49 @@ import { BaseSdkProps } from '../Base';
 
 interface Props extends BaseSdkProps {
   nameplate?: string;
+  tileId?: number;
 }
 
 export const ContentShare: React.FC<React.PropsWithChildren<Props>> = ({
   className,
+  tileId,
   ...rest
 }) => {
   const audioVideo = useAudioVideo();
-  const { tileId } = useContentShareState();
+  const contentShareState = useContentShareState();
+
+  // Use the provided tileId or fall back to the default (for backward compatibility)
+  const tileIdToRender =
+    tileId !== undefined ? tileId : contentShareState.tileId;
+
+  const attendeeId = tileIdToRender
+    ? contentShareState.tileIdToAttendeeId[tileIdToRender.toString()]
+    : null;
+
   const videoEl = useRef<HTMLVideoElement | null>(null);
 
   useEffect(() => {
-    if (!audioVideo || !videoEl.current || !tileId) {
+    if (!audioVideo || !videoEl.current || !tileIdToRender) {
       return;
     }
 
-    audioVideo.bindVideoElement(tileId, videoEl.current);
+    audioVideo.bindVideoElement(tileIdToRender, videoEl.current);
 
     return () => {
-      const tile = audioVideo.getVideoTile(tileId);
+      const tile = audioVideo.getVideoTile(tileIdToRender);
       if (tile) {
-        audioVideo.unbindVideoElement(tileId);
+        audioVideo.unbindVideoElement(tileIdToRender);
       }
     };
-  }, [audioVideo, tileId]);
+  }, [audioVideo, tileIdToRender]);
 
-  return tileId ? (
+  return tileIdToRender ? (
     <ContentTile
       objectFit="contain"
-      className={className || ''}
+      className={`ch-content-share--${tileIdToRender} ${className || ''}`}
       {...rest}
       ref={videoEl}
+      data-content-share-attendee={attendeeId}
     />
   ) : null;
 };
--- a/src/providers/ContentShareProvider/docs/ContentShareProvider.mdx
+++ b/src/providers/ContentShareProvider/docs/ContentShareProvider.mdx
@@ -6,12 +6,26 @@ import { Meta } from '@storybook/blocks';
 
 The `ContentShareProvider` provides state and functionality for content sharing.
 
+### Props
+
+```javascript
+{
+  // Maximum number of concurrent content shares allowed (default: 1, range: 1-2)
+  maxContentShares?: number;
+}
+```
+
 ### State
 
 ```javascript
 {
-  // The tile ID of the active content share
+  // The tile ID of the active content share (deprecated, maintained for backward compatibility)
+  // When multiple content shares are present, this points to the most recently started content share
   tileId: number | null;
+
+  // The chime attendee ID of the user sharing (deprecated, maintained for backward compatibility)
+  // When multiple content shares are present, this points to the attendee of the most recently started content share
+  sharingAttendeeId: string | null;
 
   // Whether the content share is paused
   paused: boolean;
@@ -22,8 +36,17 @@ The `ContentShareProvider` provides state and functionality for content sharing.
   // Whether or not the local user's content share is loading
   isLocalShareLoading: boolean;
 
-  // The chime attendee ID of the user sharing
-  sharingAttendeeId: string | null;
+  // Whether the user can start a content share based on current limits and `isLocalUserSharing`
+  canStartContentShare: boolean;
+
+  // Array of all content share tile IDs
+  tiles: number[];
+
+  // Map of tile IDs to attendee IDs
+  tileIdToAttendeeId: { [key: string]: string };
+
+  // Map of attendee IDs to tile IDs
+  attendeeIdToTileId: { [key: string]: number };
 }
 ```
 
@@ -50,6 +73,49 @@ const MyChild = () => {
   const { isLocalUserSharing } = useContentShareState();
 
   return isLocalUserSharing ? <div>Sharing</div> : <div>Not sharing</div>;
+};
+```
+
+## Multiple Content Shares
+
+You can configure the maximum number of concurrent content shares allowed (up to 2):
+
+```jsx
+import React from 'react';
+import {
+  MeetingProvider,
+  ContentShareProvider,
+  useContentShareState,
+} from 'amazon-chime-sdk-component-library-react';
+
+const App = () => (
+  <MeetingProvider>
+    {/* Override the default maxContentShares value (1) in ContentShareProvider */}
+    <ContentShareProvider maxContentShares={2}>
+      <MyComponent />
+    </ContentShareProvider>
+  </MeetingProvider>
+);
+
+const MyComponent = () => {
+  const { tiles, tileIdToAttendeeId, canStartContentShare } =
+    useContentShareState();
+
+  return (
+    <div>
+      <p>Content shares available: {tiles.length}</p>
+      <p>Can start content share: {canStartContentShare ? 'Yes' : 'No'}</p>
+      <div>
+        {tiles.map((tileId) => (
+          <ContentShare
+            key={tileId}
+            tileId={tileId}
+            nameplate={`Shared by: ${tileIdToAttendeeId[tileId.toString()]}`}
+          />
+        ))}
+      </div>
+    </div>
+  );
 };
 ```
 
--- a/src/providers/ContentShareProvider/docs/useContentShareControls.mdx
+++ b/src/providers/ContentShareProvider/docs/useContentShareControls.mdx
@@ -13,6 +13,18 @@ The `useContentShareControls` hook returns the state and functionality around st
   // Whether or not the local content share is paused
   paused: boolean;
 
+  // Whether or not the local user is currently sharing content
+  isLocalUserSharing: boolean;
+
+  // Whether or not the local user's content share is loading
+  isLocalShareLoading: boolean;
+
+  // Determines if the user can start a new content share
+  // Returns false in two cases:
+  // 1. When the ContentShareProvider has reached its `maxContentShares` limit
+  // 2. When the local user is already sharing content
+  canStartContentShare: boolean;
+
   // A function to toggle content share.
   // If the user is sharing, it will stop the content share.
   // If a user is not sharing, it will start the content share workflow.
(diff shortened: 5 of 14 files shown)


src/components/sdk/ContentShare/ContentShare.mdx
```diff
@@ -19,26 +23,85 @@ import { ContentShare } from 'amazon-chime-sdk-component-library-react';
...
+  useContentShareControls,
 } from 'amazon-chime-sdk-component-library-react';
 
 const App = () => {
   const { toggleContentShare } = useContentShareControls();
 
   return (
     <MeetingProvider>
-      <ContentShare nameplate='Content share' />
+      <ContentShare nameplate="Content share" />
```
ltrung, line 41: Are we using " or ' in this repo?

src/components/sdk/ContentShare/index.tsx
```diff
@@ -10,37 +10,49 @@ import { BaseSdkProps } from '../Base';
 
 interface Props extends BaseSdkProps {
   nameplate?: string;
+  tileId?: number;
```
xuesichao, line 13: Note that since the `VideoGrid` UI Component only supports one featured area/tile, when there's two content shares available, I split the featured area into two parts to display both content share tiles there. This change only exists in meeting demo  (I implemented a custom VideoTileGrid component in demo app) for testing purpose. 

For builders, they will need to create VideoTileGrid to handle multiple content share tiles based on their UX using updated `ContentTile` component.

src/providers/ContentShareProvider/docs/ContentShareProvider.mdx
```diff
@@ -22,8 +36,17 @@ The `ContentShareProvider` provides state and functionality for content sharing.
   // Whether or not the local user's content share is loading
   isLocalShareLoading: boolean;
 
-  // The chime attendee ID of the user sharing
-  sharingAttendeeId: string | null;
+  // Whether the user can start a content share based on current limits and `isLocalUserSharing`
+  canStartContentShare: boolean;
```
xuesichao, line 40: We might want to change this to `hasReachedContentShareLimit` so it's decoupled from `isLocalUserSharing`, let me know your thoughts

src/components/sdk/ContentShare/ContentShare.mdx
xuesichao: We use single quote for JS, but here it should be double quote because it's a property on JSX element. Like we should use double quote on HTML element

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added support for multiple content shares in meetings through the `ContentShareProvider`. The provider now accepts a `maxContentShares` prop (default: 1, range 1-2) to specify the maximum number of concurrent content shares allowed.
```
ziyiz-amzn, line 14: nonblocker: `maxConcurrentContentShare` may be a better name?

src/components/sdk/ContentShare/index.tsx
```diff
@@ -10,37 +10,49 @@ import { BaseSdkProps } from '../Base';
...
     };
-  }, [audioVideo, tileId]);
+  }, [audioVideo, tileIdToRender]);
 
-  return tileId ? (
+  return tileIdToRender ? (
     <ContentTile
       objectFit="contain"
-      className={className || ''}
+      className={`ch-content-share--${tileIdToRender} ${className || ''}`}
```
ziyiz-amzn, line 52: is this a defined class name?

src/providers/ContentShareProvider/docs/ContentShareProvider.mdx
ziyiz-amzn: decouple +1

src/providers/ContentShareProvider/state.tsx
```diff
@@ -90,38 +198,102 @@ export function reducer(
...
         return state;
       }
 
+      const collections = addTileToCollections(
+        state,
+        tileIdToUpdate,
+        attendeeIdToUpdate
+      );
+
+      const backwardCompatibility = updateBackwardCompatibility(
```
ziyiz-amzn, line 218: may have a better var and method name. maybe `updatedState` and `updateForBackwardCompatibility`?

CHANGELOG.md
xuesichao: I feel 'concurrent' is redundant here, since 'maxContentShares' itself is already clear and unambiguous.

src/components/sdk/ContentShare/index.tsx
xuesichao: No, this className is not used in fact, I added here to follow how we did for `RemoteVideo` component, it serves as an identifier
https://github.com/aws/amazon-chime-sdk-component-library-react/blob/efb7f0f4387a0929a1e2c9813a0afc57df1e2286/src/components/sdk/RemoteVideo/index.tsx#L50

CHANGELOG.md
devalevenkatesh: What if we dont have such a prop? Can we take Video tiles like route. For example, we have a max limit on video streams but we dont have a prop for the max limit for them limiting builders. Correct me if not here. Have to check how that is limited component usage wise.

src/components/sdk/ContentShare/ContentShare.mdx
```diff
@@ -5,12 +5,16 @@ import { ContentShare } from './';
...
-The `ContentShare` component renders a `ContentTile` for the active content share video, remote or local.
+The `ContentShare` component renders a `ContentTile` for a content share video, remote or local.
 
 If used within the `VideoGrid` component, it will automatically place the active tile in the featured grid slot. It takes precedence over the featured video tile.
 
 Once a meeting session has been started, a user can start and stop content sharing by using the `useContentShareControls` hook.
 
+## Multiple Content Shares
+
+With the support for multiple content shares, you can now specify which content share tile to render by providing the `tileId` prop. If no `tileId` is provided, the component will render the default content share tile from (`const { tileId } = useContentShareState()`).
```
devalevenkatesh, line 16: So what will the `tileId` default to in case one does not provide it as a prop? Suggest adding a clarification for builder to help get answer to this question in the docs.

src/providers/ContentShareProvider/state.tsx
xuesichao: This is to handle backward compatibility for existing `tileId` and `sharingAttendeeId`.

```ts
const backwardCompatibility = 
  {
    tileId: number | null;
    sharingAttendeeId: string | null;
  }
```

src/components/sdk/ContentShare/ContentShare.mdx
devalevenkatesh: We should also link to `ContentShareProvider` with a small explanation on how `ContentShareProvider maxContentShares={2}` impacts the `ContentShare` component.

xuesichao: >If no `tileId` is provided, the component will render the default content share tile from (`const { tileId } = useContentShareState()`).

I think this is clear enough. It will use the existing `tileId` property from `useContentShareState()` hook

xuesichao: >We should also link to ContentShareProvider with a small explanation on how ContentShareProvider maxContentShares={2} impacts the ContentShare component.

To clarify, `maxContentShares` doesn't impact `ContentShare`.

CHANGELOG.md
devalevenkatesh: Right, just trying to see whether we adding a new prop vs using the `tiles.length` is fine too. If the later is fine, why introduce a new one? Now still going through review, will understand and sync on the breaking change part.

src/components/sdk/ContentShare/index.tsx
```diff
@@ -10,37 +10,49 @@ import { BaseSdkProps } from '../Base';
...
 }
 
 export const ContentShare: React.FC<React.PropsWithChildren<Props>> = ({
   className,
+  tileId,
   ...rest
 }) => {
   const audioVideo = useAudioVideo();
-  const { tileId } = useContentShareState();
+  const contentShareState = useContentShareState();
```
devalevenkatesh, line 22: Thinking from keeping the props limited with no new optional props idea. How about we maintain `tiles` as an array in the content share state. Based on the number of tiles, we should show those many `ContentTile`s?

devalevenkatesh: But yes, we would have to maintain support of `contantShareState.tileId`.

src/components/sdk/ContentShare/index.tsx
devalevenkatesh: The prefix classname is to make sure we can confidently override the styles on this one and is a way to let builders know that they should not change these I believe. `ch` stands for `chime` :)

src/components/sdk/ContentShare/index.tsx
```diff
@@ -10,37 +10,49 @@ import { BaseSdkProps } from '../Base';
...
 
-  return tileId ? (
+  return tileIdToRender ? (
     <ContentTile
       objectFit="contain"
-      className={className || ''}
+      className={`ch-content-share--${tileIdToRender} ${className || ''}`}
       {...rest}
       ref={videoEl}
+      data-content-share-attendee={attendeeId}
```
devalevenkatesh, line 55: Is this for testing?

src/components/sdk/ContentShare/index.tsx
xuesichao: In short, I decide to let builder to decide how and where to render the multiple content share tiles via the new `tileId` of `ContentShare` component. Check code example in its documentation for example.

src/components/sdk/ContentShare/index.tsx
xuesichao: yeah, this is to make it easier for builder to know which tile is for which attendee so they can locate the element quickly.

src/providers/ContentShareProvider/docs/ContentShareProvider.mdx
ziyiz-amzn: if there an use case that builder can implement to kick off the 1st sharing when the 3rd starts? In this case, even though it reaches the max limit, but it literally is allowed to start content share.

xuesichao: Today we don't allow three content shares from media side. So they cannot start a 3rd one, but they can stop  one content share first, then `canStartContentShare` becomes true and then start the 3rd (which is in fact the second content share stream in the meeting)

src/providers/ContentShareProvider/state.tsx
ziyiz-amzn: I think const `updatedStateForBackwardCompatibility` and `updateForBackwardCompatibility()` best describe the purpose, but they are a little verbose.

ziyiz-amzn:
lgtm overall.

src/providers/ContentShareProvider/docs/ContentShareProvider.mdx
```diff
@@ -22,8 +36,17 @@ The `ContentShareProvider` provides state and functionality for content sharing.
   // Whether or not the local user's content share is loading
   isLocalShareLoading: boolean;
 
-  // The chime attendee ID of the user sharing
-  sharingAttendeeId: string | null;
+  // Whether the user can start a content share based on current limits and `isLocalUserSharing`
```
devalevenkatesh, line 39: We can expand a bit more on "current limits" for clarity here. Is this just the `maxContentShares`?

xuesichao: Yeah, let me update it, it should be `maxContentShares`, current limits is not very clear.

src/providers/ContentShareProvider/docs/ContentShareProvider.mdx
```diff
@@ -63,7 +128,7 @@ import { ContentShareProvider } from 'amazon-chime-sdk-component-library-react';
 
 const App = () => (
   <CustomAudioVideoProvider>
-    <ContentShareProvider>
+    <ContentShareProvider maxContentShares={2}>
```
devalevenkatesh, line 131: Its fine to drop the drop from here as its covered under the example updated above.

src/providers/ContentShareProvider/docs/useContentShareControls.mdx
```diff
@@ -13,6 +13,17 @@ The `useContentShareControls` hook returns the state and functionality around st
   // Whether or not the local content share is paused
   paused: boolean;
 
+  // Whether or not the local user is currently sharing content
+  isLocalUserSharing: boolean;
+
+  // Whether or not the local user's content share is loading
+  isLocalShareLoading: boolean;
```
devalevenkatesh, line 20: What is the need for this one, do we need to really add this? For UX?

src/providers/ContentShareProvider/docs/useContentShareControls.mdx
```diff
@@ -13,6 +13,17 @@ The `useContentShareControls` hook returns the state and functionality around st
   // Whether or not the local content share is paused
   paused: boolean;
 
+  // Whether or not the local user is currently sharing content
+  isLocalUserSharing: boolean;
```
devalevenkatesh, line 17: Can these be accessed from `useContentShareState` and we do not add them in this hook as well? The idea seems like this provides actions on content share vs the other one providing the state. I do see we have `paused` here, but checking if the other hook works or we need them repeated at two places.

src/providers/ContentShareProvider/docs/useContentShareState.mdx
```diff
@@ -57,6 +75,167 @@ const MyChild = () => {
...
+      <div className="content-share-container">
+        <p>No content is being shared</p>
+        <p>Can start content share: {canStartContentShare ? 'Yes' : 'No'}</p>
+      </div>
+    );
+  }
+
+  // Content is being shared
+  const contentShareTileId = tiles[0] // Or deprecated `tileId`
+  const attendeeId = attendeeIdToTileId[contentShareTileId.toString()] // Or deprecated `sharingAttendeeId`
```
devalevenkatesh, line 124: Can the TileId's be `number`s and the key in `attendeeIdToTileId` be `number` to avoid this `toString()` explicit conversions?

src/providers/ContentShareProvider/docs/useContentShareState.mdx
```diff
@@ -57,6 +75,167 @@ const MyChild = () => {
...
+
+  return (
+    <div className="content-share-container">
+      <div className="content-share-status">
+        <h3>Content Share Status</h3>
+        <ul>
+          <li>Active content shares: {tiles.length}</li>
+          <li>Current content share ID: {contentShareTileId}</li>
+          <li>Shared by: {attendeeId[]}</li>
```
devalevenkatesh, line 133: Have to correct to -> `attendeeId`?

src/providers/ContentShareProvider/state.tsx
```diff
@@ -11,71 +13,177 @@ export enum ContentActionType {
...
+// Action types
 export type Action =
-  | StartingAction
-  | DidStopAction
-  | UpdateAction
-  | TogglePauseAction
-  | RemoveAction
-  | DeniedAction
-  | ResetAction;
+  | { type: ContentActionType.STARTING; payload?: StartingPayload }
```
devalevenkatesh, line 42: We are updating some of the types for the payload here. What if someone is depending on this component currently and it starts failing for them as the types updated. Is this a backward incompatible change?

src/providers/ContentShareProvider/state.tsx
```diff
@@ -11,71 +13,177 @@ export enum ContentActionType {
...
+  state: ContentShareState,
+  tileId: number,
+  attendeeId: string
+): {
+  tiles: number[];
+  tileIdToAttendeeId: TileMap;
+  attendeeIdToTileId: AttendeeMap;
+} => {
+  return {
+    tiles: [...state.tiles, tileId],
```
devalevenkatesh, line 104: How about we add the new `tileId` at the start? Would it help us in not calculating the max always?

src/providers/RemoteVideoTileProvider/state.tsx
```diff
@@ -1,18 +1,12 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-type tileMap = {
-  [key: string]: string;
```
devalevenkatesh, line 5: Since this is tileId -> attendeeId so  `[key: number]: string;`? Will save those `toString()`s.

src/providers/ContentShareProvider/docs/useContentShareState.mdx
xuesichao: Yeah object key can be a number. But this how RemoteVideoTileProvider is doing it, I decide to follow the same.

src/providers/ContentShareProvider/docs/useContentShareControls.mdx
xuesichao: These are already available in this hook, we simply forgot updating the type definition and documentation 5 years ago. The only new property is `canStartContentShare`

src/providers/ContentShareProvider/state.tsx
xuesichao: These are internal type, so it's fine. It's `export` here but not `export` in index.ts file. It's not public available.

src/providers/ContentShareProvider/docs/useContentShareState.mdx
```diff
@@ -57,6 +75,167 @@ const MyChild = () => {
...
+
+      <div className="content-share-view">
+        <ContentShare nameplate={`Shared by: ${attendeeId}`} />
+      </div>
+    </div>
+  );
+};
+```
+
+### With Multiple Content Shares
```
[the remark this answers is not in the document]
xuesichao: I think that's fine, we don't want to encourage builders to use deprecated properties after this change.

I improved the comment a little bit to make it more clear
```
  // Get the content share tile ID
  const contentShareTileId = tiles[0]; // Note: Previously known as `tileId` (deprecated)

  // Attendee Id for content sharer
  const attendeeId = attendeeIdToTileId[contentShareTileId.toString()]; // Note: Previously known as `sharingAttendeeId` (deprecated)
  ```

File: src/providers/ContentShareProvider/state.tsx (hunk 17 above)
Comment (xuesichao) [reply src/providers/ContentShareProvider/state.tsx:104]: That would work too. But I'd prefer existing explicit check because 
1. in JS SDK video tile id is always increasing
    https://github.com/aws/amazon-chime-sdk-js/blob/320ba8446744b756b2deab9d1254704a7f400a3d/src/videotilecontroller/DefaultVideoTileController.ts#L179-L182

2. Most simple approach, eliminate the need to maintain this in data structure. I think the most promising data structure should be LRU. But it's overkill here since we only have 2 content shares and it's already handled in JS SDK.

3. we are using this check today in line 60 below: https://github.com/aws/amazon-chime-sdk-component-library-react/blob/efb7f0f4387a0929a1e2c9813a0afc57df1e2286/src/providers/ContentShareProvider/index.tsx#L57-L64
```

```

File: src/providers/ContentShareProvider/docs/useContentShareState.mdx (hunk 15 above)
Comment (xuesichao) [reply src/providers/ContentShareProvider/docs/useContentShareState.mdx]: updated

File: src/providers/ContentShareProvider/docs/ContentShareProvider.mdx (hunk 11 above)
Comment (xuesichao) [reply src/providers/ContentShareProvider/docs/ContentShareProvider.mdx]: updated

File: src/providers/ContentShareProvider/state.tsx (hunk 6 above)
Comment (xuesichao) [reply src/providers/ContentShareProvider/state.tsx]: updated to a better naming in my opinion

**Review by devalevenkatesh — DISMISSED** [review]

File: CHANGELOG.md (hunk 4 above)
Comment (devalevenkatesh) [reply CHANGELOG.md:14]: Right I get it now, its more like a feature flag in a way where its honored and backward compatible in the `ContentShareProvider`.

File: src/providers/ContentShareProvider/docs/ContentShareProvider.mdx [hunk 20]
```diff
@@ -6,13 +6,27 @@ import { Meta } from '@storybook/blocks';
...
+}
+```
+
 ### State
 
 ```javascript
 {
-  // The tile ID of the active content share
+  // The tile ID of the active content share (deprecated, maintained for backward compatibility)
+  // When multiple content shares are present, this points to the most recently started content share
```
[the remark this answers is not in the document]
devalevenkatesh: Yes it seems now the active tiles will be under `tiles` and current tileId -> tiles[0] where tiles will have a single content share to preserve backward compatibility.

src/providers/ContentShareProvider/docs/useContentShareState.mdx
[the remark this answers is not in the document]
devalevenkatesh: Hmm, that makes sense, this would remove how one might try to use with existing `tileId`. I was more caring towards explaining that in more details with explicit examples but this works too.

src/providers/ContentShareProvider/state.tsx
devalevenkatesh: Cool no issue!

src/providers/ContentShareProvider/docs/ContentShareProvider.mdx
xuesichao: Cool, I'll go with option 1 now. We can always add `hasReachedContentShareLimit` if we find it needed later.

src/providers/ContentShareProvider/docs/useContentShareState.mdx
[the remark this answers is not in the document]
xuesichao: Yeah, we add deprecated annotation to them. So when trying to use them it looks like this now:
image

ziyiz-amzn (approved):

src/providers/ContentShareProvider/docs/ContentShareProvider.mdx
devalevenkatesh: Seems like still needs an update?

PR #1035, opened 2026-07-15
feat(devices): add standalone DeviceManager (Part 1 of standalone device layer)

aryanbakshi25:
## Context
  
Team-sanctioned work for the standalone-device-layer milestone (internal design review completed with the owning team). Targets the `standalone-device-layer` feature branch, not `main`. This is Part 1 of a multi-part series; nothing merges to `main` until the full layer is built and tested.

## What

  Adds `DeviceManager` — the device-layer class for the Option 3 standalone device
  layer (Milestone 1). It owns a standalone `DefaultDeviceController` plus all device
  state, observers, and methods lifted from `MeetingManager`, so device setup
  (enumerate / select / preview / mic-level meter / permissions) works **before,
  during, and after** a meeting — and with no `MeetingProvider` mounted at all.

  Key APIs:
  - **`getController()`** — lends the controller to the session layer (`MeetingManager`)
    to build a `DefaultMeetingSession`; the session *borrows*, never owns or destroys it.
  - **`setupDevices()`** — pre-meeting entry point (label trigger + enumerate + select).
  - **`releaseMedia()`** — stops mic/camera and clears input selections (does not
    `destroy()` the controller; it stays reusable).

  ## Scope

  Additive only — `DeviceManager` is **not yet wired** to any provider or consumer
  (that's a later part), so existing behavior is unchanged. This PR does **not**
  include `DeviceProvider`, consumer re-pointing, or the `MeetingManager` reduction.

  ## Design note

  All controller calls are funneled through `DeviceManager` methods (a single choke
  point), so a future JS SDK `DeviceController` API change is a one-file update rather
  than a change at every call site (per design review).
  
  ## Testing

  - `npm run build` → `created lib/index.js, lib/index.esm.js`
  - `npm run test` → 60 suites / 252 tests pass
  - Ran a high-effort code review; 3 findings fixed, rest deferred to the wiring parts.

  ## Not included yet
  
  Dedicated `DeviceManager` unit tests (planned for the tests part).

  Part 1 of the `standalone-device-layer` work; targets the feature branch.

### DIFF
--- a/src/providers/DeviceProvider/DeviceManager.ts
+++ b/src/providers/DeviceProvider/DeviceManager.ts
@@ -1,0 +1,538 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+  AudioInputDevice,
+  DefaultBrowserBehavior,
+  DefaultDeviceController,
+  DeviceChangeObserver,
+  Logger,
+  RemovableAnalyserNode,
+  VideoInputDevice,
+} from 'amazon-chime-sdk-js';
+
+import {
+  DeviceLabels,
+  DeviceLabelTrigger,
+  DeviceLabelTriggerStatus,
+} from '../../types';
+
+/**
+ * Callback a builder supplies (via the `DeviceProvider` `onDeviceReplacement` prop) to control
+ * which device is chosen when the currently selected audio input is lost / a device change occurs.
+ * Relocated from the `MeetingProvider`/`DevicesProvider` `onDeviceReplacement` prop.
+ */
+export type OnDeviceReplacement = (
+  nextDevice: string,
+  currentDevice: AudioInputDevice | undefined
+) => Promise<AudioInputDevice>;
+
+/**
+ * Options for {@link DeviceManager}. `enableWebAudio` is a **constructor-only** option on
+ * `DefaultDeviceController` (it decides whether Amazon Voice Focus is possible), so it is fixed
+ * for the lifetime of the controller this manager owns.
+ */
+export interface DeviceManagerOptions {
+  enableWebAudio?: boolean;
+  onDeviceReplacement?: OnDeviceReplacement;
+}
+
+function noOpDeviceLabelHook(): Promise<MediaStream> {
+  return Promise.resolve(new MediaStream());
+}
+
+/**
+ * `DeviceManager` is the device-layer counterpart to `MeetingManager`. It owns a standalone
+ * `DefaultDeviceController` **and** all device state / observers / methods, so device setup
+ * (enumerate / select / preview / mic-level meter / permissions) works **before, during, and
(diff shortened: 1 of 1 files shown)


src/providers/DeviceProvider/DeviceManager.ts
```diff
@@ -0,0 +1,538 @@
...
+
+  getDeviceLabels(): DeviceLabels | DeviceLabelTrigger {
+    return this.deviceLabels;
+  }
+
+  /**
+   * Pre-meeting entry point: install the permission/label trigger, then enumerate and
+   * default-select devices — all with no `MeetingSession`. Safe to call before a meeting exists.
+   */
+  async setupDevices(
```
xuesichao, line 129: Do we need this new API?

src/providers/DeviceProvider/DeviceManager.ts
```diff
@@ -0,0 +1,538 @@
...
+) => Promise<AudioInputDevice>;
+
+/**
+ * Options for {@link DeviceManager}. `enableWebAudio` is a **constructor-only** option on
+ * `DefaultDeviceController` (it decides whether Amazon Voice Focus is possible), so it is fixed
+ * for the lifetime of the controller this manager owns.
+ */
+export interface DeviceManagerOptions {
+  enableWebAudio?: boolean;
+  onDeviceReplacement?: OnDeviceReplacement;
```
xuesichao, line 37: 1. Could you explain why we need `onDeviceReplacement` in device manager?
2. Should we make this a setter API instead of a constructor params if we need it?

src/providers/DeviceProvider/DeviceManager.ts
```diff
@@ -0,0 +1,538 @@
...
+   *
+   * Declared as an arrow property (like the other device methods) so it can be passed as a bare
+   * callback (e.g. a React effect cleanup) without losing `this`.
+   *
+   * Audio-output (speaker) teardown is intentionally NOT done here. `chooseAudioOutput(null)` only
+   * *selects the default output device* (per the JS SDK) — it releases no hardware. The real speaker
+   * unbind is `audioVideo.unbindAudioElement()`, a session-layer concern handled by
+   * `MeetingManager.leave()`, not the device layer.
+   */
+  releaseMedia = async (): Promise<void> => {
```
xuesichao, line 156: Could we rename this API? It's low level. Ideally react component is at "device" level not "media stream" level. Let's keep it consistent with other api naming.

src/providers/DeviceProvider/DeviceManager.ts
```diff
@@ -0,0 +1,538 @@
...
+      if (error instanceof Error) {
+        newError.name = error.name;
+        newError.message += ' ' + error.message;
+      }
+      this.logger.error(newError.toString());
+      throw newError;
+    }
+  };
+
+  stopVideoInputDevice = async (): Promise<void> => {
```
xuesichao, line 367: Do we need `stopAudioInputDevice()` API? and it's observer and publish?

PR #1037, opened 2026-07-21
feat: Migrate BackgroundBlur and BackgroundReplacement providers to use BackgroundSegmentationProcessor

harshhv-dev:
**Issue #:** 
  
Add new unified `BackgroundSegmentationProvider` and `useBackgroundSegmentation` hook with improved segmentation quality, native color/image replacement, and runtime effect switching — while keeping existing V1 providers unchanged for backward compatibility.
  
**Description of changes:**
  
Core:
  
- Add `BackgroundSegmentationProvider` and `useBackgroundSegmentation` hook wrapping the new `BackgroundSegmentationVideoFrameProcessor` (CDN-loaded, WebGL2-based)
- Unified provider handles blur, color-replacement, and image-replacement in a single processor instance
- Auto-mounted inside `MeetingProvider` below `DevicesProvider` (zero-cost until consumer activates an effect)
- `createSegmentationDevice(device, config, options?)` creates a `VideoTransformDevice` with the specified effect, model type, and CPU budget at creation time
- `updateEffect()` switches between blur/color-replacement/image-replacement at runtime via `setConfig()` — no blank frame, no processor recreation
- `updateModelType()` switches between `SELFIE_GENERAL` and `SELFIE_MULTICLASS` at runtime
- `updateCPUUsagePercentage()` adjusts CPU budget (10-80%) at runtime
- `isSupported` state reflects browser compatibility only (WebGL2, WASM, Workers, OffscreenCanvas) — set once on mount
- Existing V1 `BackgroundBlurProvider` and `BackgroundReplacementProvider` remain 100% unchanged (no breaking changes)
- Consumers use the new hook by calling `useBackgroundSegmentation()` within `MeetingProvider` — no explicit provider wrapping needed
- Only `useBackgroundSegmentation` hook exported publicly (matching V1 pattern)
  
Testing:
  
- Add unit tests for `BackgroundSegmentationProvider` (zero-cost mount, unsupported environment handling, method exposure, cleanup on unmount, hook-outside-provider error)
- Add `RTCRtpSender`/`RTCRtpReceiver` mock in setupTests.js for JSSDK 3.32.0 JSDOM compatibility
- Bump `amazon-chime-sdk-js` peer dependency to ^3.32.0
  
Documentation:
  
- Add Storybook MDX for `BackgroundSegmentationProvider` and `useBackgroundSegmentation`
- CHANGELOG.md updated under [3.13.0]
  
**Testing**
1. Have you successfully run `npm run build:release` locally? **Yes**
  
2. How did you test these changes? **Using React SDK meeting demo app with consolidated single-dropdown UI showcasing all 10 effect options (6 blur + 4 replacement across General and Multiclass models). Tested seamless switching, None reset, preview-to-meeting persistence, CPU utilization control, and model type switching.**
  
3. If you made changes to the component library, have you provided corresponding documentation changes? **Yes — Storybook MDX docs added for provider and hook. CHANGELOG.md updated under [3.13.0] with Added/Changed sections.**
  

hensmi-amazon (dismissed):

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -3,18 +3,20 @@
 
 import {
   BackgroundBlurOptions,
-  BackgroundBlurProcessor,
-  BackgroundBlurVideoFrameProcessor,
   BackgroundFilterSpec,
+  BackgroundSegmentationVideoFrameProcessor,
+  BackgroundSegmentationBlurStrength as BlurStrength,
```
devalevenkatesh: Is `BackgroundSegmentationBlurStrength` not working and we have to cast it to `BlurStrength`?

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -24,29 +26,54 @@ import useMemoCompare from '../../utils/use-memo-compare';
 import { useLogger } from '../LoggerProvider';
 
 interface Props extends BaseSdkProps {
-  /** The spec defines the assets that will be used for adding background blur to a frame. For more information, refer to
-   * [Amazon Chime SDK for JavaScript Background Filter Guide](https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/15_Background_Filter_Video_Processor.md#adding-a-background-filter-to-your-application). */
+  /** The spec defines the assets that will be used for adding background blur to a frame.
+   * @deprecated Background blur processor (V3) processor handles asset loading internally. This prop is accepted but ignored. */
```
devalevenkatesh: nit: `@deprecated Background blur processor (V3) handles asset loading internally. This prop is accepted but ignored.`

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -24,29 +26,54 @@ import useMemoCompare from '../../utils/use-memo-compare';
...
+  /** The segmentation model type to use. Defaults to SELFIE_GENERAL. */
+  modelType?: ModelType;
 }
 
 interface BackgroundBlurProviderState {
   createBackgroundBlurDevice: (
     device: Device
   ) => Promise<DefaultVideoTransformDevice>;
   isBackgroundBlurSupported: boolean | undefined;
-  backgroundBlurProcessor: BackgroundBlurProcessor | undefined;
```
devalevenkatesh, line 40: So we are removing an accepted type here so isn't this a breaking change? What if some customers are passing in `BackgroundBlurProcessor`?

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -24,29 +26,54 @@ import useMemoCompare from '../../utils/use-memo-compare';
...
 
 const BackgroundBlurProviderContext = createContext<
   BackgroundBlurProviderState | undefined
 >(undefined);
 
+/**
+ * Maps numeric blur strength to BlurStrength enum for Background blur processor (V3).
+ */
+function mapBlurStrength(numBlurStrength?: number): BlurStrength {
+  if (numBlurStrength === undefined || numBlurStrength === null) {
```
devalevenkatesh: nit: No need to update but simple truthy check works too. `if (numBlurStrength) {..}`

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -93,56 +109,72 @@ export const BackgroundBlurProvider: FC<React.PropsWithChildren<Props>> = ({
...
+    try {
+      const createdProcessor =
+        await BackgroundSegmentationVideoFrameProcessor.create(
+          logger,
+          { type: ProcessorEffect.BLUR, blurStrength },
+          modelType,
+          cpuUsagePercentage
+        );
+
+      if (!createdProcessor.getConfig()) {
```
devalevenkatesh: Guess too late since the JS SDK changes are merged, but wonder this could be explicit with `isSupported` API. We have such an API for the VoiceFocus in JS SDK. [Reference](https://aws.github.io/amazon-chime-sdk-js/modules/amazonvoice_focus.html#checking-for-support-before-offering-noise-suppression-and-echo-reduction)

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -93,56 +109,72 @@ export const BackgroundBlurProvider: FC<React.PropsWithChildren<Props>> = ({
...
+        logger.warn(
+          'Background blur processor (V3) created but browser is not supported.'
+        );
         setBackgroundBlurProcessor(undefined);
         setIsBackgroundBlurSupported(false);
+        processorRef.current = undefined;
```
devalevenkatesh: After this we have to destroy the created processor too?

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -93,56 +109,72 @@ export const BackgroundBlurProvider: FC<React.PropsWithChildren<Props>> = ({
...
         return undefined;
-      } else {
-        logger.info('Initialized background blur processor');
-        setBackgroundBlurProcessor(createdProcessor);
-        setIsBackgroundBlurSupported(true);
-        return createdProcessor;
       }
+
+      logger.info('Successfully initialized background blur processor (V3).');
+      setBackgroundBlurProcessor(createdProcessor);
```
devalevenkatesh: Why do we need the state if we are going ahead with the `ref`?

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -24,21 +26,31 @@ import useMemoCompare from '../../utils/use-memo-compare';
...
-   * [Amazon Chime SDK for JavaScript Background Filter Guide](https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/15_Background_Filter_Video_Processor.md#adding-a-background-filter-to-your-application). */
+  /** The spec defines the assets that will be used for adding background replacement to a frame.
+   * @deprecated Background replacment processor (V3) handles asset loading internally. This prop is accepted but ignored. */
   spec?: BackgroundFilterSpec;
-  /* A set of options that can be supplied when creating a background replacement video frame processor such as the background replacement image blob. For more information, refer to
-   * [Amazon Chime SDK for JavaScript Background Filter Guide](https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/15_Background_Filter_Video_Processor.md#adding-a-background-filter-to-your-application). */
-  options?: BackgroundReplacementOptions;
+  /** A set of options that can be supplied when creating a background replacement video frame processor. */
+  options?: BackgroundReplacementOptions & {
```
devalevenkatesh: Any reason for ` & {..}`? Can we add the `replacementColor` to `BackgroundReplacementOptions`?

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -24,21 +26,31 @@ import useMemoCompare from '../../utils/use-memo-compare';
...
+  /** The segmentation model type to use. Defaults to SELFIE_GENERAL. */
+  modelType?: ModelType;
 }
 
 interface BackgroundReplacementProviderState {
   createBackgroundReplacementDevice: (
     device: Device
   ) => Promise<DefaultVideoTransformDevice>;
   isBackgroundReplacementSupported: boolean | undefined;
-  backgroundReplacementProcessor: BackgroundReplacementProcessor | undefined;
```
devalevenkatesh, line 40: Checking if this type update would break customer's usage in any way. Same comment on the `BackgroundBlurProvider`. Typically type changes like these are considered backward incompatible.

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -91,55 +95,107 @@ export const BackgroundReplacementProvider: FC<
...
+      };
+    } else if (options?.replacementColor) {
+      config = {
+        type: ProcessorEffect.COLOR_REPLACEMENT,
+        replacementColor: options.replacementColor,
+      };
+    } else {
+      config = {
+        type: ProcessorEffect.COLOR_REPLACEMENT,
+        replacementColor: '#0000FF',
```
devalevenkatesh: nit: Add `DEFAULT_REPLACMENT_COLOR`

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -91,55 +95,107 @@ export const BackgroundReplacementProvider: FC<
...
         );
-      // BackgroundReplacementVideoFrameProcessor.create will return a NoOpVideoFrameProcessor
-      // in the case that BackgroundReplacementVideoFrameProcessor.isSupported() returns false.
-      // BackgroundReplacementVideoFrameProcessor.create() can also throw an error in case loading
-      // the assets are not fetched successfully.
-      if (createdProcessor instanceof NoOpVideoFrameProcessor) {
-        logger.warn('Initialized NoOpVideoFrameProcessor');
         setBackgroundReplacementProcessor(undefined);
         setIsBackgroundReplacementSupported(false);
+        processorRef.current = undefined;
```
devalevenkatesh: Needs createdProcessor destroy?

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -91,55 +95,107 @@ export const BackgroundReplacementProvider: FC<
...
+        objectUrlRef.current = undefined;
+      }
     };
-  }, [replacementSpec, replacementOptions]);
+  }, [replacementOptions]);
+
+  /**
+   * Converts a imageBlob to an object URL for Background Replacement processor (V3).
+   */
+  function blobToObjectURL(blob: Blob): string {
```
devalevenkatesh: Does this need any error handling? May be CSP blocks this or CSP is not of concern here or if this fails do we can gracefully handle or bubble this up?

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -197,11 +257,49 @@ export const BackgroundReplacementProvider: FC<
     }
   };
 
+  const changeBackgroundReplacementColor = async (
+    color: string
+  ): Promise<void> => {
+    if (!processorRef.current) {
+      logger.warn(
```
devalevenkatesh: Should these `change..` APIs need to throw errors or silent failing is expected?

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -55,21 +82,11 @@ export const BackgroundBlurProvider: FC<React.PropsWithChildren<Props>> = ({
...
-      prev: BackgroundFilterSpec | undefined,
-      next: BackgroundFilterSpec | undefined
-    ): boolean => {
-      if (Object.is(prev, next)) {
-        return true;
-      }
-      return false;
-    }
-  );
+  const processorRef = useRef<
```
devalevenkatesh: For understanding: Why a `ref` over the `state`. If any state changes take place with `useEffect`'s we can listen on that states and update/call processors APIs hence checking. Have commented somewhere on whether we need the `useState` then if using `useRef`.

CHANGELOG.md
```diff
@@ -7,6 +7,26 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
ltrung: This should be under 3.13 below. The current development version is 3.13 and we dont have unreleased section in CHANGELOG.

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -24,29 +26,54 @@ import useMemoCompare from '../../utils/use-memo-compare';
...
-   * [Amazon Chime SDK for JavaScript Background Filter Guide](https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/15_Background_Filter_Video_Processor.md#adding-a-background-filter-to-your-application). */
+  /** The spec defines the assets that will be used for adding background blur to a frame.
+   * @deprecated Background blur processor (V3) processor handles asset loading internally. This prop is accepted but ignored. */
   spec?: BackgroundFilterSpec;
-  /** A set of options that can be supplied when creating a background blur video frame processor. For more information, refer to
-   * [Amazon Chime SDK for JavaScript Background Filter Guide](https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/15_Background_Filter_Video_Processor.md#adding-a-background-filter-to-your-application). */
+  /** A set of options that can be supplied when creating a background blur video frame processor (V3). */
   options?: BackgroundBlurOptions;
+  /** The segmentation model type to use. Defaults to SELFIE_GENERAL. */
+  modelType?: ModelType;
```
ltrung: What are other value here? Can we link to JS SDK doc if there is one?

CHANGELOG.md
```diff
@@ -7,6 +7,26 @@ All notable changes to this project will be documented in this file.
...
+
+### Added
+
+- Added `changeBlurStrength()` and `changeModelType()` methods to `useBackgroundBlur` hook for seamless runtime switching without blank frame.
```
ltrung: Is there a way to make this backward compatible in case customers want to keep old blur/replacement behavior? Simplest way is just to create blurProviderV2

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -24,29 +26,54 @@ import useMemoCompare from '../../utils/use-memo-compare';
...
 
 interface Props extends BaseSdkProps {
-  /** The spec defines the assets that will be used for adding background blur to a frame. For more information, refer to
-   * [Amazon Chime SDK for JavaScript Background Filter Guide](https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/15_Background_Filter_Video_Processor.md#adding-a-background-filter-to-your-application). */
+  /** The spec defines the assets that will be used for adding background blur to a frame.
+   * @deprecated Background blur processor (V3) processor handles asset loading internally. This prop is accepted but ignored. */
   spec?: BackgroundFilterSpec;
-  /** A set of options that can be supplied when creating a background blur video frame processor. For more information, refer to
-   * [Amazon Chime SDK for JavaScript Background Filter Guide](https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/15_Background_Filter_Video_Processor.md#adding-a-background-filter-to-your-application). */
+  /** A set of options that can be supplied when creating a background blur video frame processor (V3). */
```
ziyiz-amzn: Do we have a link to these options? else, may need an internal TODO item.

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -24,29 +26,54 @@ import useMemoCompare from '../../utils/use-memo-compare';
...
   createBackgroundBlurDevice: (
     device: Device
   ) => Promise<DefaultVideoTransformDevice>;
   isBackgroundBlurSupported: boolean | undefined;
-  backgroundBlurProcessor: BackgroundBlurProcessor | undefined;
+  backgroundBlurProcessor:
+    | BackgroundSegmentationVideoFrameProcessor
+    | undefined;
+  changeBlurStrength: (strength: BlurStrength) => void;
+  changeModelType: (newModelType: ModelType) => void;
```
ziyiz-amzn: Are they required or can be optional? Also, prefer `update` than `change`. e.g updateBlurStrength

src/providers/BackgroundBlurProvider/index.tsx
ziyiz-amzn: please make sure to not break the existing customer who is using the `spec`.

src/providers/BackgroundBlurProvider/index.tsx
ziyiz-amzn: +1, this has to be resolved.

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -93,56 +109,72 @@ export const BackgroundBlurProvider: FC<React.PropsWithChildren<Props>> = ({
...
-    // One reason we need to initialize first, even though we'll destroy this background blur processor when we create a new device
-    // is because we need to check if background blur is supported by initializing the background blur processor to see if the browser supports
     initializeBackgroundBlur();
     return () => {
       logger.info(
-        'Specs or options were changed. Destroying and re-initializing background blur processor.'
+        'Options were changed. Destroying and re-initializing background blur processor (V3).'
       );
-      backgroundBlurProcessor?.destroy();
+      processorRef.current?.destroy();
```
ziyiz-amzn: avoid using a too general var name

src/providers/BackgroundBlurProvider/index.tsx
```diff
@@ -93,56 +109,72 @@ export const BackgroundBlurProvider: FC<React.PropsWithChildren<Props>> = ({
...
-      // BackgroundBlurVideoFrameProcessor.create will return a NoOpVideoFrameProcessor
-      // in the case that BackgroundBlurVideoFrameProcessor.isSupported() returns false.
-      // BackgroundBlurVideoFrameProcessor.create() can also throw an error in case loading
-      // the assets are not fetched successfully.
-      if (createdProcessor instanceof NoOpVideoFrameProcessor) {
-        logger.warn('Initialized NoOpVideoFrameProcessor');
+    }
+
+    try {
+      const createdProcessor =
```
ziyiz-amzn: too general var name, what is the created processor? what is it for?

src/providers/BackgroundReplacementProvider/index.tsx
```diff
@@ -47,42 +59,34 @@ const BackgroundReplacementProviderContext = createContext<
...
+      prev: (BackgroundReplacementOptions & { replacementColor?: string }) | undefined,
+      next: (BackgroundReplacementOptions & { replacementColor?: string }) | undefined
     ): boolean => {
       if (isPrevNextEmpty(prev, next) || isPrevNextUndefined(prev, next)) {
         return true;
       }
       if (
         prev?.filterCPUUtilization !== next?.filterCPUUtilization ||
-        prev?.logger !== next?.logger ||
-        prev?.reportingPeriodMillis !== next?.reportingPeriodMillis
```
ziyiz-amzn, line 85: is `reportingPeriodMillis` a public option builder can set? if so, it's a breaking change.

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: Addressed — JSSDK already exposes `BackgroundSegmentationCompatibilityChecker.checkCompatibility(logger)` which returns `{ isCompatible: boolean, missingFeatures: string[] }`. Updated the V2 providers to call this before processor creation rather than creating and then checking `getConfig()`. This avoids wasted processor instantiation and provides explicit missing feature logging for debugging.

src/providers/BackgroundReplacementProvider/index.tsx
harshhv-dev: V2 methods warn and return early (matching V1's `NoOp` behavior). Throwing would break consumers who call these without checking `isBackgroundBlurSupported` first. The warn gives visibility without crashing.

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: Comment no longer applies. Since we are not deprecating anything and introducing separate cleaner hooks/providers for these new blur/replacement improvements.

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: Comment no longer applies -  V1 providers remain completely unchanged. V2 providers are additive — new files with their own types. No breaking changes.

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: Comment no longer applies - `mapBlurStrength()` was removed. V2 accepts BlurStrength enum directly from consumers — no mapping needed.

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: Ack. Addressed — V2 provider now calls `segmentationProcessor.destroy()` before attempting V1 fallback when the segmentation processor is unsupported.

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: This is just an import alias. JSSDK exports the enum as BackgroundSegmentationBlurStrength (verbose name to distinguish from V1's BackgroundBlurStrength and V2's VideoFxBlurStrength). I used alias as BlurStrength locally for readability and consistency with JSSDK. The enum values are the same — no runtime conversion happens. But in ReactSDK, we will eventually use only one processor and hence one type of blur strength. So it makes sense to just use the exact naming. 

Corrected the import.

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: Addressed — removed `backgroundBlurProcessor`/`backgroundReplacementProcessor` state from the hook return value entirely. Consumers interact through the provided methods `(createBackgroundBlurDevice, updateBlurStrength, updateModelType)` rather than accessing the processor directly. 

Internally, only `useRef` is used for the processor (synchronous access needed by `update*` methods). The only remaining `useState` is `isBackgroundBlurSupported`/`isBackgroundReplacementSupported` which consumers need for reactive UI updates (e.g., conditionally rendering blur/replacement controls).

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: Addressed — removed `backgroundBlurProcessor`/`backgroundReplacementProcessor` state from the hook return value entirely. Consumers interact through the provided methods `(createBackgroundBlurDevice, updateBlurStrength, updateModelType)` rather than accessing the processor directly. 

Internally, only `useRef` is used for the processor (synchronous access needed by `update*` methods). The only remaining `useState` is `isBackgroundBlurSupported`/`isBackgroundReplacementSupported` which consumers need for reactive UI updates (e.g., conditionally rendering blur/replacement controls).

src/providers/BackgroundReplacementProvider/index.tsx
harshhv-dev: Comment no longer applies -  This no longer applies with the V2 provider approach. V2 defines its own props interface (`replacementColor?: string, replacementImageURL?: string)` directly — it doesn't reference `BackgroundReplacementOptions` from JSSDK, which is the V1 processor's options type and doesn't support color replacement.

src/providers/BackgroundReplacementProvider/index.tsx
harshhv-dev: Comment no longer applies - V1 providers remain completely unchanged. V2 providers are additive — new files with their own types. No breaking changes.

src/providers/BackgroundReplacementProvider/index.tsx
harshhv-dev: Ack. Addressed — V2 provider uses const `DEFAULT_REPLACEMENT_COLOR = '#0000FF'`.

src/providers/BackgroundReplacementProvider/index.tsx
harshhv-dev: Ack. Addressed — V2 provider now calls `segmentationProcessor.destroy()` before attempting V1 fallback when the segmentation processor is unsupported.

src/providers/BackgroundReplacementProvider/index.tsx
harshhv-dev: Comment no longer applies -  V2 provider accepts `replacementImageURL: string` directly — no blob conversion, no `URL.createObjectURL()` at all.

CHANGELOG.md
harshhv-dev: Ack. Moved to [3.13.0] section.

CHANGELOG.md
harshhv-dev: Ack. Adopted this approach — created `BackgroundBlurProviderV2` and `BackgroundReplacementProviderV2` as new additive providers. V1 providers remain 100% unchanged. Consumers opt-in to V2 when ready.

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: Comment no longer applies - V1 providers are completely untouched — spec prop continues to work as before. V2 providers are new and don't accept spec (not needed  since the new processor loads assets from CDN internally).

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: No longer applies — V1 providers are unchanged, and V2 providers use descriptive names (`segmentationProcessor`, `v1Processor`).

src/providers/BackgroundReplacementProvider/index.tsx
harshhv-dev: No longer applies — V1 providers remain unchanged (still accept `reportingPeriodMillis`). V2 providers are new and don't include it in the interface.

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: No longer applies — V1 providers are unchanged, and V2 providers use descriptive names (`segmentationProcessor`, `v1Processor`).

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: The methods are always present on the hook return value (not optional). They handle the fallback case internally — if the segmentation processor isn't active (V1 fallback), they log a warning and return. This avoids forcing consumers to null-check before every call. The hook guarantees a stable API surface regardless of which processor is active internally.

Although renamed methods to use `update` prefix instead of `change`.

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: No dedicated JSSDK guide yet — added comments for available values as of now. Will create guide as follow-up.

src/providers/BackgroundBlurProvider/index.tsx
harshhv-dev: No dedicated JSSDK guide yet — added comments for available values as of now. Will create guide as follow-up.

xuesichao:
1. I'm not sure about the "V2" naming here, a new `BackgroundSegmentationProcessor` seems a better fit.
2. Also what's our plan to `VideoInputBackgroundBlurControl.tsx`, are we going to migrate it or build new sdk component  new blur and replacement?
3. The story book update is missing in the PR.

src/providers/BackgroundBlurProviderV2/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+   * Available values: NONE, LOW, MEDIUM, HIGH. */
+  blurStrength?: BackgroundSegmentationBlurStrength;
+  /** The segmentation model type to use. Defaults to SELFIE_GENERAL.
+   * Available values: ModelType.SELFIE_GENERAL, ModelType.SELFIE_MULTICLASS. */
+  modelType?: ModelType;
+  /** Maximum CPU usage percentage for the processor. */
+  cpuUsagePercentage?: number;
+}
+
+interface BackgroundBlurV2ProviderState {
```
devalevenkatesh: nit: I agree with Sichao. The name could be `BackgroundSegmentationProvider` and relating components can be named accordingly. Customers can prefer to keep using current vs this. Any downstream components like Video control etc or who is maintaining any state for this blur would need to handle this switch in a backward compatible way. For example, here technically if we are going with current naming then the `BackgroundSegmentationBlurStrength` makes sense or just `blurStrength`?

src/providers/BackgroundBlurProviderV2/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+  createBackgroundBlurDevice: (
+    device: Device
+  ) => Promise<DefaultVideoTransformDevice>;
```
devalevenkatesh: Where applicable can we use interfaces over concrete classes. Here: `VideoTransformDevice`.

src/providers/BackgroundBlurProviderV2/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+    }
+  );
+
+  useEffect(() => {
+    initializeProcessor();
+    return () => {
+      logger.info(
+        'BackgroundBlurProviderV2: Destroying processor on unmount or props change.'
+      );
+      if (processorRef.current && 'destroy' in processorRef.current) {
```
devalevenkatesh: Why this `"destroy" in` is required?
I see `VideoFrameProcessor` has the `destroy` method. Is any processor not implementing the `destroy` method?

src/providers/BackgroundBlurProviderV2/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+        const segmentationProcessor =
+          await BackgroundSegmentationVideoFrameProcessor.create(
+            logger,
+            { type: ProcessorEffect.BLUR, blurStrength },
+            modelType,
+            cpuUsagePercentage
+          );
+
+        logger.info('BackgroundBlurProviderV2: Segmentation processor initialized successfully.');
+        setIsBackgroundBlurSupported(true);
```
devalevenkatesh: Since a processor in this component will always exist if supported then do we need the `isBackgroundBlurSupported` state?

src/providers/BackgroundBlurProviderV2/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+        isNewProcessorRef.current = true;
+        return segmentationProcessor;
+      } catch (error) {
+        logger.warn(`BackgroundBlurProviderV2: Segmentation processor creation failed (${error}), attempting V1 fallback.`);
+      }
+    } else {
+      logger.info(`BackgroundBlurProviderV2: Background blur not supported due to missing features: [${missingFeatures.join(', ')}]. Attempting V1 fallback.`);
+    }
+
+    // Fallback to BackgroundBlurVideoFrameProcessor
```
devalevenkatesh: Do we need this? customers in their React application can choose to mount either right? Ideally, if a fallback pattern is to be introduced should we not handle this in the JS SDK itself? any consumer of the JS SDK would have to do similar logic.

src/providers/BackgroundBlurProviderV2/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+    try {
+      const v1Processor = await BackgroundBlurVideoFrameProcessor.create(
```
devalevenkatesh: `isSupported` is available on the `BackgroundBlurVideoFrameProcessor`, so checking if any problems using that?

src/providers/BackgroundBlurProviderV2/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+        prev?.cpuUsagePercentage !== next?.cpuUsagePercentage
+      ) {
+        return false;
+      }
+      return true;
+    }
+  );
+
+  useEffect(() => {
+    initializeProcessor();
```
devalevenkatesh: `initializeProcessor` method is returning `Promise<VideoFrameProcessor | undefined>` but internally its setting `ref` and some state. So:
1. Do we need to return from method?
2. If returning, can the state move after catching whats returned?
```
const processor = createProcessor(..);
if (processor) {
  // set the ref
}
```

Once processor is set, not sure we really need the `isSupported` state since the creation is dependent on the support checks.

src/providers/BackgroundBlurProviderV2/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+    logger.warn('BackgroundBlurProviderV2: Background blur is not supported in this environment.');
+    setIsBackgroundBlurSupported(false);
+    processorRef.current = undefined;
+    return undefined;
+  }
+
+  const createBackgroundBlurDevice = async (
+    selectedDevice: Device
+  ): Promise<DefaultVideoTransformDevice> => {
```
devalevenkatesh: Lets use `VideoTransformDevice`.

src/providers/BackgroundBlurProviderV2/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+    logger.info(
+      `BackgroundBlurProviderV2: Creating blur device for: ${JSON.stringify(selectedDevice)}`
+    );
+    // Always re-initialize — the processor may have been destroyed externally
+    // by DefaultVideoTransformDevice.stop() when switching effects.
```
devalevenkatesh: In a React application can this happen?

src/providers/BackgroundBlurProviderV2/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+    const currentProcessor = await initializeProcessor();
+    if (currentProcessor) {
+      return new DefaultVideoTransformDevice(logger, selectedDevice, [currentProcessor]);
+    }
+    throw new Error('BackgroundBlurProviderV2: Processor not available. Background blur is not supported.');
```
devalevenkatesh: I think we are following two models in different places:
1. Throwing the error when processor is not supported
2. Setting isBackgroundBlurSupported

We can unify them into one approach.

src/providers/BackgroundBlurProviderV2/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+  modelType = DEFAULT_MODEL_TYPE,
+  cpuUsagePercentage,
+  children,
+}) => {
+  const logger = useLogger();
+  const [isBackgroundBlurSupported, setIsBackgroundBlurSupported] = useState<
+    boolean | undefined
+  >(undefined);
+  const processorRef = useRef<VideoFrameProcessor | undefined>(undefined);
+  const isNewProcessorRef = useRef<boolean>(false);
```
devalevenkatesh: Do we need this? All the checks done using this cannot be used or based on the `processorRef`?

src/providers/BackgroundBlurProviderV2/index.tsx
```diff
@@ -0,0 +1,222 @@
...
+  blurStrength = DEFAULT_BLUR_STRENGTH,
+  modelType = DEFAULT_MODEL_TYPE,
+  cpuUsagePercentage,
+  children,
+}) => {
+  const logger = useLogger();
+  const [isBackgroundBlurSupported, setIsBackgroundBlurSupported] = useState<
+    boolean | undefined
+  >(undefined);
+  const processorRef = useRef<VideoFrameProcessor | undefined>(undefined);
```
devalevenkatesh: I see that we explicitly set the processorRef on certain conditions so checking why its a ref over the state? Its fine but just want to understand the preference.

devalevenkatesh (dismissed):

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+`BackgroundSegmentationProvider` is auto-mounted inside `MeetingProvider` with `effect="none"`, so consumers can use `useBackgroundSegmentation()` without any explicit wrapping. The provider remains inert (zero-cost) until `createSegmentationDevice()` is called — no assets are loaded and no processor is created until then.
+
+Background segmentation related logs can be found in the browser developer tools when the provider is active.
+
+## State
+
+```typescript
+{
+  // Whether background segmentation is supported. The default value is undefined, then changes to true or false.
```
devalevenkatesh: Can help builders with some docs on what influences this since we will be setting this right?

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
```diff
@@ -0,0 +1,169 @@
...
+  isSupported: boolean | undefined;
+
+  // Create a video transform device with the current segmentation effect applied.
+  createSegmentationDevice: (device: Device) => Promise<VideoTransformDevice>;
```
devalevenkatesh: If the current effect on mount is "none" and one calls this method, whats the result?

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+  // Change the segmentation model type at runtime.
+  updateModelType: (modelType: ModelType) => void;
+
+  // Change the CPU usage percentage budget at runtime (range: 10-80).
+  updateCPUUsagePercentage: (percentage: number) => void;
+}
+```
+
+You can check whether or not the processor is supported by checking the state of `isSupported`.
```
devalevenkatesh: This can go above the `isSupported` and `createSegmentationDevice` docstrings co-relating such that builders can understand there itself over separate doc mention.

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
```diff
@@ -0,0 +1,169 @@
...
+```typescript
+{
+  // Whether background segmentation is supported. The default value is undefined, then changes to true or false.
+  isSupported: boolean | undefined;
+
+  // Create a video transform device with the current segmentation effect applied.
+  createSegmentationDevice: (device: Device) => Promise<VideoTransformDevice>;
+
+  // Switch the active effect at runtime (blur, color-replacement, image-replacement).
+  updateEffect: (config: BackgroundSegmentationVideoFrameProcessorConfig) => void;
```
devalevenkatesh, line 33: Can check other docs once, but all these imports can be shown to help builder understand whether its coming from this library or JS SDK.

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
```diff
@@ -0,0 +1,169 @@
...
+  const [isBlurOn, setIsBlurOn] = useState(false);
+
+  useEffect(() => {
+    async function toggleBlur() {
+      if (!selectedDevice) return;
+
+      try {
+        let current = selectedDevice;
+        if (isBlurOn) {
+          current = await createSegmentationDevice(selectedDevice);
```
devalevenkatesh: For the case, if `isBlurOn` is `true` and `selectedDevice` is a transform device then this still holds right? In the `else` below we are not updating the effect if its already a transform device hence checking.

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
```diff
@@ -0,0 +1,169 @@
...
+const MyChild = () => {
+  const meetingManager = useMeetingManager();
+  const { selectedDevice } = useVideoInputs();
+  const {
+    isSupported,
+    createSegmentationDevice,
+    updateEffect,
+    updateModelType,
+  } = useBackgroundSegmentation();
+  const [isBlurOn, setIsBlurOn] = useState(false);
```
devalevenkatesh: In this example for builder we should provide two states:
1. Clearly showcase UI state of user clicked the blur etc.
2. Actually whether the blur did apply and accordingly error can be show by that state.
3. If we want one state then that state should be only set to true when the blur application completed successfully (device creation, effect update success etc).

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
```diff
@@ -0,0 +1,169 @@
...
+          }
+        }
+        await meetingManager.startVideoInputDevice(current);
+      } catch (error) {
+        console.error('Failed to toggle background segmentation');
+      }
+    }
+
+    toggleBlur();
+  }, [isBlurOn]);
```
devalevenkatesh: Needs the dependency on the `selectedDevice`?

src/providers/BackgroundSegmentationProvider/docs/useBackgroundSegmentation.mdx
```diff
@@ -0,0 +1,146 @@
+import { Meta } from '@storybook/blocks';
+
+<Meta title="SDK Hooks/useBackgroundSegmentation" />
+
+# useBackgroundSegmentation
+
+The `useBackgroundSegmentation` hook returns functions for creating a segmentation video device and switching effects at runtime, along with `isSupported` state indicating browser compatibility.
+
+This hook provides a unified API for background blur, color replacement, and image replacement. Once a segmentation device is created, you can switch between effects at runtime using `updateEffect` without recreating the device — avoiding blank frames during transitions.
```
devalevenkatesh: nit: wherever you are comparing or referencing a downside of the existing providers, there you can mention "avoiding blank frames during transitions, unlike XProvider. We recommend you switching to using the new `BackgroundSegmentationProvider`".

src/providers/BackgroundSegmentationProvider/docs/useBackgroundSegmentation.mdx
```diff
@@ -0,0 +1,146 @@
...
+      await selectedDevice.stop();
+      await meetingManager.startVideoInputDevice(intrinsic);
+    }
+  };
+
+  return (
+    <div>
+      {isSupported && (
+        <>
+          <button onClick={enableBlur}>Enable Blur</button>
```
devalevenkatesh: Understanding:
Technically one can create the video transform device using None effect and then later update to BLUR, COLOR_REPLACEMENT etc right?

src/providers/BackgroundSegmentationProvider/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+export type BackgroundSegmentationProviderProps = BaseSdkProps &
+  BackgroundSegmentationEffectProps & {
+    /** Segmentation model type. Default: SELFIE_GENERAL */
+    modelType?: ModelType;
+    /** CPU usage budget percentage (10-80). Default: 30 */
+    cpuUsagePercentage?: number;
+  };
+
+export interface BackgroundSegmentationState {
+  /** Whether the V3 segmentation processor is supported in the current browser. */
```
devalevenkatesh: Lets drop V3 since we moved to using unified "BackgroundSegmentation" processor as a naming convention.

src/providers/BackgroundSegmentationProvider/docs/useBackgroundSegmentation.mdx
```diff
@@ -0,0 +1,146 @@
...
+      await meetingManager.startVideoInputDevice(device);
+    } catch (error) {
+      console.error('Failed to enable background blur', error);
+    }
+  };
+
+  // Switch to replacement — no device recreation needed
+  const switchToReplacement = () => {
+    updateEffect({
+      type: ProcessorEffect.COLOR_REPLACEMENT,
```
devalevenkatesh: Should we rename `ProcessorEffect` to `VideoProcessorEffect`?

src/providers/BackgroundSegmentationProvider/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+    const compatibility =
+      BackgroundSegmentationCompatibilityChecker.checkCompatibility(logger);
+    setIsSupported(compatibility.isCompatible);
+
+    if (!compatibility.isCompatible) {
+      logger.warn(
+        `BackgroundSegmentationProvider: Browser not compatible. Missing features: [${compatibility.missingFeatures.join(', ')}]`
+      );
+    }
+  }, []);
```
devalevenkatesh, line 69: Does this need `logger` in dependency?

src/providers/BackgroundSegmentationProvider/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+  modelType = ModelType.SELFIE_GENERAL,
+  cpuUsagePercentage = 30,
+  children,
+}) => {
+  const logger = useLogger();
+  const [isSupported, setIsSupported] = useState<boolean | undefined>(
+    undefined
+  );
+  const processorRef = useRef<VideoFrameProcessor | undefined>(undefined);
+  const modelTypeRef = useRef<ModelType>(modelType);
```
devalevenkatesh: Why do we use refs here? Is it to avoid the un-necessary component mounts on change? If yes, may be the effects need to be split further.

src/providers/BackgroundSegmentationProvider/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+
+    try {
+      const processor = await BackgroundSegmentationVideoFrameProcessor.create(
+        logger,
+        config,
+        modelTypeRef.current,
+        cpuUsageRef.current
+      );
+      processorRef.current = processor;
+      setIsSupported(true);
```
devalevenkatesh: Can we make this state update only based on the compatibility check? and only call initializeProcessor if `isSupported` is true? On mount that supported flag is set. 

Not sure iff:
1. Compatiblity checker is checking for browser APIs only
2. And we need a separate state to actually mention that the model is supported at current provided CPU etc constraints.

src/providers/BackgroundSegmentationProvider/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+    logger.info(
+      `BackgroundSegmentationProvider: Creating segmentation device for: ${JSON.stringify(selectedDevice)}`
+    );
+
+    // Determine the config to use
+    let config = buildConfigFromProps();
+
+    // If effect is 'none' at the provider level, default to blur when consumer
+    // explicitly calls createSegmentationDevice (they want an active effect)
+    if (!config) {
```
devalevenkatesh: Its weird to mount the provider with None effect, but then when they call this API we use Blur + we have updateEffect to actually update the effect. So:
1. Do we need to document that by default the `createSegmentationDevice` will create a blur device?
2. OR take an effect into while creating with the default set to BLUR?

src/providers/BackgroundSegmentationProvider/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+    await initializeProcessor(config);
+    const processor = processorRef.current;
+
+    if (processor) {
+      return new DefaultVideoTransformDevice(logger, selectedDevice, [
+        processor,
+      ]);
+    }
+
+    throw new Error(
```
devalevenkatesh: Should the `isSupported` be set to `false` here?

src/providers/BackgroundSegmentationProvider/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+    if (processorRef.current) {
+      try {
+        await processorRef.current.destroy();
+      } catch {
+        // Processor might have been destroyed already.
+      }
+      processorRef.current = undefined;
+    }
+
+    await initializeProcessor(config);
```
devalevenkatesh: So does this work:
1. Maintain a `config` state
2. Add a `useEffect` on the `config` state update. Whenever it updates, a `processor` is re-created and the a new state for `processor` is updated.
3. Now which ever flow depends on the `processor`, those effects can depend on the processor and then re-update themselves. 

Is this not applicable in this flow and we need to use `ref`?

src/providers/BackgroundSegmentationProvider/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+    (config: BackgroundSegmentationVideoFrameProcessorConfig): void => {
+      if (!processorRef.current) {
+        logger.warn(
+          'BackgroundSegmentationProvider: Cannot updateEffect — processor not initialized. Call createSegmentationDevice first.'
+        );
+        return;
+      }
+
+      const processor =
+        processorRef.current as BackgroundSegmentationVideoFrameProcessor;
```
devalevenkatesh: Instead of `as` does `VideoFrameProcessor` not work? If not should we move the processor to concrete `BackgroundSegmentationVideoFrameProcessor` type when ref is created?

src/providers/BackgroundSegmentationProvider/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+      processor.setConfig(config);
+      logger.info(
+        `BackgroundSegmentationProvider: Effect updated to ${config.type}`
+      );
+    },
+    []
```
devalevenkatesh: Why is this a `useCallback` explicitly? Its mainly using the `ref` which wont update on component re-renders too, so plain JS function is fine?

src/providers/BackgroundSegmentationProvider/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    const processor =
+      processorRef.current as BackgroundSegmentationVideoFrameProcessor;
+    processor.setModelType(newModelType);
+    modelTypeRef.current = newModelType;
+    logger.info(
+      `BackgroundSegmentationProvider: Model type update requested: ${newModelType} (async load in worker)`
+    );
+  }, []);
```
devalevenkatesh: Same as https://github.com/aws/amazon-chime-sdk-component-library-react/pull/1037/changes#r3753416424

src/providers/BackgroundSegmentationProvider/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+
+      const processor =
+        processorRef.current as BackgroundSegmentationVideoFrameProcessor;
+      processor.setMaxCPUUsagePercentage(percentage);
+      cpuUsageRef.current = percentage;
+      logger.info(
+        `BackgroundSegmentationProvider: CPU usage percentage updated to ${percentage}`
+      );
+    },
+    []
```
devalevenkatesh: Same as: https://github.com/aws/amazon-chime-sdk-component-library-react/pull/1037/changes#r3753416424

I think we should have states for each of these and add a `useEffect` depending on them to update the processor. Have requested clarification on understanding whether they need to state or ref up top somewhere.

src/providers/MeetingProvider/index.tsx
```diff
@@ -82,7 +83,9 @@ export const MeetingProvider: React.FC<React.PropsWithChildren<Props>> = ({
                   <LocalAudioOutputProvider>
                     <ContentShareProvider maxContentShares={maxContentShares}>
                       <FeaturedVideoTileProvider>
-                        {children}
+                        <BackgroundSegmentationProvider effect="none">
```
devalevenkatesh: Should this be mounted just below `DevicesProvider`? Mainly we need access to devices hence checking for hierarychy and whether if the segmentation support and APIs are needed / may be needed in other providers like LocalVideo, Roster to show case / act on any state from segmentation provider.

src/providers/BackgroundSegmentationProvider/index.tsx
harshhv-dev: Done. All V3 references removed.

src/providers/BackgroundSegmentationProvider/docs/useBackgroundSegmentation.mdx
harshhv-dev: ProcessorEffect is exported from amazon-chime-sdk-js. Re-exporting with an alias would create naming confusion between ReactSDK docs and JSSDK docs. Prefer keeping the JSSDK name as-is for consistency across the ecosystem. Builders import it from `amazon-chime-sdk-js` directly

src/providers/MeetingProvider/index.tsx
xuesichao: Could we remove `effect` from the parameter and only track it inside `BackgroundSegmentationProvider`?

src/index.ts
```diff
@@ -163,6 +164,11 @@ export { FeaturedVideoTileProvider } from './providers/FeaturedVideoTileProvider
 export { VoiceFocusProvider } from './providers/VoiceFocusProvider';
 export { BackgroundBlurProvider } from './providers/BackgroundBlurProvider';
 export { BackgroundReplacementProvider } from './providers/BackgroundReplacementProvider';
+export { BackgroundSegmentationProvider } from './providers/BackgroundSegmentationProvider';
+export type {
+  BackgroundSegmentationProviderProps,
+  BackgroundSegmentationState,
+} from './providers/BackgroundSegmentationProvider';
```
xuesichao: I'm not sure if we need to export all these, looks like at least we don't need to export `BackgroundSegmentationProvider` cause consumer will use `MeetingProvider`. Not sure about the type, but hook needs to be exported

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
harshhv-dev: Updated. `isSupported` docs now explain it's determined by checking for WebGL2, WebAssembly, Web Workers, and OffscreenCanvas — inline with the field definition.

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
harshhv-dev: Addressed differently — removed `effect` prop entirely. `createSegmentationDevice` now takes `config` as a required second param, so the consumer always specifies the desired effect at creation time. No hidden defaults.

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
harshhv-dev: Done. Explanation moved inline with field definitions in the code block. Separate paragraph removed.

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
harshhv-dev: Added JSSDK imports (including `Device`, `VideoTransformDevice`, `BackgroundSegmentationVideoFrameProcessorConfig`, etc.) in the Importing section alongside the React SDK hook import. Library names distinguish the source.

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
harshhv-dev: My bad. This is misleading. Updated example to show intrinsicDevice() + stop() before passing to createSegmentationDevice. This will be showcases via Demo App sample code as well in upcoming PR

src/providers/BackgroundSegmentationProvider/docs/useBackgroundSegmentation.mdx
harshhv-dev: Updated the docs as suggested.

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
harshhv-dev: Same as https://github.com/aws/amazon-chime-sdk-component-library-react/pull/1037#discussion_r3769352062 -> Code was misleading. Rewrote example using `isProcessorActive` state that's only set to true after successful device creation. No more `isBlurOn` toggle.

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
harshhv-dev: Same as https://github.com/aws/amazon-chime-sdk-component-library-react/pull/1037#discussion_r3769352062 -> Code was misleading. No longer applicable — removed `useEffect` toggle pattern entirely. Example now uses direct `onClick` handlers.

src/providers/BackgroundSegmentationProvider/docs/useBackgroundSegmentation.mdx
harshhv-dev: Yes. createSegmentationDevice(device, config) now takes the effect config directly. updateEffect() is for runtime switching after creation. Both patterns supported.

src/providers/BackgroundSegmentationProvider/index.tsx
harshhv-dev: logger comes from useLogger() (React Context) which provides a stable reference — it doesn't change between renders. The compatibility check should only run once on mount since browser capabilities don't change mid-session. Adding it as a dependency would be harmless but unnecessary.

src/providers/BackgroundSegmentationProvider/index.tsx
harshhv-dev: The reactive pattern (state + useEffect) works for data-driven flows, but processor creation here is user-initiated (button click → create device). The consumer needs the returned VideoTransformDevice immediately to pass to startVideoInputDevice. useEffect can't return values to the caller. Additionally, processor as state would re-render the subtree on every creation/destruction. Thoughts?

src/providers/BackgroundSegmentationProvider/index.tsx
harshhv-dev: Done. `processorRef` typed as `BackgroundSegmentationVideoFrameProcessor | undefined` directly. No more `as` casts.

src/providers/BackgroundSegmentationProvider/index.tsx
harshhv-dev: Removed `modelTypeRef`/`cpuUsageRef` entirely. After removing `useCallback`, these refs were no longer needed — props are read directly at call time. `processorRef` remains as a ref because the processor is an internal implementation detail that shouldn't trigger re-renders when created/destroyed — consumers interact via methods (`updateEffect`, `updateModelType`), not by reading the processor directly. Only `isSupported` (which drives conditional UI) is state.

src/providers/BackgroundSegmentationProvider/index.tsx
harshhv-dev: Agreed. `isSupported` now only reflects the browser compatibility check — set once on mount, never changed by processor creation outcomes. Added `isSupported === false` guard inside `createSegmentationDevice` as safety net.

src/providers/BackgroundSegmentationProvider/index.tsx
harshhv-dev: Done. `createSegmentationDevice(device, config, options?)` — consumer passes effect `config` and optional `modelType`/`cpuUsagePercentage` at creation time. No hidden blur/medium default.

src/providers/BackgroundSegmentationProvider/index.tsx
harshhv-dev: Same as https://github.com/aws/amazon-chime-sdk-component-library-react/pull/1037#discussion_r3771115022 -> `isSupported` is no longer set inside `initializeProcessor`. Throw path doesn't touch `isSupported`.

src/providers/BackgroundSegmentationProvider/index.tsx
harshhv-dev: Done. Removed `useCallback` — plain functions since they only access refs and stable context values.

src/providers/BackgroundSegmentationProvider/index.tsx
harshhv-dev: Same as https://github.com/aws/amazon-chime-sdk-component-library-react/pull/1037#discussion_r3771129549

src/providers/BackgroundSegmentationProvider/index.tsx
harshhv-dev: Same as https://github.com/aws/amazon-chime-sdk-component-library-react/pull/1037#discussion_r3771129549

src/providers/MeetingProvider/index.tsx
harshhv-dev: Done. Although completely irrelevant at this point, moved `BackgroundSegmentationProvider` below `DevicesProvider` in the tree for future use-cases. Gives device context access and allows downstream providers to access segmentation state if needed in future.

harshhv-dev: Done. Removed effect prop entirely. Provider has no props. Consumer controls everything via `createSegmentationDevice(device, config, options?)`.

src/index.ts
harshhv-dev: Removed `BackgroundSegmentationProvider` and `BackgroundSegmentationProviderProps` and `BackgroundSegmentationState` exports. Only `useBackgroundSegmentation` is publicly exported.

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+  // Enable an effect (first activation creates the processor)
+  const enableBlur = async () => {
+    if (!selectedDevice || !isSupported) return;
+
+    try {
+      let current = selectedDevice;
+      if (isVideoTransformDevice(current)) {
+        const intrinsic = await current.intrinsicDevice();
+        await current.stop();
```
devalevenkatesh: Does this not stop and make the video go blank?

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
```diff
@@ -0,0 +1,187 @@
...
+        current = intrinsic;
+      }
+      const device = await createSegmentationDevice(current, {
+        type: ProcessorEffect.BLUR,
+        blurStrength: BackgroundSegmentationBlurStrength.MEDIUM,
+      });
+      await meetingManager.startVideoInputDevice(device);
+      setIsProcessorActive(true);
+    } catch (error) {
+      console.error('Failed to enable background effect');
```
devalevenkatesh: Should we throw this back to builder?

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
```diff
@@ -0,0 +1,187 @@
...
+  const {
+    isSupported,
+    createSegmentationDevice,
+    updateEffect,
+    updateModelType,
+  } = useBackgroundSegmentation();
+  const [isProcessorActive, setIsProcessorActive] = useState(false);
+
+  // Enable an effect (first activation creates the processor)
+  const enableBlur = async () => {
```
devalevenkatesh: nit: we are showing `enableBlur` and then `disableEffect`. Checking should this be `enableEffect`, `disableEffect`? Or we can keep it consistent. `enableBlur`, `disableBlur` if using `enable|disableEffect` then question is "what effect is enabled, disabled?" so either we take `effect` as input or since this is an example, it could be just `enableBlur` and `disableBlur`?

I lean towards `enableEffect(effect = BLUR passed in)`, then only `disableEffect`.

src/providers/BackgroundSegmentationProvider/docs/useBackgroundSegmentation.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+```javascript
+import { useBackgroundSegmentation } from 'amazon-chime-sdk-component-library-react';
+```
+
+The hook is available within any component nested inside `MeetingProvider` — no explicit `BackgroundSegmentationProvider` wrapping needed since it is auto-mounted internally.
+
+## Usage
+
+```jsx
```
devalevenkatesh, line 66: Since this is same code example and repeating can just point to one place in either the hook or the provider. Preferrable hook here since Provider we mount and builder interacts with the hook.

src/providers/BackgroundSegmentationProvider/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+      );
+      processorRef.current = undefined;
+      return undefined;
+    }
+  }
+
+  const createSegmentationDevice = async (
+    selectedDevice: Device,
+    config: BackgroundSegmentationVideoFrameProcessorConfig,
+    options?: { modelType?: ModelType; cpuUsagePercentage?: number }
```
devalevenkatesh, line 127: Can the `options` not be in `config`? Looks like JS SDK has this type merged in already so no changes required but just checking if they should be clubbed together as part of the config itself as optionals.

src/providers/BackgroundSegmentationProvider/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+      try {
+        await processorRef.current.destroy();
+      } catch {
+        // Processor might have been destroyed already.
+      }
+      processorRef.current = undefined;
+    }
+
+    await initializeProcessor(config, resolvedModelType, resolvedCpu);
+    const processor = processorRef.current;
```
devalevenkatesh: This is weird to me. Why not move all operations about the processor, including refs into the `initailizeProcessor` method? That method takes care of destroy if existing -> create -> update the processor.ref. 

This method does not do split logic checks on processor ref etc, it simply assumes processor ref is set and use it after the `initializeProcessor` call.

CHANGELOG.md
```diff
@@ -16,13 +16,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   is omitted.
+- Added `BackgroundSegmentationProvider` and `useBackgroundSegmentation` hook with unified blur, color-replacement, and image-replacement support via `BackgroundSegmentationVideoFrameProcessor`. Auto-mounted inside `MeetingProvider` (zero-cost until activation). Supports `createSegmentationDevice(device, config, options?)` for initial effect, runtime effect switching via `updateEffect()`, model switching via `updateModelType()`, and CPU budget control via `updateCPUUsagePercentage()`.
 
 ### Removed
 
 ### Changed
 
-### Fixed
+- Bumped `amazon-chime-sdk-js` peer dependency to `^3.32.0`.
+- Added `RTCRtpSender`/`RTCRtpReceiver` mock in test setup for JSSDK 3.32.0 JSDOM compatibility.
```
devalevenkatesh: nit: Can move up in Added section.

src/providers/BackgroundSegmentationProvider/index.tsx
devalevenkatesh: If processor need not change on dependencies then I am fine with the props, but we need to make sure processor is updated in any situations explicitly. Left a comment on the current implementation to move the ref updates to the initializeProcessor method.

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
harshhv-dev: Yes, it does create a blank frame. Ideally JSSDK would internally handle stopping of old transform device. But currently in React SDK demo, existing blur and replacement options of V1(mainline) propose stopping an intrinsic device to consumers publicly. Hence, it is kept to be aligned with existing pattern. But yes, we should update this with this upgrade. I will test it out once and update it if no issues.

harshhv-dev: Tested out JSSDK demo and React SDK demo - The blank frame is a known limitation of internal device controller when switching from any processor to None effect or switching from one processor to another processor. Here, `intrinsicDevice() + stop()` pattern is the pre-existing approach used in the V1 React SDK demo (`VideoTransformDropdown.tsx` on main) since `useVideoInputs()` returns the currently active `VideoTransformDevice` rather than the raw camera ID. The JSSDK demo avoids `intrinsicDevice() + stop()` by storing the raw camera ID separately, but the blank frame still exists there too — it's caused by the device controller disconnecting the old output stream before the new camera stream is ready.

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
harshhv-dev: This is a builder-side code example in the Usage section, not library code. The catch block is intentionally minimal — builders implement their own error handling strategy (show UI error, retry, fallback, etc.) as appropriate for their app. The library itself already throws from createSegmentationDevice when unsupported — this catch handles that throw on the consumer side. Although, I will add comments to make it more clear.

src/providers/BackgroundSegmentationProvider/docs/BackgroundSegmentationProvider.mdx
harshhv-dev: You are right. This is inconsistent. Also, our sample app PR does not show different handlers for different effects. Instead it has only one single unified event handler. Will update this section so it is aligned with our sample app PR as well.

src/providers/BackgroundSegmentationProvider/docs/useBackgroundSegmentation.mdx
harshhv-dev: Got it. Will do it.

src/providers/BackgroundSegmentationProvider/index.tsx
harshhv-dev: JSSDK's `BackgroundSegmentationVideoFrameProcessorConfig` is a discriminated union for effect types only (blur/color/image). `modelType` and `cpuUsagePercentage` are processor-level settings, not effect settings — they persist across `updateEffect()` calls. Merging them into config would force consumers to re-specify them on every effect switch(although low-effort required). Keeping them separate (set once at creation, changeable via `updateModelType/updateCPUUsagePercentage`) is cleaner. Although, will discuss with @dinmin-amzn to see if we can club them. But for now, keeping them aligned with JSSDK exports.

src/providers/BackgroundSegmentationProvider/index.tsx
harshhv-dev: Yeah, you are right. Will clean it up.

CHANGELOG.md
harshhv-dev: Done.

src/providers/BackgroundSegmentationProvider/docs/useBackgroundSegmentation.mdx
```diff
@@ -0,0 +1,188 @@
...
+        setIsProcessorActive(true);
+      } else {
+        // Processor already active — seamless switch (no blank frame)
+        updateModelType(modelType);
+        updateEffect(config);
+        updateCPUUsagePercentage(cpuUsagePercentage);
+      }
+    } catch (error) {
+      // Handle effect enablement failure here
+      console.error('Failed to enable background effect', error);
```
devalevenkatesh, line 128: nit: No more background effect as its depends on the `config` passed in.

src/providers/BackgroundSegmentationProvider/docs/useBackgroundSegmentation.mdx
```diff
@@ -0,0 +1,188 @@
...
+      if (isVideoTransformDevice(current)) {
+        const intrinsic = await current.intrinsicDevice();
+        await current.stop();
+        current = intrinsic;
+      }
+      await meetingManager.startVideoInputDevice(current);
+      setIsProcessorActive(false);
+    } catch (error) {
+      // Handle effect disablement failure here
+      console.error('Failed to disable background effect', error);
```
devalevenkatesh, line 147: Same as: https://github.com/aws/amazon-chime-sdk-component-library-react/pull/1037/changes#r3785546361

src/providers/BackgroundSegmentationProvider/index.tsx
devalevenkatesh: Then should not they be reversed? Looks like `BackgroundSegmentationVideoFrameProcessorConfig` is indeed not a `Processor` config but an `Effect` config since `updateEffect` is also taking it :)

Anyways we have merged the JS SDK changes now so may be bigger refactor.

tst/providers/BackgroundSegmentationProvider/index.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+        setModelType: jest.fn(),
+        setMaxCPUUsagePercentage: jest.fn(),
+        getConfig: jest.fn().mockReturnValue({ type: 'blur', blurStrength: 'medium' }),
+        getModelType: jest.fn().mockReturnValue('selfie_general'),
+      }),
+    },
+  };
+});
+
+describe('BackgroundSegmentationProvider', () => {
```
devalevenkatesh, line 40: Currently tests only cover whether a function is exposed mainly and component lifecycle but not the exposed functions itself. Can we add the test coverage for: Actual behavior of the exported functions from the hooks where the processor APIs are called with intended passed in config.

devalevenkatesh (dismissed):

tst/providers/BackgroundSegmentationProvider/index.test.tsx
harshhv-dev: Added behavioral tests.

src/providers/BackgroundSegmentationProvider/docs/useBackgroundSegmentation.mdx
harshhv-dev: Corrected

src/providers/BackgroundSegmentationProvider/docs/useBackgroundSegmentation.mdx
harshhv-dev: Corrected

